PR #540, opened 2013-07-08
IGNORE

joshuaclayton:


tute:
@joshuaclayton should this PR continue to exist?

joshuaclayton:
@tute yes, this is something I've been playing around with, I'd like to keep it open

joshuaclayton:
I've closed this with the intention of reopening if I can work out some of the performance issues here.

Issue #644, opened 2014-04-15
Unsatisfied dependency on DatabaseCleaner, breaks Rails apps.


edelgado:
If you follow the "Linting Factories" instructions in your [GETTING_STARTED.md](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#linting-factories) file, a Rails application will break with this error message:

`uninitialized constant DatabaseCleaner (NameError)`

This is because `DatabaseCleaner` is class defined in the `database_cleaner` gem. This dependency is currently unsatisfied by the `factory_girl` and `factory_girl_rails` gems.

Seems like the gemspecs need to be updated.

jaykilleen:
Same here. I just ran into this issue and then googled it and found this page. Yay that for #645.

PR #645, opened 2014-04-16
Improving the documentation around Factory Linting and the DatabaseCleaner class.

edelgado:
If you follow the "Linting Factories" instructions in your [GETTING_STARTED.md](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#linting-factories) file, a Rails application will break with this error message:

`uninitialized constant DatabaseCleaner (NameError)`

This is because `DatabaseCleaner` is class defined in the `database_cleaner` gem.

After reporting this in [Issue #644](https://github.com/thoughtbot/factory_girl/issues/644), it was suggested that I could help by improving the documentation around Factory Linting and the DatabaseCleaner class in the `GETTING_STARTED.md` file. Hence this pull request :-]

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -88,18 +88,35 @@ RSpec.configure do |config|
   # additional factory_girl configuration
 
   config.before(:suite) do
-    begin
-      DatabaseCleaner.start
-      FactoryGirl.lint
-    ensure
-      DatabaseCleaner.clean
-    end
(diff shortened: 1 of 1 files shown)


jaykilleen:
It may also need to be updated at http://www.rubydoc.info/gems/factory_girl/file/GETTING_STARTED.md as it looks like it is still referencing the old read me.

PR #646, opened 2014-04-16
FactoryGirl should refuse to assign to a private setter.

Peeja:
As part of a refactoring, I'm making one of the association setters on one of my ActiveRecord models private, something like:

``` ruby
class Human < ActiveRecord::Base
  belongs_to :favorite_soda, class_name: :Soda
  private :favorite_soda=
end
```

But FactoryGirl will still gladly assign to it:

``` ruby
FactoryGirl.build(:human, favorite_soda: Soda.first)
```

This is because the `AttributeAssigner` uses `#send` instead of `#public_send`. This PR changes that.

I haven't added a test for it because I wasn't sure where it should be tested. `AttributeAssigner` doesn't appear to have its own spec (though I might just not be seeing it).

This change doesn't break any specs, but it's possible that users are depending on FactoryGirl being able to assign to private setters. I think it's unlikely they're intending to do that, though. At least, I can't come up with a reason you'd want to make something private but then assign it from a factory.

Thoughts?

### DIFF
--- a/lib/factory_girl/attribute_assigner.rb
+++ b/lib/factory_girl/attribute_assigner.rb
@@ -13,7 +13,7 @@ module FactoryGirl
       @evaluator.instance = build_class_instance
       build_class_instance.tap do |instance|
         attributes_to_set_on_instance.each do |attribute|
-          instance.send("#{attribute}=", get(attribute))
+          instance.public_send("#{attribute}=", get(attribute))
           @attribute_names_assigned << attribute
         end
       end


asafbrukarz:
Hi @joshuaclayton,

Sorry to rehash an old issue, but we were indeed using this functionality (whether it was intentional or not is another story), but while upgrading FactoryGirl, we only saw a minor version bump (4.4.0 -> 4.5.0). While I understand the desired functionality, this is a breaking change that should have had a major version bump and/or a clearer description in the changelog: `"Use more explicit #public_send when doing attribute assignment"`.

dnagir:
It does make sense to use `public_send` generally, but I believe FactoryGirl has been using a regular `send` for ages.

And it was actually handy to "cheat" from the factories sometimes.
Of course all the factories are "trusted" sources so it doesn't really matter too much generally.

But using `public_send` just makes it a lot harder than it used to be. We have to resort to transient attributes and modify the original class.

All just to set up a simple column (in ActiveRecord) that isn't part of public interface of the class.

I understand the good intention of using `public_send`.
Practically it probably doesn't really matter but introduces major incoveniences.

What is your take on this guys?

joshuaclayton:
@asafbrukarz @dnagir thanks for the feedback.

I'm not sure I agree there should've been a major bump - as the public interface didn't change and felt like exploiting incorrect functionality (similar to overriding a private class functionality and mentioning that a minor bump broke said functionality) - but do understand it may have been confusing and unexpected.

There are a couple of options:

1. Retrofit a flag allowing users to continue to use `#send` instead of `#public_send`, generate a deprecation warning when the method called is private/protected, pull 4.5.0, and release 4.5.1.
2. Leave the code as-is.

Thoughts?

dnagir:
@joshuaclayton thanks for the constructive reply.

My personal preference (but I'm biased here :smile: ) is to revert the commit as to keep the old functionality.
I understand the commit and generally agree with it however it isn't a response to a real issue.

As I said for all the practical purposes there's really no need to use `public_send` because we would be protecting from ourselves.

The best way IMO would be to keep using `public_send` but provide an easy to use alternative of assigning private variables somehow.

PR #662, opened 2014-05-29
public_send attribute assignment

joshuaclayton:


dre-hh:
pls guys let's just revert it to using `send` back again and release 4.5.1
i have to agree with @dnagir's comment on a pull request wich was closed before https://github.com/thoughtbot/factory_girl/pull/646

Factory Girls is not supposed to test any functionality. It's supposed to create blueprints fast. If you want to test that some assignment is working correctly you should write an extra spec.

This change for example breaks the possibility to assign a custom id to an activerecord model (because it's private) and now tons of specs at our codevase are broken

PR #709, opened 2014-10-19
Detailed list syntax method

nicohvi:
In order to add certain specific attributes for a designated amount of records, one can now specify an array of values for an attribute (in addition to the regular way of defining traits and attributes), and the first `n` records (where `n` is the length of the array) will have their specified attributes overwritten by the values in the provided array.

For instance, if one wanted to create 5 users whom all have the `admin` trait, and want to name the first *three* users `Tim`, `Frank` and `Ted`, then this could be accomplished like so:

```ruby
create_detailed_list(:user, 5, :admin, names: ['Tim', 'Frank', 'Ted'])
```

There's more detailes in the `GETTING_STARTED` document.

I've added specs for the new method, though there might be some edge cases that I have not thought of. Also, the method body of `create_detailed_list` could probably look better, but I haven't been able to think of a way to improve it.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -866,6 +866,23 @@ built_users   = build_pair(:user) # array of two built users
 created_users = create_pair(:user) # array of two created users
 ```
 
+If you need to specify certain attributes for *some* of the records, you can supply an array
+in the hash of attributes instead of a single value - though be sure to pluralize the name!
+
+```ruby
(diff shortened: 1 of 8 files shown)


joshuaclayton:
@nicohvi while I appreciate the work done here, I think the additional complexity outweighs the benefits of this; I'm going to close without merging.

nicohvi:
No worries, thanks for looking into it and for this lovely gem! :+1:

PR #710, opened 2014-10-22
Added backtraces to Linter

jackkinsella:
Previously it was impossible to tell why the Factory creation failed.

### DIFF
--- a/lib/factory_girl/linter.rb
+++ b/lib/factory_girl/linter.rb
@@ -32,9 +32,10 @@ module FactoryGirl
     end
 
     def error_message
-      lines = invalid_factories.map do |factory, exception|
-        "* #{factory.name} - #{exception.message} (#{exception.class.name})"
-      end
+       lines = invalid_factories.map do |factory, exception|
+       %Q[* #{factory.name} (#{exception.class.name})\n
+          #{exception.backtrace.join("\n")} ]
+       end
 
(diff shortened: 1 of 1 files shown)


jackkinsella:
Sorry, I don't check my Github notifications often enough :-P

I was retrofitting tests onto a 100 model untested codebase and running the linter gave the following output.

 The following factories are invalid:

       * line_item_adjustment - undefined method `taxons' for nil:NilClass (NoMethodError)
       * buyer_inquiry - undefined method `taxons' for nil:NilClass (NoMethodError)
       * line_item - undefined method `taxons' for nil:NilClass (NoMethodError)
     ..... 

With the proposed changes I got more informative output which enabled rapid debugging:

The following factories are invalid:
       * line_item_adjustment (NoMethodError)

      /Users/jkinsella/code/oxnotes4/app/models/product_builder.rb:142:in `assign_to_appropriate_taxon'
      /Users/jkinsella/code/oxnotes4/app/models/spree/product.rb:562:in `handle_country_change'

I would propose that full stack traces at least be *an option* within the factory_girl linter.

seanlinsley:
I hate to +1 a year-old PR, but this would be really useful when trying to figure out linting failures. I've come across an esoteric exception via the linter many times, and it takes much longer than it should to find the source. This PR would help significantly.

Issue #66, opened 2010-07-03
Factory.build + association should not hit the db?


shwoodard:
Ordinarily when using Factory.build one does not expect FG will hit the db?  Should FG possibly create and populate the associated records when Factory.build  is used on a factory (as opposed to Factor.create) that has associations, but not create the records?  Would this be possible?

Issue #64, opened 2010-06-29
Building an object creates associations


jcf:
When you define and build a factory any associations are created, rather than built. This is slow and feels kinda quirky. Instead the associated objects could be built as you would in Rails, only using memory as opposed to foreign keys and potentially lots of selects and inserts.

I've forked, branched and committed a change, with passing tests that hopefully will fix this in a way you agree is acceptable.

http://github.com/jcf/factory_girl/tree/build_association

Regards,

James

PR #749, opened 2015-02-22
Use parent strategy when building associations

bradx3:
This PR makes associations be created using the same strategy as their parent. i.e. if a Post has a User association and we build a Post, we should build a User as well - not create it.

#### Background

The source of this behaviour seems to go back to [this gist](https://gist.github.com/jferris/569231).

There's a bit of discussion on #64 and #66, but then it seems to fall off the map a bit when the option to explicitly name a strategy is added.

#### Motivation

The main project I'm on at the moment has a fairly deep association chain. We use build_stubbed where possible, but find that we need to duplicate a lot of setup because after(:build) traits don't run with build_stubbed. At times, we're hitting the DB 5 times for associated creates for a single insert.

When looking into why we were hitting into the DB at all when building factories, I came across a number of other blog posts and pull requests of confused people. This does seem like confusing behaviour to most and the particular use case in the gist seems arguably incorrect - Post should probably validate _user_ is present, and if it really does want to validate _user_id_ then the factory should either set that or use create explicitly.

#### Results

I've struggled to find public codebases using factory girl a lot - please point me to any if you know of them.

Pending that info, I've tried this on private projects and have seen non-js suites drop by ~ 60% with no changes or failures.

### Summary

So.... the situation seems to be:

Cons:
- This *might* break things for people relying on (arguably) broken validates_presence_of foreign_key tests

Pros:
- This *should* speed things up for people using build and associations
- This makes the build association behaviour much less surprising

All those changes are pretty hard to quantify, so I totally understand if it's not a change you think it worth the risk. No harm in raising the discussion though, right? :-) Thanks for reading!

### DIFF
--- a/lib/factory_girl.rb
+++ b/lib/factory_girl.rb
@@ -60,9 +60,12 @@ module FactoryGirl
   end
 
   class << self
-    delegate :factories, :sequences, :traits, :callbacks, :strategies, :callback_names,
-      :to_create, :skip_create, :initialize_with, :constructor, :duplicate_attribute_assignment_from_initialize_with,
-      :duplicate_attribute_assignment_from_initialize_with=, to: :configuration
+    delegate :factories, :sequences, :traits, :callbacks, :strategies,
+             :callback_names, :use_parent_strategy, :to_create, :skip_create,
+             :initialize_with, :constructor,
+             :duplicate_attribute_assignment_from_initialize_with,
+             :duplicate_attribute_assignment_from_initialize_with=,
+             to: :configuration
   end
 
   def self.register_factory(factory)
@@ -125,6 +128,10 @@ module FactoryGirl
   def self.register_callback(name)
     name = name.to_sym
     callback_names << name
+  end
+
+  def self.use_parent_strategy=(value)
+    configuration.use_parent_strategy = value
   end
 end
 
--- a/lib/factory_girl/configuration.rb
+++ b/lib/factory_girl/configuration.rb
@@ -3,6 +3,8 @@ module FactoryGirl
   class Configuration
     attr_reader :factories, :sequences, :traits, :strategies, :callback_names
 
+    attr_accessor :use_parent_strategy
+
     def initialize
       @factories      = Decorator::DisallowsDuplicatesRegistry.new(Registry.new('Factory'))
       @sequences      = Decorator::DisallowsDuplicatesRegistry.new(Registry.new('Sequence'))
--- a/lib/factory_girl/evaluator.rb
+++ b/lib/factory_girl/evaluator.rb
@@ -23,7 +23,13 @@ module FactoryGirl
 
     def association(factory_name, *traits_and_overrides)
       overrides = traits_and_overrides.extract_options!
-      strategy_override = overrides.fetch(:strategy) { :create }
+      strategy_override = overrides.fetch(:strategy) do
+        if FactoryGirl.use_parent_strategy
+          @build_strategy.class
+        else
+          :create
+        end
+      end
 
       traits_and_overrides += [overrides.except(:strategy)]
 
--- a/spec/acceptance/build_spec.rb
+++ b/spec/acceptance/build_spec.rb
@@ -27,9 +27,16 @@ describe "a built instance" do
     expect(subject.user).to be_kind_of(User)
     expect(subject.user).not_to be_new_record
   end
+
+  it "assigns but does not save associations when using parent strategy" do
+    FactoryGirl.use_parent_strategy = true
+
+    expect(subject.user).to be_kind_of(User)
+    expect(subject.user).to be_new_record
+  end
 end
 
-describe "a built instance with strategy: :build" do
+describe "a built instance with strategy: :create" do
   include FactoryGirl::Syntax::Methods
 
   before do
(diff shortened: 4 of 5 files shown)


imajes:
on quick scan, :+1: like a bunch. I am really for any and all work that simplifies the effort it takes to reduce hits to the DB.

lime:
This would be an immense improvement!

Currently, I end up explicitly changing the strategy to `:build_stubbed` for every association. However, the few times I actually `:create` the parent model, it often want the association to use `:create` as well. This change would allow me to remove the explicit strategies for most factories and rely on the default behaviour.

If there's anything I can do to help this getting merged, please tell me.

joshuaclayton:
@bradx3 thanks for bringing this up - it's been on my mind for a while, and I'm often confused as to why we'd gone down the path of there being differences anyway (specifically, `build` will `create` associations).

@jferris do you recall what the context was, or if there were any specs that accurately captured the issues we'd seen and are trying to prevent, to see if AR has changed in such a manner that it's now not applicable?

jferris:
Assigning an unsaved object to an association has some pitfalls in ActiveRecord. One of the most common was having validations like `validates_presence_of :user_id`; an unsaved object won't have a foreign key, so it ends up being invalid.

I think we'd need to double check our integration tests to figure out if we have coverage for all these scenarios. Alternatively, you could make the change and try it out in a few apps. I wouldn't be surprised if you saw some issues.

lime:
If it turns out to be a breaking change (somewhat likely) then I'd suggest making this behaviour a configurable option in a minor release, and the default behaviour in the next major version bump.

CJBridges:
I'm really interested in having this change made available.  I didn't test for the functionality of your changes, but did verify on a private project (fork factory_girl master + cherry-pick the commit in this PR, my Gemfile pointing to local repo) that 100% of my suite (~1300 examples, not all using factory_girl of course) was passing properly with the changes included.

Our reason is the fairly simple one mentioned by Brad in the original filing: we do a fair bit of testing using factory_girl and are seeing a good number of cases where "build" ends up "creating" a lot of objects in the background, solely due to the relations defined in the factories.  This in turn slows down tests overall since they're hitting the database way more than they need to.

One caveat is that ActiveRecord relations often trigger a save just by the way that making the association functions.  See one list (not mine) here: https://gist.github.com/demisx/9896113  Even with these changes, I believe that "build" can still create some objects in the DB.

bradx3:
@jferris there's been a few changes to the acceptance test behaviour in this PR. Are there any other integration tests you're thinking of that I can check? Are they not in this project maybe?

Re: the ```validates_presence_of :user_id``` risk - I still see that as the symptom of an incorrect factory. If your model *really* needs to check ```:user_id``` rather than ```:user```, shouldn't the factory be setting ```user_id``` explicitly? 

Regardless of the correct/incorrect way to deal with that particular case, it does seem like a problem that will come up. Is there interest in the suggestion of:

1. Adding a configurable option, and possibly
2. Removing the option next major release?

jferris:
@bradx3 I like the idea of adding an option for the transition. We did that before and I feel like it worked out well.

It would be cool to pair this option with something that makes it easier to get an ID if you need one.

lime:
> It would be cool to pair this option with something that makes it easier to get an ID if you need one.

Just a humble request: can we consider that as a completely separate issue, so as not to slow this PR down? :grin:

bradx3:
Sounds fair @lime. I'll dig around in the repo for some examples of user-set options and implement that first.

Unfortunately my next two weeks are crazy, so that might not be until later this month. I'm sorry I've let this PR drag on for so long, but I guess it's more a nice-to-have than a must-have, so a bit of extra time to make sure we get it right doens't seem *too* bad. :grimacing: :

lime:
You're the one doing the work here @bradx3, I don't mean to rush you at all. :)

I was referring to _the rest of us_, that we shouldn't bloat the PR with feature requests that only add to your workload.

bradx3:
Hey folks. I've had a play with adding an option to keep the old behaviour (create on built associations), but i'm not really happy with the implementation - I've basically had to add a "create_all_associations" boolean onto the FactoryGirl.configuration object. That's doesn't feel right to me at all, but it's the best I've got :-( I've looked through history and can't find anything similar. Does anybody have any ideas or links for sample implementations?

The only "option" I can see is to explain the change in the changelog/readme - it's still possible and easy for users to set a create strategy manually (see example below) so maybe explaining that is sufficient?

    association(:user, strategy: :create)

joshuaclayton:
@bradx3 I think the PR you've got is actually pretty close - and adding a configuration option (as a boolean) seems totally reasonable to me. Once this is merged, we can cut a minor release, leaving the existing behavior, and allowing people to opt into it building instead of creating to see how it goes. If the consensus is that the new behavior makes sense, we can add a deprecation warning, and once 5.0 is shipped, flip the behavior (defaulting to the parent build strategy instead of `:create`). Thoughts?

lime:
Sounds perfect to me! :+1:

bradx3:
Thanks for the feedback @joshuaclayton @lime. I've added a FactoryGirl.use_parent_strategy option. Thoughts?

conf:
Any progress here? Just a kind reminder, guys :)

bradx3:
I'm not able to push this PR forward any more, so I think the best thing might be to close it.

Obviously the code will be there for anybody else to grab and take forward, but I think "closed" is probably the most accurate for this PR at this point.

PR #751, opened 2015-02-25
Linting should not be the first chapter of "getting started"

jaredbeck:
In fact, it probably shouldn't be in a "getting started" guide at all, but this commit simply moves it down towards the bottom.


jaredbeck:
@joshuaclayton One month ping.  Can we get a yes/no on this?

rafaelgonzalez:
Just chiming in to give my two cents.

Whenever I set up FactoryGirl on a new project, I usually stop at/skip the linting section. I simply never ever used it. Model validations always proved to be sufficient.

joshuaclayton:
@jaredbeck thanks; merged as d2ceaac72ead5f78f87819d703403458558f4259

PR #760, opened 2015-03-16
Fix warnings

agrimm:
Fix a couple of warnings generated when Ruby is run in verbose mode and the gem is required.

### DIFF
--- a/lib/factory_girl/linter.rb
+++ b/lib/factory_girl/linter.rb
@@ -15,9 +15,10 @@ module FactoryGirl
       end
     end
 
-    private
-
     attr_reader :factories_to_lint, :invalid_factories
+    private     :factories_to_lint, :invalid_factories
+
+    private
 
     def calculate_invalid_factories
       factories_to_lint.inject({}) do |result, factory|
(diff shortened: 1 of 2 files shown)


lib/factory_girl/linter.rb
```diff
@@ -10,17 +10,15 @@ def initialize(factories_to_lint)
...
     def lint!
-      if invalid_factories.any?
+      if @invalid_factories.any?
         raise InvalidFactoryError, error_message
       end
     end
 
     private
 
-    attr_reader :factories_to_lint, :invalid_factories
```
chastell: Rather than removing the private `attr_readers` and falling back to instance variables, the ([silly](https://bugs.ruby-lang.org/issues/10967)) warning can be worked around like this:

```ruby
# …public methods…

attr_reader :factories_to_lint, :invalid_factories
private     :factories_to_lint, :invalid_factories

private

# …private methods…
```

bruno-:
Hi guys,
I'm also getting warnings for this and find it annoying. It would be great if we could get a word from the maintainers on this.

joshuaclayton:
@chastell thanks for pointing that out - @agrimm mind making that update, squashing commits, and rebasing against master so we can get this merged?

agrimm:
@joshuaclayton @bruno- Fixed.

joshuaclayton:
Merged as 37d61983ef5446ab3ece900eb49785059457eba4

PR #762, opened 2015-03-26
Formatting

rubyengineer:
Formatting desc according to general guidelines, let's be perfectionists. :)


joshuaclayton:
@rubyengineer we've actually been gradually moving back to double-quotes, so I'm going to close this without merging. Thanks!

PR #771, opened 2015-04-17
Make debugger friendly.

yoshiokatsuneo:
When debugger's break points are set at FactoryGirl, FactoryGirl does not
work but show error like below.
 "FactoryGirl::AttributeDefinitionError: Attribute already defined: is_a?"
At breakpoint, debugger tried to show variables but FactoryGirl defines
all unknown methods that may be called from debbuger that cause trouble.
This commit changes method_missing to call DefinitionProxy's original method
or just return when methods are called from debugger.

### DIFF
--- a/lib/factory_girl/definition_proxy.rb
+++ b/lib/factory_girl/definition_proxy.rb
@@ -1,8 +1,9 @@
 module FactoryGirl
   class DefinitionProxy
-    UNPROXIED_METHODS = %w(__send__ __id__ nil? send object_id extend instance_eval initialize block_given? raise caller method)
+    UNPROXIED_METHODS = %w(__send__ __id__ nil? send object_id extend instance_eval initialize block_given? raise caller caller_locations method methods respond_to? puts)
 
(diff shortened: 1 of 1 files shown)


joshuaclayton:
@yoshiokatsuneo can you provide a bit more context with where you'd be introducing the debugger (in the factory definition block itself vs in attribute blocks)?

This introduces a fair amount of complexity that I'm hesitant to merge given everything it touches.

yoshiokatsuneo:
@joshuaclayton It is about "factory" blocks. Any method debugger try to call inside factory is handled not as the method but as attribute name, so just calling like "self.is_a?" inside factory block break things.

joshuaclayton:
@yoshiokatsuneo I'm going to close this without merging; I think the additional complexity, as well as where the debugging is taking place, makes this difficult to merge. `debugger` after executing a factory is almost always going to get closer to the problem at hand; `debugger` in the scope of the factory definition will put you in FG and not the model itself.

PR #773, opened 2015-04-25
Added linting for traits

clemcke:


joshuaclayton:
@clemcke I like the direction this is going; I do think it's important to also include information about the trait that's incorrect, in addition to the factory. Thoughts?

clemcke:
I agree.  I've been using this locally and it would be useful to have more descriptive errors.  I was also thinking about linting all possible combinations of traits.  Do you think that would be useful or do you think that would be overkill?

joshuaclayton:
@clemcke combinations of traits seem like it'd add too much overhead/time to the point where these wouldn't be useful - I'm also unsure that there's a time where two traits would interact in such a manner that would invalidate the model where each on its own wouldn't (that's not to say I couldn't fabricate one in my head - just that my hunch is it's very highly unlikely).

I think sticking with printing out the factory+trait in the error, and avoiding combinations, is the best way to proceed here.

clemcke:
I've added more descriptive error messages for invalid lints.

joshuaclayton:
@clemcke this is definitely a good direction! There are two corner cases I think we'll want to handle:

* Invalid factory *and* an invalid trait for that factory - I believe the rescue will short-circuit subsequent runs for traits, meaning only the first will be captured
* Valid factory but multiple invalid traits - similar to above, if there are two invalid traits for one factory, we'd want to see both listed in the failures

clemcke:
Ah good point, I'll make sure it is verbose for all errors.

ndp:
Sorry, didn't see this before I put in #800  This is the same idea, but #800 make puts in a switch to turn it on-- so maybe easier to absorb into the project.

joshuaclayton:
Linting traits has been merged via #800 and 702487072ca67a37e9b19b97419fdfec1ee03b29

PR #777, opened 2015-05-09
Add TOC/navigation to GETTING_STARTED.md

tigershen23:
It's a long file full of useful information; this list of links adds:
* a list of documented FactoryGirl features and examples at a glance
* easier navigation
* an outline of the document itself

Cheers!

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -1,6 +1,30 @@
 Getting Started
 ===============
 
+* [Update Your Gemfile](#update-your-gemfile)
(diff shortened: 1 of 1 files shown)


joshuaclayton:
@tigershen23 is there a way to generate this automatically? My (perhaps unfounded!) concern is that the headers would become out of date, relative to the document, and could potentially cause churn if GH were to change how it generates ids for anchoring. Thoughts?

joshuaclayton:
Given the overhead of manually having to update this, and have every contributor to do so, I'm going to close this without merging.

PR #783, opened 2015-05-20
Add info to notification's payload

lastcat:
Thank you for this useful gem very much. It always help me with my work.
Well, I wanted to know the detailed information of more executed factory.  Example at `GETTING_STARTED.md `, give me factory name, execution time , and stragy. However, I want to get more details infos. for example traits, association, ... etc.
 
So when I read code,I found factory_runner can access `@traits, @overrides`, and `factory`object. These data help my testing to more effective.

I think there is no reason not to add these data to payload. Please consider this pr.
Finally, I'm sorry my poor English...

### DIFF
--- a/lib/factory_girl/factory_runner.rb
+++ b/lib/factory_girl/factory_runner.rb
@@ -17,7 +17,7 @@ module FactoryGirl
         factory = factory.with_traits(@traits)
       end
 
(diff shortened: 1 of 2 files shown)


joshuaclayton:
@lastcat thanks! I've merged this as 95c6d96306fe90e6ef83f6b80f67d206b4da4a37

PR #791, opened 2015-06-26
Add instructions to GETTING_STARTED.md

amosjyng:
Additional "require 'support/factory_girl'" is necessary after https://github.com/rspec/rspec-rails/commit/defb19a4db9e817dbdb068f8e0b4db552ffef061

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -34,6 +34,7 @@ Configure your test suite
 RSpec.configure do |config|
   config.include FactoryGirl::Syntax::Methods
 end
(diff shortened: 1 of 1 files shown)


joshuaclayton:
@amosjyng it's up to each developer to ensure the RSpec configuration gets run, and I'd prefer to decouple this from the lack of requirement RSpec has by default.

amosjyng:
Sure thing, not sure I understand though, it's an instruction for using factory_girl with the latest rspec. I personally stumbled on this bit when trying to use factory_girl in my project.

PR #794, opened 2015-07-02
rspec configuration changed from spec/suport to rails_helper.rb

xw19:
config line in rspec is now added to the rails_helper.rb file so that factor girl methods can be accessed the previously on support/factory_girl.rb which dont work.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -30,8 +30,11 @@ Configure your test suite
 
 ```ruby
 # RSpec
-# spec/support/factory_girl.rb
+# Add the following line inside the Rspec.configure block of the spec/rails_helper.rb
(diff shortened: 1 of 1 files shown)


joshuaclayton:
@xw19 I wouldn't actually recommend moving FG configuration into rails_helper itself; instead, I'd require the file explicitly (since the code to require everything from `spec/support` isn't enabled by default). This seems to better follow "single responsibility principle", since `spec/support/factory_girl.rb` is the canonical place for FG config. Thoughts?

xw19:
@joshuaclayton I agree but in documentation it should be clearly mentioned to require that file in rails_helper

composerinteralia:
See #1141

PR #797, opened 2015-07-14
Remove reference to DatabaseCleaner in README

vjt:
As not everyone wants to include database_cleaner, the example above uses FactoryGirl's APIs to get the factored models, on which you can then call their APIs to clean the records up.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -840,10 +840,11 @@ RSpec.configure do |config|
 
   config.before(:suite) do
     begin
-      DatabaseCleaner.start
       FactoryGirl.lint
     ensure
-      DatabaseCleaner.clean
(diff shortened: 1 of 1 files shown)


joshuaclayton:
@vjt One of the benefits of DatabaseCleaner is that it *does* use the ORM/ODM's correct API. Additionally, `delete_all` isn't guaranteed to delete dependent relationships, so in a case where there's a `before_create` callback on a User model that creates a profile, but FG doesn't define a profile factory, that record would not be deleted.

As the goal is to suggest a DB-agnostic solution that ensures the database is entirely clean, and because we have no control over callbacks and how others' write their code, I think DatabaseCleaner remains the best recommendation.

vjt:
OK, makes sense. Thank you.

PR #798, opened 2015-07-16
Update docs on associations

cpjolicoeur:
According to the example given, the `User` record is the one that is not persisted, not the `Post`


joshuaclayton:
@cpjolicoeur the documentation looks correct to me; when we `build(:post)`, it creates any associations (in this case, an instance of `User`) but does not save the factory being built (`post`).

PR #799, opened 2015-07-21
Undef possible method before defining

amatsuda:
Without these changes, factory_girl causes massive number of Ruby :warning:s as follows when run with `--warning` flag:

> warning: method redefined; discarding old ...
> warning: previous definition of ... was here

### DIFF
--- a/lib/factory_girl/evaluator.rb
+++ b/lib/factory_girl/evaluator.rb
@@ -64,6 +64,10 @@ module FactoryGirl
     end
 
     def self.define_attribute(name, &block)
+      if method_defined?(name) || private_method_defined?(name)
+        undef_method(name)
+      end
+
       define_method(name) do
         if @cached_attributes.key?(name)
           @cached_attributes[name]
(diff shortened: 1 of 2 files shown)


joshuaclayton:
@amatsuda any ideas of why these methods are being redefined? Might it make more sense to remove any duplication that would cause them to be defined multiple times instead of redefining?

amatsuda:
For example, it happens while loading a factory like this:

```ruby
factory :user do
  name "Friendly User"
  login { name }

  trait :male do
    name   "John Doe"
    gender "Male"
    login { "#{name} (M)" }
  end

  trait :female do
    name   "Jane Doe"
    gender "Female"
    login { "#{name} (F)" }
  end

  trait :admin do
    admin true
    login { "admin-#{name}" }
  end

  factory :male_admin,   traits: [:male, :admin]   # login will be "admin-John Doe"
  factory :female_admin, traits: [:admin, :female] # login will be "Jane Doe (F)"
end
```

This example is copied from the ***Traits that define the same attributes won't raise AttributeDefinitionErrors; the trait that defines the attribute latest gets precedence.*** section of the [Getting Started document](https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md).

joshuaclayton:
@amatsuda thanks! I've merged this as 241e8e5fb4e79865c2a9cd3c8ff858668f6ce2a9

amatsuda:
@joshuaclayton Awesome! Thank you!

PR #800, opened 2015-08-01
Linting Traits

ndp:
This PR adds the ability to turn on linting for traits as well as factories. This is something we use in our own projects, so we wanted to share it.

To get there, I needed to allow passing options into the linting code.


lib/factory_girl/linter.rb
```diff
@@ -1,11 +1,12 @@
 module FactoryGirl
   class Linter
-    def self.lint!(factories_to_lint)
-      new(factories_to_lint).lint!
+    def self.lint!(factories_to_lint, options)
```
joshuaclayton: instead of having the conditionals spread across this class, thoughts on having there be two linters, one for factories and one for traits, or having two class methods and having the things being linted have a more standard interface?

ndp: Thanks for the input. I think I see what you're saying. Is it mainly one or both?
* you'd like an API that separates the single lint! call into two calls? I naturally wrote ``lint!(factories, traits: true)`-- a single-- but can certainly appreciate other opinions
* `calculate_invalid_factories` is too complex, doing two different things.

Let me know, and I can rev per your feedback. I'm eager to get this merged.

joshuaclayton: @ndp from an interface perspective, yes; you're passing a flag to a class method to do something additional conditionally, and the method to calculate factories has a second round of behavior that's now significantly more complex to test.

What if a "factory" in the sense here is an array of size N, where N can be 1 (just the factory) or 2 (factory plus a trait) or 4 (if we want to get into trait combinations and the factory has three traits) - and treat each as its own factory that can be built (with `FactoryGirl.create(*factory_and_traits)`)?

I still think there's value in opting into trait linting - since it's going to add a lot of additional time, depending on the object model - but I'm not sure where the best place to handle that would be. Possibly configuration?

ndp: Thanks. Given all that, I think just a separate class will make it cleaner, so it's just a separate feature. Revision on its way.

ndp:
Okay, I separated out the trait linting from the factory linting. I think this did simplify the code. To share code, I created a base class with common behaviors. It's backward compatible, and includes tests and docs.

ndp:
@joshuaclayton I think this is ready to go. The implementation is decoupled, but the actual API is unified, which I think will match the desired usage. Ok to merge?

ndp:
@joshuaclayton ping

lib/factory_girl/linter_base.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      <<-ERROR_MESSAGE.strip
+The following factories are invalid:
+
+#{lines.join("\n")}
+      ERROR_MESSAGE
+    end
+  end
+end
+
```
ndp: Okay, I got you @houndci

GETTING_STARTED.md
```diff
@@ -866,6 +866,20 @@ FactoryGirl.lint factories_to_lint
 
 This would lint all factories that aren't prefixed with `old_`.
 
+Traits can also be linted. This option verifies that each  
+and every trait of a factory generates a valid object on its own. 
+This is turned on by passing `traits: true` to the `lint` method:
+
+```ruby
+FactoryGirl.link traits:true
```
joshuaclayton: `link` -> `lint`?

spec/acceptance/lint_spec.rb
```diff
@@ -62,4 +62,102 @@
...
+            name "Yep"
+            trait :renamed do
+              name "Yessir"
+            end
+          end
+        end
+
+        expect do
+          FactoryGirl.lint traits: true
+        end.to_not raise_error
```
joshuaclayton: `to_not` -> `not_to`?

spec/acceptance/lint_spec.rb
```diff
@@ -62,4 +62,102 @@
...
+            trait :unnamed do
+              name nil
+            end
+          end
+        end
+
+        expect do
+          FactoryGirl.lint traits: false
+          FactoryGirl.lint
+        end.to_not raise_error
```
joshuaclayton: `to_not` -> `not_to`?

spec/acceptance/lint_spec.rb
```diff
@@ -62,4 +62,102 @@
...
+      factory :user do
+        name "assigned"
+      end
+    end
+
+    FactoryGirl.lint validate_traits: true
+  end
+
+  describe "trait validation" do
+    context "turned on" do
```
joshuaclayton: "turned on" -> "enabled"?

spec/acceptance/lint_spec.rb
```diff
@@ -62,4 +62,102 @@
...
+          end
+        end
+
+        expect do
+          FactoryGirl.lint traits: true
+        end.to_not raise_error
+      end
+    end
+
+    context "turned off" do
```
joshuaclayton: "turned off" -> "disabled"?

lib/factory_girl.rb
```diff
@@ -55,8 +57,16 @@ def self.reset_configuration
     @configuration = nil
   end
 
-  def self.lint(factories_to_lint = FactoryGirl.factories)
-    Linter.lint!(factories_to_lint)
+  def self.lint(factories = nil, options = {})
+    if factories.is_a?(Hash)
+      factories_to_lint = FactoryGirl.factories
+      options = factories
```
joshuaclayton: this feels a bit strange to me - is there a way we could take in arguments and make this all more clear? I believe ActiveSupport has an `extract_options!` method that *might* be doing what you want

lib/factory_girl.rb
```diff
@@ -55,8 +57,16 @@ def self.reset_configuration
...
+    else
+      factories_to_lint = factories || FactoryGirl.factories
```
joshuaclayton: does changing the method signature to have this be the default (versus `nil`) allow this just to be `factories`?

lib/factory_girl/linter_base.rb
```diff
@@ -0,0 +1,49 @@
+module FactoryGirl
+  class LinterBase
```
joshuaclayton: instead of the template method pattern, could we instead use the strategy pattern, where the linter itself can be calculated where the flag of including traits (or not) passes in different strategies for building out the list of things to be linted? I think the code here is close to that already, but it removes the reliance on inheritance here, which I think we should shoot for.

ndp: Sure, if this is what you want. I was trying to preserve the existing code and not make this relative small feature have too much complexity to it. But yeah, a strategy would be better and my preference as well.

ndp:
Yeah, green again. Good to release?

ndp:
Okay, good to go?

joshuaclayton:
All set! Merged as 702487072ca67a37e9b19b97419fdfec1ee03b29.

Thanks @ndp!

PR #812, opened 2015-09-10
Update README.md

willricketts:
Adding a proper reference to GETTING_STARTED.md from README.md. This mislink sent me around and around looking for documentation for the `create_list` method.


tute:
Link goes to http://www.rubydoc.info/gems/factory_girl/file/GETTING_STARTED.md in master, and it works well right now. I wonder if rubydoc was down when you tried? Thanks anyway!

PR #814, opened 2015-09-11
port test suite to rspec3

boutil:
This this a slight change of the specs to be able to run them with RSpec v3.
This commit also bumps the version of rspec in the gemspec file and adds a dependency on rspec-its.


joshuaclayton:
@boutil thanks for starting the work on this! I've merged this as 4a4a0b90f278f48298ef97a12425e1a120cb8abd, with some additional changes including bumping the gemfiles and resolving deprecation messages.

Issue #821, opened 2015-09-24
Ordering of String sequences


jhanggi:
I've now run into this on three separate projects. We commonly use a sequence for creating names for our records (to avoid uniqueness constraints), and we're often testing against a specific ordering of results.

`sequence(:name) { |n| "name #{n}" }`

We'll often have some similar scope:

```
def self.ordered
  order(:name)
end
```

Everything works when the test runs in isolation, but if we run a full test suite, we run into transient test failures where `"name 100" < "name 99"` when the test runs across that three digit/four digit border. We tried using `sequence(:name, "a")`, but that still runs into a similar problem when we cross the 26-item border. We can solve this by changing our sequence to:

`sequence(:name, "aaaaaaa") { |n| "name #{n} }`

giving us millions of options, so we'll likely never hit the border of `zzzzzzzz`, `aaaaaaaaa` (8 "z" vs 9 "a"), but it is theoretically possible.

Is there a smarter way to solve this across the board, or would it be possible to make a default for string-based sequences so if you said `sequence(:name, "a")`, it always started with a long string (10-12 characters)?

We're working with a lot of legacy codebases where FactoryGirl has been in place for a long time, and we're only recently forcing the use of CI in our development process, so these intermittent failures are popping up more and more. 

Maybe the best solution is improved documentation around `sequence` regarding its use in strings. If someone has a recommended place for that documentation, I'm happy to build a PR around it.

joshuaclayton:
@jhanggi I'm going to close this issue, deferring to conversations on each of the other PRs for the various conversations about this.

PR #822, opened 2015-09-24
Add documentation around using alphabetic `sequence`

jhanggi:
Documentation related to Issue #821 where I have run into problems stemming from ordering issues related to alphabetization and sequences.

After looking at the FactoryGirl code, I didn't see clean way to implement a nice way to handle this, I figured documentation would be useful. The explanation is longer than the usual style for the guide, so I am happy to figure out how to trim it down.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -423,6 +423,26 @@ factory :user do
 end
 ```
 
+`sequence` uses `Enumerator#next` to determine the next value, so you can use a
+string as your initial value.
+
+```ruby
+factory :user do
+  sequence(:email, "aaaaaaaa") { |n| "person-#{n}example.com" }
+end
+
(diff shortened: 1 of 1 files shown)


joshuaclayton:
@jhanggi Right [above the traits section](https://github.com/jhanggi/factory_girl/blob/explain_alphabetic_sequences/GETTING_STARTED.md#traits), we discuss alphabetic starting points. I think explaining that the default value can be anything that responds to `next` will work, and then mentioning that strings will increment, makes sense.

I do think that being overly prescriptive about testing patterns, or reiterating how Ruby sorts strings, is out of the scope of this document. If one were to test sorting email addresses in a test, I'd actually recommend they use email address they provide (instead of being auto-generated).

Thoughts?

joshuaclayton:

PR #823, opened 2015-09-25
Mention factory_factory_girl to readme

st0012:
Hello, I use factory_girl so often that I created a gem called [factory_factory_girl](https://github.com/st0012/factory_factory_girl) two months ago. It inherited factory_girl_rails's generator and can help generating factory file more convenient. And I wonder if I can add it's link to readme to let more people know about it, thanks.


joshuaclayton:
@st0012 thanks for pointing this out; however, because we don't maintain that gem, I don't feel comfortable linking to it within the docs right now.

PR #825, opened 2015-10-01
Raise an error when a sequence returns a value that is less than the last value

jhanggi:
The prime example is:

`sequence(:name) { |n| "name#{n}" }`

If your code evers sort by `name` alphabetically, then you could end up
with the case where your sequence returns `name9`, `name10`. The test
might pass in isolation, but when run as part of a full suite, you could
end up with intermittent failures based on the ordering of the specs.

This raises an error if a value is ever less than (`<`) the previous value.
The solution is to change your sequence to `sequence(:name, "00000000")`
(or some longer string if you call the sequence several millions of times).

If you are using an enumerator, this check is skipped.

This could be considered a breaking change, so I am open to adding a configuration setting that issues a deprecation warning instead of raising an error as the default behavior. This addresses my issue #821. I have an alternative that I will issue a PR for that adds a new type of sequence that has this behavior without changing how `sequence` currently works.

### DIFF
--- a/lib/factory_girl/errors.rb
+++ b/lib/factory_girl/errors.rb
@@ -11,6 +11,9 @@ module FactoryGirl
   # Raised when attempting to register a sequence from a dynamic attribute block
   class SequenceAbuseError < RuntimeError; end
 
+  # Raised when the next value in a sequence is < the previous value (e.g. "z" < "aa" or "9" < "10")
+  class SequenceOverflowError < RuntimeError; end
+
   # Raised when defining an invalid attribute:
   # * Defining an attribute which has a name ending in "="
   # * Defining an attribute with both a static and lazy value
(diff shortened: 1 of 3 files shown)


joshuaclayton:
@jhanggi I'm having trouble understanding where this would be a problem, or how tests in isolation would cause this to come up.

Specifically following your email example, if there were an acceptance test which creates more than nine records, ensuring sorting to view users seems entirely tangential to whatever the acceptance test is doing. If you are in fact testing sorting by email, I'd instead be explicit about the email addresses and worry less about the "edge cases" of sorting strings with incremented integers.

Somewhat related to your solution, instead of using a string default, could you instead pad the integers?

```ruby
sequence(:email) { |n| "person-#{"%010d" % n}@example.com" }
```

joshuaclayton:

PR #826, opened 2015-10-01
Add an alphabetic_sequence method

jhanggi:
Use `alphabetic_sequence(:name) { |n| "=#{n}" }` and all values will be
alphabetically sorted. This solves the issue when writing tests that
rely on alphabetical sorting and the factories cross a string length
boundary (e.g. "Z".."AA" ("Z" < "AA") or "9".."10") will often pass
in isolation, but when run in a full test suite might fail intermittently.

If your tests ever reach a point where the next value is `<` the previous
value, an error will be raised.

This new sequence type breaks down if using an Enumerator for the sequence.

This is an alternative to PR #825 addressing issue #821. Perhaps there is some combination of the two that can work well.


jhanggi:
I'm happy to address the houndci feedback if y'all consider the idea sound. I based my style off (specifically around aligning `{}`) of what was there previously.

joshuaclayton:

PR #833, opened 2015-10-20
Fix getting started for minitest

manuelmeurer:
I just tried to include the syntax methods in my project that uses `minitest-rails` and failed. Turns out there is no `MiniTest::Rails::ActiveSupport::TestCase` anymore. It works if I just include the methods in `Minitest::Spec` though, so I guess there is no need for an extra entry for `minitest-rails` anymore.
Also, I changed `MiniTest` to `Minitest`, which is the [preferred syntax](https://github.com/seattlerb/minitest/commit/9a57c520ceac76abfe6105866f8548a94eb357b6) since v5. (`MiniTest` still works but is an alias of `Minitest`).

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -48,18 +48,18 @@ class Spinach::FeatureSteps
   include FactoryGirl::Syntax::Methods
 end
 
-# MiniTest
-class MiniTest::Unit::TestCase
+# Minitest
+class Minitest::Unit::TestCase
   include FactoryGirl::Syntax::Methods
 end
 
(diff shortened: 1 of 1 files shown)


manuelmeurer:
Actually I forgot to add `require 'minitest/rails'` to my `test_helper.rb` (tests ran fine without it). Once it's added, the syntax method must be included in `ActiveSupport::TestCase`, not `Minitest::Spec` to make it work. This is what I added in the last commit.
Sorry for the confusion...

tute:
Thank you! :clap:

PR #836, opened 2015-10-31
Add space, make it easier for Opal to test

wied03:
I know this isn't strictly broken on MRI, Rubinius, etc., but Opal's compiler doesn't like the lack of a space. If you could add it, that would be great. And it looks nice. :+1:


joshuaclayton:
@wied03 thanks catching this! I've merged this as f5e568e47e7bae261e4cd6ad4bbfc13fb9b1f305

Issue #837, opened 2015-11-04
build_stubbed strategy should set changes to empty


randoum:
Greetings,

Currently when creating a record with `build_stubbed` it has `changes` populated as if it was a non persisted object.

```ruby
user = build_stubbed :user, name: 'Joe'
user.changes
# => {"name"=>[nil, "Joe"]}
user.name_was
# => nil
```

As `build_stubbed` is supposed to mimic a persisted object without accessing the database, a correct behaviour would be to have its `changes` empty.
I can simulate this by doing the following:

```ruby
user = build_stubbed :user, name: 'Joe'
user.clear_changes_information
user.changes
# => {}
user.name_was
# => "Joe"
```

So my questions:
- Would you agree with me on the expected behaviour?
- If yes, could you point me to the right direction to implement this in FG source so I can post a PR?

quadule:
I just ran into this today and I agree with your reasoning: I would expect a record returned by `build_stubbed` to be unchanged as it would if it was loaded through ActiveRecord.

Any maintainers care to comment? Thanks!

randoum:
Closing. Related PR on going: #838

PR #838, opened 2015-11-10
Clear changed attributes with build_stubbed strategy

quadule:
Since `build_stubbed` returns an record that appears to be `persisted?`, I would expect there to be no changes tracked by `ActiveModel::Dirty`, but this is not currently the case as documented in #837. 

With these changes, `FactoryGirl::Strategy::Stub` will call `ActiveModel::Dirty#clear_changes_information` after stubbing database interaction on a result.

### DIFF
--- a/lib/factory_girl/strategy/stub.rb
+++ b/lib/factory_girl/strategy/stub.rb
@@ -10,6 +10,7 @@ module FactoryGirl
       def result(evaluation)
         evaluation.object.tap do |instance|
           stub_database_interaction_on_result(instance)
+          clear_changed_attributes_on_result(instance)
           evaluation.notify(:after_stub, instance)
         end
       end
@@ -67,6 +68,21 @@ module FactoryGirl
             end
           end
         end
+      end
+
+      def clear_changed_attributes_on_result(result_instance)
+        unless result_instance.respond_to?(:clear_changes_information)
+          result_instance.extend ActiveModelDirtyBackport
+        end
+
+        result_instance.clear_changes_information
+      end
+    end
+
+    module ActiveModelDirtyBackport
+      def clear_changes_information
+        @previously_changed = ActiveSupport::HashWithIndifferentAccess.new
+        @changed_attributes = ActiveSupport::HashWithIndifferentAccess.new
       end
     end
   end
--- a/spec/acceptance/build_stubbed_spec.rb
+++ b/spec/acceptance/build_stubbed_spec.rb
@@ -51,6 +51,10 @@ describe "a generated stub instance" do
     expect(subject).not_to be_new_record
   end
 
+  it "isn't changed" do
+    expect(subject).not_to be_changed
+  end
+
   it "disables connection" do
     expect { subject.connection }.to raise_error(RuntimeError)
   end


randoum:
@tute could you spare some times to have a look please?

tute:
This change makes sense to me:

```ruby
[1] pry(main)> u = FactoryGirl.build_stubbed(:user)
=> #<User:0x007fa04ac2a348
 id: 1001,
 created_at: nil,
 updated_at: nil,
 first_name: "First 1",
 last_name: "Last 1",
 role: "",
 email: "user_1@example.com",
 encrypted_password: "$2a$10$d6twLh6e7WKI5LPW3guU0uHgIp4jrczErhaOnytmDodHW2LmIKs6O",
 confirmation_token: nil,
 remember_token: nil,
 organization_id: nil,
 admin: false,
 email_confirmation_token: "",
 email_confirmed_at: Sun, 08 Nov 2015 17:05:01 UTC +00:00>
[2] pry(main)> u.changes
=> {"email"=>[nil, "user_1@example.com"], "first_name"=>["", "First 1"], "last_name"=>["", "Last 1"], "encrypted_password"=>[nil, "$2a$10$d6twLh6e7WKI5LPW3guU0uHgIp4jrczErhaOnytmDodHW2LmIKs6O"], "email_confirmed_at"=>[nil, Sun, 08 Nov 2015 17:05:01 UTC +00:00], "id"=>[nil, 1001]}
[5] pry(main)> u.changed?
=> true
```

@joshuaclayton?

spec/factory_girl/strategy/stub_spec.rb
```diff
@@ -18,6 +18,12 @@
     it { expect(subject.result(evaluation)).not_to be_new_record }
     it { expect(subject.result(evaluation)).to be_persisted }
 
+    it "clears changed attributes" do
+      evaluation.object.stubs(:clear_changes_information)
+      subject.result(evaluation)
+      expect(evaluation.object).to have_received(:clear_changes_information).once
```
tute: Is this test necessary?

quadule: Nah, this shouldn't be necessary since there's no actual logic being tested. I'll remove it.

tute:
What do you think of this, @joshuaclayton?

joshuaclayton:
This seems very highly coupled to `ActiveModel::Dirty`; there are now four pieces of data we need to keep in sync with their code (both ivars, and the values being assigned to each).

Are there any internal methods in `ActiveModel::Dirty` that provide us the ability to reset changed state?

lib/factory_girl/strategy/stub.rb
```diff
@@ -68,6 +69,19 @@ def created_at
           end
         end
       end
+
+      def clear_changed_attributes_on_result(result_instance)
+        unless result_instance.respond_to?(:clear_changes_information)
```
joshuaclayton: could we denote this as a "more private" method by using the double-under prefix/suffix? e.g. `:__clear_changes_information__`

joshuaclayton:
@quadule have you checked other strategies to see if they also leave the instances dirty?

quadule:
@joshuaclayton I probably should have left a comment; `clear_changes_information` is the supported method of setting a model as unchanged in `ActiveModel::Dirty`. It was [added in 4.2](https://github.com/rails/rails/commit/66d0a0153578ce760d822580c5b8c0b726042ac2) to replace the `reset_changes` method introduced in 4.1. `ActiveModel::Dirty` has used the same `@previously_changed` and `@changed_attributes` ivars since Rails 3 though.

Instances from the build strategy are dirty like all new records, and instances from the create strategy aren't dirty since they've been saved. Do you want specs covering those?

joshuaclayton:
@quadule:

> Instances from the build strategy are dirty like all new records, and instances from the create strategy aren't dirty since they've been saved. Do you want specs covering those?

No, I don't think it's necessary, and great point - since this is the only one which deviates from expected behavior, let's focus on just updating this.

As for `clear_changes_information` and a comment, in the past I've thrown this behavior in a well-named module and conditionally mixed it in. Thoughts on something like this?

```ruby
def clear_changed_attributes_on_result(result_instance)
  unless result_instance.respond_to?(:clear_changes_information)
    result_instance.extend ActiveModelDirtyBackport
  end

  result_instance.clear_changes_information
end

module ActiveModelDirtyBackport
  def clear_changes_information
    @previously_changed = ActiveSupport::HashWithIndifferentAccess.new
    @changed_attributes = ActiveSupport::HashWithIndifferentAccess.new
  end
end
```

quadule:
I like the module approach, and it [seems to be slightly faster](https://gist.github.com/quadule/8cf281244901dd966a09) too. Does this work now?

joshuaclayton:
Thanks @quadule! I've merged this as 6f7b94b159078dcf0a4c832e42ffc256f90e5253

PR #843, opened 2015-12-07
Find factories registered for namespaced models

### LINKED ISSUE #740: Invoke factory with class name instead of factory name
I've got an ActiveRecord model and a basic factory set up for it, under the normal Rails naming convention:

```ruby
# app/models/panel.rb
class Panel < ActiveRecord::Base
  # model stuff
end

# spec/factories/panels.rb
FactoryGirl.define do
  factory :panel do
    # factory stuff
  end
end
```

I've noticed in my RSpec tests I can invoke this factory with either the factory name _or_ the model's class name:

```ruby
FactoryGirl.create(:panel)

FactoryGirl.create(Panel)
```

The latter is preferable, because it means I can replace literal symbols with RSpec's `described_class` helper.

So now, I've got another model that I'm attempting to do this for, but when invoking the factory, I see:

```ruby
ArgumentError:
  Factory not registered: ModelNamespace::OtherModel
```

I've defined the model and factory as follows:

```ruby
# app/models/model_namespace/other_model.rb
class ModelNamespace::OtherModel < ActiveRecord::Base
  # database table for model is named 'model_namespace_other_models', yay for inheriting legacy systems!
  # model stuff
end

# spec/factories/model_namespace_other_models.rb
FactoryGirl.define do
  factory(:other_model, class: ModelNamespace::OtherModel) do
    # factory stuff
  end
end
```

Is there a way I could redefine either to fix the resolution issue, or am I misusing this feature of FactoryGirl?

monkbroc:
Namespaced models like `Invoice::Item` should be able to be found by `FactoryGirl.create(Invoice::Item)` if registered as `factory :invoice_item`.

I added a test for this new behavior.

Fixes #740

### DIFF
--- a/lib/factory_girl/decorator/class_key_hash.rb
+++ b/lib/factory_girl/decorator/class_key_hash.rb
@@ -19,7 +19,7 @@ module FactoryGirl
         if key.respond_to?(:to_sym)
           key.to_sym
         else
-          key.to_s.underscore.to_sym
+          key.to_s.underscore.gsub("/", "_").to_sym
         end
       end
     end
--- a/spec/factory_girl/registry_spec.rb
+++ b/spec/factory_girl/registry_spec.rb
@@ -65,4 +65,13 @@ describe FactoryGirl::Registry do
     expect(subject.find(:user)).to eq registered_object
     expect(subject.find(User)).to eq registered_object
   end
+
+  it "registers namespaced classes" do
+    define_class("Invoice")
+    define_class("Invoice::Item")
+    subject.register(Invoice::Item, registered_object)
+    expect(subject.to_a).to eq [registered_object]
+    expect(subject.find(:invoice_item)).to eq registered_object
+    expect(subject.find(Invoice::Item)).to eq registered_object
+  end
 end


joshuaclayton:
@monkbroc I'm not sure accessing factories by the class itself is something we want to promote. While I understand some amount of ease with using `described_class` in RSpec, I think it's a way of causing confusion, as there's not a 1:1 mapping between the classes.

```ruby
factory :invoice_item, class: Invoice::Item do
  # base attributes

  factory :awesome_invoice_item do
    # custom attributes to make it awesome
  end
end
```

Both of these will build an `Invoice::Item` instance, but when running `build(:invoice_item)`, which takes precedence, the last-declared or the one with the class? The fact that it's not obvious is concerning.

The other aspect is that there's a tight coupling when using classes (not that there wasn't before, but this change further enables class usage). Given:

```ruby
factory :item, class: Invoice::Item
```

`build(:item)` works correctly, while `build(Invoice::Item)` does not. Because of the conditional nature of this, I do wonder if the more appropriate action would be to deprecate providing a class (to be removed in FG 5.0), and instead require a string or symbol directly matching how the factory key is registered.

I do appreciate you taking the time to solve this, but I do wonder if it's something that should rather be disallowed rather than further enabled. Thoughts?

monkbroc:
I agree that providing fewer variants of the FG API would be better. Deprecating finding a factory using a class is the way to go for FG 5.0. I can submit a PR for this if you'd like (don't know if FG 5 is on the horizon yet).

joshuaclayton:
@monkbroc if you'd like to tackle this, I'd love to see a PR. In the past, we've wrapped this sort of functionality in feature flags and included deprecations to help the upgrade process. Let me see if I can find some older features we'd done this for as examples.

Issue #834, opened 2015-10-28
README.md "Need Help?" link is deprecated 


Holyoke:
`coaching.thoughtbot.com/rails` is a 404. Perhaps Just changing to`coaching.thoughtbot.com`?

tejasbubane:
http://coaching.thoughtbot.com redirects to https://thoughtbot.com

PR #845, opened 2015-12-08
Remove Coaching section from README.md [ci skip]

RochesterinNYC:
Coaching on Rails is no longer available from Thoughtbot.

Addresses Issue #834


RochesterinNYC:
As far as I can tell from searches and Thoughtbot's organization website, coaching is no longer available. Please let me know if I'm mistaken or looking in the wrong places ?

PR #847, opened 2015-12-08
Clarify file used to config Cucumber for Factory Girl syntax

ckib16:
# A PR to help fellow BDD/TDD noobs

## Problem
I could not figure out where to add the designated Cucumber integration code for Factory Girl. Most tutorials seem to assume users will implicitly know this information.

## Solution
I finally found the correct location when visiting this tutorial: https://coderwall.com/p/kxfobg/taking-the-factorygirl-out-of-factorygirl-create.

## Request
Please consider adding this minor update to explicitly clarify which file the code is supposed to be added to. It may help other new developers avoid frustration.

Thanks!
Chris

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -41,7 +41,7 @@ class Test::Unit::TestCase
 end
 
 # Cucumber
-# env.rb (Rails example location - app/features/support/env.rb)
+# env.rb (Rails example location - RAILS_ROOT/features/support/env.rb)
 World(FactoryGirl::Syntax::Methods)
 
 # Spinach
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -41,6 +41,7 @@ class Test::Unit::TestCase
 end
 
 # Cucumber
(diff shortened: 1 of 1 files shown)


GETTING_STARTED.md
```diff
@@ -41,6 +41,7 @@ class Test::Unit::TestCase
 end
 
 # Cucumber
+# env.rb (Rails example location - app/features/support/env.rb)
```
joshuaclayton: including the `app/` prefix here is a bit confusing to me, since that's where models, controllers, etc. live. How do you feel about `RAILS_ROOT` instead?

ckib16: Wow...my bad! Sorry about that. I obviously spent too long looking at code that day...

Yes, that is indeed better. Apologies for the confusion, and thanks for the reply.

ckib16:
Thanks @joshuaclayton  - updated the code per you suggestion. Apologies for the oversight, and thanks for the feedback!

joshuaclayton:
Thanks @ckib16; I've merged this as 3d846f8465a057b8fe7996695491c1c95f53cb1e

PR #854, opened 2016-01-13
Add how to assign values for JSON columns

extrajordanary:
Not 100% sure where the best placement of this info would be, but made sense to me to include it right after lazy attributes since I mention the use of brackets related to that.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -162,6 +162,19 @@ factory :user do
 end
 ```
 
+JSON Properties
+---------------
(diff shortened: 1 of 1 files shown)


joshuaclayton:
@extrajordanary thanks for pointing this out, and I agree, this can be a bit confusing. I've introduced 08c65ebd74dae370513ce57e7a787aae2d0f168b, which is a type-agnostic approach. Thanks!

extrajordanary:
Thanks @joshuaclayton !

PR #856, opened 2016-01-18
Don't package up specs and feature tests

iainbeeston:
Tests aren't useful for end users of the gem, and excluding them from
the gem them reduces the size of the download.

Since bundler/bundler#3207 this has been the default for gems created
using `bundler gem`, and I've taken the code for excluding them from
there.


joshuaclayton:
Thank you @iainbeeston! I've merged this as be67cbd91fa568757a86565ca786a4d74abc752d

PR #873, opened 2016-02-18
Add link to video on Upcase

christoomey:


README.md
```diff
@@ -10,7 +10,10 @@ Documentation
 
 You should find the documentation for your version of factory_girl on [Rubygems](https://rubygems.org/gems/factory_girl).
 
-See [GETTING_STARTED] for information on defining and using factories.
+See [GETTING_STARTED] for information on defining and using factories. We also
+have [a detailed introductory video][], available for free on Upcase.
```
joshuaclayton: @christoomey thoughts on also linking to Upcase for the last word?

christoomey: We already have 2 links in this paragraph, I'd prefer to keep it focused. Also, the goal here isn't so much to sell Upcase, but instead to introduce people to it, via a free video on factory_girl, so I like keeping it focused on that. Does that make sense to you?

joshuaclayton: @christoomey wonderful point, and agreed

joshuaclayton:
@christoomey :ship: :rainbow: :tada:

christoomey:
Per conversation with @joshuaclayton, not worrying about rbx2 failure.

PR #874, opened 2016-02-25
Updated the location of GETTING_STARTED

thoran:
Updated the location of GETTING_STARTED to github.com, since rdoc.info is out of date.


joshuaclayton:
I'm going to close this without merging. The goal of pointing to the documentation on Rubydoc is that it aligns with the latest released version of the gem. In this case, pointing to 4.5.0 is correct. I'll be cutting a new version of the gem today.

PR #877, opened 2016-02-29
Deprecate factory lookup by class

### LINKED ISSUE #871: Disallow classes be used for factory lookup
Based on the discussion in https://github.com/thoughtbot/factory_girl/pull/843, and from other thoughts around simplifying FG's interface, I think we should deprecate the ability to find a factory based on class.

@monkbroc had mentioned interest in deprecating this behavior.

@monkbroc here's an example of a feature flag: https://github.com/thoughtbot/factory_girl/commit/a5b3a97c9d72090c4a0d814a84077634184774a2 and the removal in prep for a major release (along with other changes!): https://github.com/thoughtbot/factory_girl/commit/e9d9e3099b65e08ae18988c8f3f9a39b09a60902

My thought is that we'll want to:

1. Ensure there's an acceptance test around factory lookup by class
2. Introduce a flag disallowing that behavior available in FG's configuration
3. Introduce a test where, with the flag set, it raises a 'factory not found' exception
4. Introduce a deprecation warning when we actually do look up by class
5. Silence deprecation warnings in the appropriate tests

I'm not sure I have an opinion as to whether we also disallow string lookup as well, only allowing symbols, but it's another level of specificity I wouldn't be opposed to introducing.

monkbroc:
This PR deprecates the ability to look up a factory based on class instead of symbol.

- The acceptance test `keyed_by_class_spec.rb` tests lookup by class
- Add flag `allow_class_lookup` (default `true`)
- Show a deprecation on class lookup warning when flag is true
- Raise an exception on class lookup when flag is false
- Silence deprecation warnings in tests using RSpec metadata

Fixes #871

### DIFF
--- a/lib/factory_girl.rb
+++ b/lib/factory_girl.rb
@@ -69,8 +69,10 @@ module FactoryGirl
 
   class << self
     delegate :factories, :sequences, :traits, :callbacks, :strategies, :callback_names,
-      :to_create, :skip_create, :initialize_with, :constructor, :duplicate_attribute_assignment_from_initialize_with,
-      :duplicate_attribute_assignment_from_initialize_with=, to: :configuration
+             :to_create, :skip_create, :initialize_with, :constructor,
+             :duplicate_attribute_assignment_from_initialize_with,
+             :duplicate_attribute_assignment_from_initialize_with=,
+             :allow_class_lookup, :allow_class_lookup=, to: :configuration
   end
 
   def self.register_factory(factory)
--- a/lib/factory_girl.rb
+++ b/lib/factory_girl.rb
@@ -69,10 +69,9 @@ module FactoryGirl
 
   class << self
     delegate :factories, :sequences, :traits, :callbacks, :strategies, :callback_names,
-             :to_create, :skip_create, :initialize_with, :constructor,
-             :duplicate_attribute_assignment_from_initialize_with,
-             :duplicate_attribute_assignment_from_initialize_with=,
-             :allow_class_lookup, :allow_class_lookup=, to: :configuration
+      :to_create, :skip_create, :initialize_with, :constructor,
+      :duplicate_attribute_assignment_from_initialize_with, :duplicate_attribute_assignment_from_initialize_with=,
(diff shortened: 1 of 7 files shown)


lib/factory_girl.rb
```diff
@@ -69,8 +69,10 @@ def self.lint(*args)
 
   class << self
     delegate :factories, :sequences, :traits, :callbacks, :strategies, :callback_names,
-      :to_create, :skip_create, :initialize_with, :constructor, :duplicate_attribute_assignment_from_initialize_with,
-      :duplicate_attribute_assignment_from_initialize_with=, to: :configuration
+             :to_create, :skip_create, :initialize_with, :constructor,
+             :duplicate_attribute_assignment_from_initialize_with,
+             :duplicate_attribute_assignment_from_initialize_with=,
+             :allow_class_lookup, :allow_class_lookup=, to: :configuration
```
joshuaclayton: @monkbroc could you revert the formatting (but keep the new getter/setter)?

monkbroc: Sure. I reformatted it to fix a Hound warning, but I'll put it back.

lib/factory_girl.rb
```diff
@@ -69,8 +69,9 @@ def self.lint(*args)
 
   class << self
     delegate :factories, :sequences, :traits, :callbacks, :strategies, :callback_names,
-      :to_create, :skip_create, :initialize_with, :constructor, :duplicate_attribute_assignment_from_initialize_with,
-      :duplicate_attribute_assignment_from_initialize_with=, to: :configuration
+      :to_create, :skip_create, :initialize_with, :constructor,
+      :duplicate_attribute_assignment_from_initialize_with, :duplicate_attribute_assignment_from_initialize_with=,
```
monkbroc: I reverted the formatting

joshuaclayton:
@monkbroc thanks again for tackling this! I've merged this as e84bfeba9b9859ebf0884c0c47f84dd7ded381e1

monkbroc:
Thanks for the clear direction @joshuaclayton. There was no need for back and forth because you wrote down your expectations. A better use of your time and mine. 👍

PR #886, opened 2016-03-19
Transient attributes in traits

jeffcole:
I had the use case shown in the [committed example code](https://github.com/thoughtbot/factory_girl/commit/1442c9e040923424a8e153eab3796d1d8b919e2b#diff-df17b1aeaf5e7d797809ea45cc1271acR728), where I wanted to DRY up my creation of associated objects with attributes. It wasn't immediately clear from the docs that this was possible, so I thought this might help some people.

This also [removes](https://github.com/thoughtbot/factory_girl/commit/dacc3467c156c7ba5dc6cc756dd47e9c356a32db) an invocation of `FactoryGirl.define` that was otherwise the only such use within the traits section, to keep the example code consistent.

Thanks for this great project!

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -711,16 +711,14 @@ create(:post).author
 Traits can be used within other traits to mix in their attributes.
 
 ```ruby
-FactoryGirl.define do
-  factory :order do
-    trait :completed do
-      completed_at { 3.days.ago }
-    end
+factory :order do
+  trait :completed do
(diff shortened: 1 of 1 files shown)


joshuaclayton:
Merged as 972c71d1e449851972a23746d45ed0bf319cec5a and 07ece7c4169a3c3d9adce4471b3f790c460af7a4; thanks @jeffcole!

PR #891, opened 2016-04-01
FG 4.7.0

joshuaclayton:


NEWS
```diff
@@ -1,3 +1,12 @@
+4.7.0 (April 1, 2016)
+  Improve documentation
+  Improve instrumentation payload to include traits, overrides, and the factory itself
+  Allow linting of traits
+  Deprecate factory lookup by class name in preparation for 5.0
+  Improve internal performance by using flat_map instead of map and compact
+  Improve handling of dirty attributes after building a stubbed object
+  Reduce warnings from redefining methods
+
```
mike-burns: Can you mention `4.6.0` in here, for the sake of those debugging later?

Something as simple as:

```
4.6.0 skipped
```

calleerlandsson:
Looks good to me.

NEWS
joshuaclayton: Great point! Fixed in bd5838d91778872b5e017edecadcb4d356b9b02a

tute: Why did we skip 4.6?

PR #897, opened 2016-04-04
Don't test against old versions of Ruby and Rails

joshuaclayton:


tute:
This can be closed?

joshuaclayton:
@tute this was for FG 5, not Rails 5. There's some other things I'd like to explore before releasing FG 5.

composerinteralia:
I think we may be able to drop support for Ruby 2.2 at this point as well. It doesn't seem to officially be EOL, but the scheduled EOL date was last March, so I expect it will be done by the time a release is ready. I can take over getting this branch updated.

Also I would love to chat at some point about what you were thinking of including in FB 5.

composerinteralia:
Ruby 2.2 officially unsupported https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/

PR #905, opened 2016-05-13
Allow factories to be linted with build

danrarr:
Useful for objects that do not support `#save`


joshuaclayton:
@danrarr You can just declare those factories with `skip_create`: https://github.com/thoughtbot/factory_girl/blob/master/GETTING_STARTED.md#custom-methods-to-persist-objects

PR #908, opened 2016-05-19
Add error message when count is missing for create_list

### LINKED ISSUE #889: create_list without the count should give a proper error message
Sometimes I forget to pass in the count attribute for `create_list` (eg. when I change `create` to `create_list` and I forget to add the count):
```ruby
create_list(:user, profile: profile)
```

The error in this case is:
```
Failure/Error: create_list(:user, profile: profile)

     NoMethodError:
       undefined method `times' for #<Hash:0x007fda3bd50738>
# /Users/tejas/.rvm/gems/ruby-2.3.0@oudux/gems/factory_girl-4.5.0/lib/factory_girl/strategy_syntax_method_registrar.rb:28:in `block in define_list_strategy_method'
```

I feel this error should be a little more elaborate saying something like the `create_list` method expects a count.

FactoryGirl version in use here: `4.5.0`

tejasbubane:
Closes #889

`create_list` was failing abruptly if the second argument was not count. Give a descriptive error message for this case.

### DIFF
--- a/lib/factory_girl/strategy_syntax_method_registrar.rb
+++ b/lib/factory_girl/strategy_syntax_method_registrar.rb
@@ -25,6 +25,10 @@ module FactoryGirl
       strategy_name = @strategy_name
 
       define_syntax_method("#{strategy_name}_list") do |name, amount, *traits_and_overrides, &block|
+        unless amount.respond_to?(:times)
+          raise ArgumentError, "count missing for #{strategy_name}_list"
+        end
+
         amount.times.map { send(strategy_name, name, *traits_and_overrides, &block) }
       end
     end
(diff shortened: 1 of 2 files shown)


lib/factory_girl/strategy_syntax_method_registrar.rb
```diff
@@ -25,6 +25,10 @@ def define_list_strategy_method
       strategy_name = @strategy_name
 
       define_syntax_method("#{strategy_name}_list") do |name, amount, *traits_and_overrides, &block|
+        unless amount.is_a?(Integer)
```
drapergeek: What do you think about asking for `respond_to(:times)` here instead? It's probably a minor difference but we don't care what the type is really, only that it responds to the `times` method.

tejasbubane:
@drapergeek Fixed as per your comment.

drapergeek:
Thanks @tejasbubane

PR #909, opened 2016-05-20
Fix warning: `&' interpreted as argument prefix

czuger:
This fix a small warning. Didn't create test, because the modification is trivial.


joshuaclayton:
Thanks @czuger, I've merged as 330646ac15fb28aaf6f732da339e2bdce33e1fc6

PR #915, opened 2016-06-13
Adds require line in order to configure test suite successfully

ivancortesromero:
When one runs `rspec` with spec which uses `FactoryGirl` with Rails 5 and Ruby 2.3.1, `uninitialized constant FactoryGirl (NameError)` is raised, because when it attemps to include `FactoryGirl::Syntax::Methods`, `FactoryGirl` has not been required at top of the file.


ivancortesromero:
@jeffcole

PR #917, opened 2016-06-14
replaced inject with each_with_object to reduce line

humzashah:
`each_with_object` doesn't require the hash to be returned at the end of each iteration. this enables us to save a line.

the speed difference between `inject` and `each_with_object` will most likely be inconsequential in this scenario.


joshuaclayton:
@humzashah thanks! I've merged this as 0424c5a1744c72af8526201ef55d6551262aaf08

PR #922, opened 2016-06-23
Disable all ActiveRecord::Persistence methods in Stub strategy

urbanautomaton:
Was: "Don't allow `#save!` to be called on stubbed instances"

The `Stub` strategy is used to build instances that appear to be persisted, but don't write to the database. The stubs created have a [number of methods overwritten](https://github.com/thoughtbot/factory_girl/blob/07ece7c4169a3c3d9adce4471b3f790c460af7a4/lib/factory_girl/strategy/stub.rb#L36-L66) to prevent subsequent interaction with the database.

However, `#save!` is not included, meaning that it can be called on stubs without error (as can other methods like `#touch` that eventually call `#save!`).

Adding it to the blacklist is a simple change, but I'd like to check that its omission wasn't intentional. I can't find any specific reference to it in the history so I assume it's just an oversight - can anyone who's worked on the `Stub` strategy think of any reason it might've been deliberately left out?

Another thing worth considering: this change will probably break a few test suites out there. Is the benefit of consistency worth the downside of having a breaking change?

### DIFF
--- a/lib/factory_girl/strategy/stub.rb
+++ b/lib/factory_girl/strategy/stub.rb
@@ -4,15 +4,29 @@ module FactoryGirl
       @@next_id = 1000
 
       DISABLED_PERSISTENCE_METHODS = [
-        :save,
-        :destroy,
         :connection,
-        :reload,
-        :update_attribute,
-        :update_column,
+        :decrement!,
+        :decrement,
+        :delete,
+        :destroy!,
+        :destroy,
+        :destroyed?,
         :increment!,
-        :decrement!
-      ]
+        :increment,
+        :reload,
+        :save!,
+        :save,
+        :toggle!,
+        :toggle,
+        :touch,
+        :update!,
+        :update,
+        :update_attribute,
+        :update_attributes!,
+        :update_attributes,
+        :update_column,
+        :update_columns,
+      ].freeze
 
       def association(runner)
         runner.run(:build_stubbed)
--- a/lib/factory_girl/strategy/stub.rb
+++ b/lib/factory_girl/strategy/stub.rb
@@ -3,6 +3,17 @@ module FactoryGirl
     class Stub
       @@next_id = 1000
 
+      DISABLED_PERSISTENCE_METHODS = [
+        :save,
+        :destroy,
+        :connection,
+        :reload,
+        :update_attribute,
+        :update_column,
+        :increment!,
+        :decrement!
+      ]
+
       def association(runner)
         runner.run(:build_stubbed)
       end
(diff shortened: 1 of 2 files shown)


joshuaclayton:
@urbanautomaton interesting - I was fairly sure this was captured by disallowing `save`; does `save!` not call `save` behind the scenes?

urbanautomaton:
@joshuaclayton Thanks for the response. As it turns out, no - both `#save` ([link](https://github.com/rails/rails/blob/e8c21ff645eefb889acd7799bd12c9757cd8fd8c/activerecord/lib/active_record/persistence.rb#L124-L128)) and `#save!` ([link](https://github.com/rails/rails/blob/e8c21ff645eefb889acd7799bd12c9757cd8fd8c/activerecord/lib/active_record/persistence.rb#L151-L153)) are independently defined in `ActiveRecord::Persistence` using `#create_or_update`. (They each have a chain of `super` calls in front of this, but these are the definitions that handle the actual writes.)

Looking at it a bit more closely, there may be a few other methods in `Persistence` that aren't being blacklisted - I'm away now for a few days but can have a proper look when I get back, assuming a fuller blacklist is something you'd be happy to merge.

joshuaclayton:
@urbanautomaton I'd definitely be interested - I'd recently had to track one of these down IIRC, and it's not fun.

Might be good to see if we can maintain a list in both the implementation and tests (which we already have) and dynamically generate the methods. At this point, there's a fair bit of redundancy in the implementation.

Take your time, no rush on moving this forward right this second!

urbanautomaton:
I've updated the branch - hopefully the commits should be explanatory, but the main changes are:

* Refactored the specs to use a shared example group
* Refactored the disabled methods to de-duplicate the overrides, as suggested
* Expanded the blacklist to cover all of `ActiveRecord::Persistence`

This means that as of Rails 5.0, the blacklist will cover any instance methods that might interact with the database. It also matches the interface of the Rails 4.2-stable and 4.1-stable branches.

It implements some methods that 3.2-stable does not (`#destroy!`, `#update` and `#update!`). It also will not catch any new persistence methods introduced in new versions of Rails. I'd welcome feedback on whether this is a problem or not.

We could potentially address it by using `ActiveRecord::Persistence.public_instance_methods` to only override the methods defined by the current version of Rails, e.g.

```ruby
DISABLED_PERSISTENCE_METHODS = ActiveRecord::Persistence.public_instance_methods - [
  :persisted?,
  :new_record?,
  :becomes,
  :becomes!
]
```

This would involve extra complexity that I'm unsure is warranted, though (and at present there's no explicit dependency on ActiveRecord for the stub strategy). What do you think?

joshuaclayton:
@urbanautomaton This is looking awesome!

To address some of your comments/questions:

> It implements some methods that 3.2-stable does not (#destroy!, #update and #update!). It also will not catch any new persistence methods introduced in new versions of Rails. I'd welcome feedback on whether this is a problem or not.

This may introduce very-far-edge-cases where developers define these methods for use in 3.2 apps, but I'd guess they'd do so to mirror behavior in later releases. If it does come up, we can look to see how to address it, but my guess is that this won't have any impacts.

As for not catching new persistence methods, that seems fine to me - specifically, I think we'll want to opt-in to checks and be explicit about it.

I think avoiding a dependency on `ActiveRecord::Persistence` makes sense right now.

I could imagine a refactor (either in this PR, or sometime later if you want to get this merged) where this list is configured at the `FactoryGirl::Configuration` layer, and then developers could swap the provided list with `ActiveRecord::Persistence`'s instance methods to stay up-to-date (or, if they're using a different ORM/ODM, use that tool to dictate which methods are stubbed). That would allow for much easier testing (as we could inject a controlled list), and possibly remove the `shared_examples` block (as we'd only need to test one or two methods).

urbanautomaton:
@joshuaclayton: thanks for the feedback!

> I could imagine a refactor (either in this PR, or sometime later if you want to get this merged) where this list is configured at the `FactoryGirl::Configuration` layer

Yep, that makes sense - another option could be to maintain a blacklist per major Rails version (e.g. `DISABLED_PERSISTENCE_METHODS::RAILS_3` etc.), and conditionally swap them in by detecting the loaded rails version.

Given that the interface seems relatively static at this point, though (it's unchanged in rails 5, so I'd anticipate no further additions for at least a couple of years), I'd be tempted to suggest waiting and seeing what happens before adding the extra complexity. What do you think?

joshuaclayton:
@urbanautomaton love it. Thank you so much for your work on this; this part of the codebase has driven me batty for a while, and the implementation here feels so much more straightforward now.

This has been merged as fd3bf889f8ba33ba8cd0a69fd891a1b0a0bae5e8. Thanks again!

urbanautomaton:
Thanks @joshuaclayton! 🎉

PR #924, opened 2016-07-01
Appraise against Rails 5.0

joshuaclayton:


teoljungberg:
Seems straightforward.

PR #927, opened 2016-07-07
exclude rubies that rails 5.0 does not support

mathieujobin:
was looking at your travis and saw these allowed failures.
you can actually exclude build for unsupported version.

I'm not sure if rbx-2 should be able to run rails 5, so I left that one under allowed_failures

also bump minor versions of recent rubies

cheers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 rvm:
   - 1.9.3
   - 2.0.0
-  - 2.1.0
-  - 2.2.2
-  - 2.3.0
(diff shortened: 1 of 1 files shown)


mathieujobin:
any reason against merging this ?

joshuaclayton:
https://github.com/thoughtbot/factory_girl/commit/2fa1bf876b61805feaa4b9204976e09714699c66 looks to address this; if there are further changes, could you make them in a separate PR?

Thanks @mathieujobin!

PR #928, opened 2016-07-11
lockdown the activesupport version to <5

goughy000:
activesupport v5 is out now with a dependency on Ruby 2.2.2

This causes factory_girl gem to fail install with error "Gem::InstallError: activesupport requires Ruby version >= 2.2.2."  Even though "required_ruby_version" of factory_girl is ">= 1.9.2"

Can either do this, or alternatively up the "required_ruby_version" of this gem

Don't mind either way :)


joshuaclayton:
@goughy000 what version of factory_girl did you attempt to install?

I just spun up a new Rails 5.0.0 app with Ruby 2.3.1, added `factory_girl_rails` to the gemfile under the development and test groups, and it bundled correctly.

goughy000:
I think I might be confused, but here is what happened...
```
jack$ rvm gemset use fg_test
Using ruby-2.2.1 with gemset fg_test

jack$ gem install bundler
Fetching: bundler-1.10.6.1.gem (100%)
Successfully installed bundler-1.10.6.1
Parsing documentation for bundler-1.10.6.1
Installing ri documentation for bundler-1.10.6.1
Done installing documentation for bundler after 3 seconds
1 gem installed

jack$ bundle update --full-index
Fetching source index from https://rubygems.org/
Resolving dependencies..........
Installing concurrent-ruby 1.0.2
Installing i18n 0.7.0
Installing minitest 5.9.0
Installing thread_safe 0.3.5
Using bundler 1.10.6.1
Installing tzinfo 1.2.2
Installing activesupport 5.0.0 (was 4.2.6)

Gem::InstallError: activesupport requires Ruby version >= 2.2.2.
An error occurred while installing activesupport (5.0.0), and Bundler cannot continue.
Make sure that `gem install activesupport -v '5.0.0'` succeeds before bundling.

jack$ cat Gemfile
source 'https://rubygems.org/'

gem 'factory_girl'

```

danielkza:
@joshuaclayton Ruby >= 2.2.2 will work fine, as activesupport will still work. Any older version of Ruby will break. Restricting the gem version to <5 will disallow factory_girl from being used in Rails 5 projects.

joshuaclayton:
@goughy000 I'd commented https://github.com/thoughtbot/factory_girl/issues/929 on how to resolve; the short answer is that you need to lock (or install) AS 4.2.6 prior to FG.

PR #930, opened 2016-07-12
Reprioritize how to install and add caveat for AS 5.0

### LINKED ISSUE #929: Gem cannot be installed in Rubies older than 2.2.2 due to activesupport
```
$ ruby --version
ruby 2.1.8p440 (2015-12-16 revision 53160) [x86_64-linux]
$ gem install factory_girl
ERROR:  Error installing factory_girl:
	activesupport requires Ruby version >= 2.2.2.
```

This happens because activesupport 5 was released without support for older rubies. I'm seeing the same issue in older gems and have not found a good solution: restricting the maximum version will break compatibility with Rails 5 projects. It seems the only sane workaround is using a fake native extension that just installs the correct activesupport version depending on the Ruby version being used.

joshuaclayton:
FactoryGirl installation should occur through Bundler, so it can resolve
dependencies correctly.

When installing manually, include a caveat when installing on older
Rubies on how to handle breakages in the README.

Resolves #929, Resolves #928

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,15 +18,32 @@ have [a detailed introductory video][], available for free on Upcase.
 Install
 --------
 
-```shell
-gem install factory_girl
-```
-or add the following line to Gemfile:
+Add the following line to Gemfile:
 
 ```ruby
 gem 'factory_girl'
 ```
+
 and run `bundle install` from your shell.
+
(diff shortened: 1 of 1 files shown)


mike-burns:
LGTM.

PR #935, opened 2016-08-06
Moved inheritance above associations in the Getting Started guide

victornamuso:
I was reading through the Getting Started guide and noticed that Inheritance is introduced in the Associations section which precedes the Inheritance section. Since the Inheritance section is small I think it flows better to introduce it first, then it makes more sense when you encounter it in the Associations section.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -288,6 +288,42 @@ end
 
 ```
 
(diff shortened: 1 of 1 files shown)


joshuaclayton:
@victornamuso nice catch; thanks for making this change!

Issue #954, opened 2016-10-13
Set updated_at column on model when using build_stubbed


coorasse:
I didn't see a previous issue regarding that and I'd like to open a discussion about this point.
When using `build_stubbed` to generate a model which "pretends" to be persisted on the database the column `updated_at` is `nil`. build_stubbed correctly sets the `created_at` column but not the `updated_at`. When you really persist an object on DB (in Rails) also the column `updated_at` is set so I'd expect it to have a value.

joshuaclayton:
@coorasse that's an excellent point, and agreed that there should be consistency.

Want to open up a PR testing and implementing this behavior? If you need help getting started, please let me know!

prasadsurase:
I am using factory_girl(4.7.0) with mongoid(5.1.4). When using build_stubbed, both created_at and updated_at aren't populated and are nil.

```ruby
2.3.0 :002 > r = FactoryGirl.build_stubbed :round
 => #<Round _id: 57ff92195f10de2aa59114b3, created_at: nil, updated_at: nil, name: "Y-find", from_date: 2016-10-12 21:27:49 UTC, end_date: 2016-11-11 23:47:46 UTC, status: "open"> 
2.3.0 :003 > r.persisted?
 => true 
2.3.0 :004 > r.new_record?
 => false 
```

schmijos:
👍

PR #955, opened 2016-10-13
Add default updated_at column

coorasse:
Fixes issue #954 .

### DIFF
--- a/lib/factory_girl/strategy/stub.rb
+++ b/lib/factory_girl/strategy/stub.rb
@@ -75,6 +75,18 @@ module FactoryGirl
             end
           end
         end
+
+        has_updated_at = result_instance.respond_to?(:updated_at)
+        updated_at_no_default = has_updated_at && !result_instance.updated_at
+        result_instance_missing_updated_at = !has_updated_at
+
+        if updated_at_no_default || result_instance_missing_updated_at
+          result_instance.instance_eval do
+            def updated_at
+              @updated_at ||= Time.current
+            end
+          end
+        end
       end
 
       def clear_changed_attributes_on_result(result_instance)
--- a/spec/acceptance/build_stubbed_spec.rb
+++ b/spec/acceptance/build_stubbed_spec.rb
@@ -154,6 +154,41 @@ describe "defaulting `created_at`" do
   end
 end
 
+describe "defaulting `updated_at`" do
+  include FactoryGirl::Syntax::Methods
+
+  before do
+    define_model("ThingWithTimestamp", updated_at: :datetime)
+    define_model("ThingWithoutTimestamp")
+
+    FactoryGirl.define do
+      factory :thing_with_timestamp
+      factory :thing_without_timestamp
+    end
+
+    Timecop.freeze 2012, 1, 1
+  end
+
+  it "defaults updated_at for objects with updated_at" do
+    expect(build_stubbed(:thing_with_timestamp).updated_at).to eq Time.current
+  end
+
+  it "adds updated_at to objects who don't have the method" do
+    expect(build_stubbed(:thing_without_timestamp)).to respond_to(:updated_at)
+  end
+
+  it "allows overriding updated_at for objects with updated_at" do
+    stubbed = build_stubbed(:thing_with_timestamp, updated_at: 3.days.ago)
+    expect(stubbed.updated_at).to eq 3.days.ago
+  end
+
+  it "doesn't allow setting updated_at on an object that doesn't define it" do
+    expect do
+      build_stubbed(:thing_without_timestamp, updated_at: Time.now)
+    end.to raise_error(NoMethodError, /updated_at=/)
+  end
+end
+
 describe 'defaulting `id`' do
   before do
     define_model('Post')


coorasse:
Hello @joshuaclayton . Do I have to fix Houd comments?

lib/factory_girl/strategy/stub.rb
```diff
@@ -75,6 +75,28 @@ def created_at
...
         end
+
+        responds_to_updated_at = result_instance.respond_to?(:updated_at)
+        updated_at_missing_default = responds_to_updated_at && !result_instance.updated_at
+        result_instance_missing_updated_at = !responds_to_updated_at
+
+        if updated_at_missing_default || result_instance_missing_updated_at
+          result_instance.instance_eval do
+            def updated_at
+              @updated_at ||= Time.now.in_time_zone
```
joshuaclayton: Thoughts on using `Time.current`, which is Rails' standard that takes timezones into account?

spec/acceptance/build_stubbed_spec.rb
```diff
@@ -150,7 +150,53 @@
   end
 
   it "doesn't allow setting created_at on an object that doesn't define it" do
-    expect { build_stubbed(:thing_without_timestamp, :created_at => Time.now) }.to raise_error(NoMethodError, /created_at=/)
+    expect { build_stubbed(:thing_without_timestamp, :created_at => Time.now) }
```
joshuaclayton: Same comment here re: `Time.now` -> `Time.current`

spec/acceptance/build_stubbed_spec.rb
```diff
@@ -150,7 +150,53 @@
...
+    FactoryGirl.define do
+      factory :thing_with_timestamp
+      factory :thing_without_timestamp
+    end
+
+    Timecop.freeze 2012, 1, 1
+  end
+
+  it "defaults updated_at for objects with updated_at" do
+    expect(build_stubbed(:thing_with_timestamp).updated_at).to eq Time.now
```
joshuaclayton: `Time.now` -> `Time.current`?

lib/factory_girl/strategy/stub.rb
```diff
@@ -75,6 +75,28 @@ def created_at
...
+            end
+          end
+        end
+      end
+
+      def add_missing_attribute(result_instance, column_name)
```
joshuaclayton: Looks like this isn't used anywhere - should this get removed?

lib/factory_girl/strategy/stub.rb
```diff
@@ -75,6 +75,28 @@ def created_at
             end
           end
         end
+
+        responds_to_updated_at = result_instance.respond_to?(:updated_at)
+        updated_at_missing_default = responds_to_updated_at && !result_instance.updated_at
+        result_instance_missing_updated_at = !responds_to_updated_at
+
+        if updated_at_missing_default || result_instance_missing_updated_at
```
joshuaclayton: This brings up an interesting situation - why would we add `updated_at` to an object that doesn't already have that method defined?

spec/acceptance/build_stubbed_spec.rb
```diff
@@ -150,7 +150,53 @@
...
+  end
+
+  it "defaults updated_at for objects with updated_at to the correct time with zone" do
```
joshuaclayton: I think by switching to `Time.current` everywhere, we can remove this

lib/factory_girl/strategy/stub.rb
coorasse: Nice question. I obviously kept the behaviour consistent with created_at behaviour. Let me know what would you prefer. We would have to make them consistent anyway.

lib/factory_girl/strategy/stub.rb
coorasse: sorry. my bad. forgot here.

coorasse:
I would prefer to replace `Time.current` for created_at in a separate PR.

coorasse:
Back to you @joshuaclayton . There are a couple of points which I found difficult in this PR and I'd like to feedback you:
* Hound only runs in the PR: I cannot simulate it locally (or I don't know how to do that, so I need to commit and push to see if Hound will accept it)
* Line length set to 80 is very short IMHO (I would suggest to set it to 120 now that the project doesn't respect it anyway)
* We need an editor config file so I have the preferences set correctly (for example for the line length)

I'd glad to keep helping. I'll make another PR to use `Time.current` also in `created_at`.
Thanks

coorasse:
Hei there @joshuaclayton , any update on this one? I was waiting for a feedback. Thanks!

joshuaclayton:
@coorasse sorry for the late reply. Noted on the line length, although I think we should stick with 80 as a general rule of thumb, and opt to "violate" if we want. In that regard, I think this should be set. Mind squashing into one commit in preparation for merge?

coorasse:
squashed and ready

joshuaclayton:
@coorasse thank you for putting this together; excited to have this merged!

PR #957, opened 2016-10-18
refer to gem itself consistently

brandondrew:
Everywhere else the gem is referred to as "factory_girl" instead of "FactoryGirl".


joshuaclayton:
Thanks for this, @brandondrew!

PR #960, opened 2016-10-24
Fixed typos

Lechindianer:


GETTING_STARTED.md
```diff
@@ -142,13 +142,13 @@ Using factories
 factory\_girl supports several different build strategies: build, create, attributes\_for and build\_stubbed:
 
 ```ruby
-# Returns a User instance that's not saved
+# Returns an User instance that's not saved
```
joshuaclayton: This (and all subsequent) case of `a User` (or `a user`) is actually correct (hooray English grammar!) - mind changing these back?

composerinteralia:
Updated and merged as b53c5addc792b39aa32ba71488c4cdc5381c988a

PR #964, opened 2016-11-16
Small clarifying updates to comments

tinabme:
Super small update to the comments. It just happened to trip me up for a split second, so I thought I'd offer a quick change. Plus it makes it consistent with the other @param descriptions.   :)


joshuaclayton:
Looks great, thanks for this @tinabme!

PR #967, opened 2016-11-24
method for generating list of values in sequence

rutaka-n:
I use FactoryGirl in my work project and at one day I needed a list of phone numbers from sequence and I wrote `generate_list(:phone_numbers, 10)` but FactoryGirl has not this method. I decided that method might be useful for me and somebody else.

Thank you for your time to review my PR.


joshuaclayton:
Thanks for putting this together @rutaka-n. I like the consistency!

PR #973, opened 2016-12-21
Update Ruby versions

tricknotes:


composerinteralia:
Updated and merged as 116d36780c5023be3c7d522940b7ea423a020694. Thanks!

PR #978, opened 2016-12-28
Add note about factories in transient attributes

mltsy:
Using a factory as the default value for a transient attribute causes the test environment to fail if the table representing the factory used is not yet present in the database, because FactoryGirl tries to evaluate the default value upon loading, which checks for an existing model(?), and if present, tries to extract information about it from the database.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -250,6 +250,7 @@ factory :user do
   transient do
     rockstar true
     upcased  false
+    github_group { FactoryGirl.create(:github_group) }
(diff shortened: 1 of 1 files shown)


composerinteralia:
We are going to deprecate static attributes with https://github.com/thoughtbot/factory_bot/pull/1135, which will make this documentation update no longer necessary. Thanks for your help!

mltsy:
Whoa!  Didn't see that coming.  Alright, cool :smile: :+1:

PR #982, opened 2017-01-15
Stub out ActiveRecord::Base#destroyed?

### LINKED ISSUE #981: Disabled persistence method doesn't access the database
Using activerecord 4.2.7.1 and factory_girl 4.8.0.  I upgraded from a previous version of activerecord and factory_girl and a number of our specs started failing.

When passing an instance created using the stub strategy as a relation to an instance created with the create strategy, and saving the persisted instance, factory girl raises an error on the `#destroyed?` method.

e.g.:

```
foo = build_stubbed(:foo)
bar = create(:bar, foo: foo)
bar.save!
```
This breaks the spec.

`#destroyed?` is called in `ActiveRecord::AutosaveAssociation#save_belongs_to_association`.  I'm fine with the notion of preventing access to the database in stubbed models, but I don't think `#destroyed?` [actually calls the database](https://github.com/rails/rails/blob/75f274c8e42e9df3eeebb557583e1aafd2396047/activerecord/lib/active_record/persistence.rb#L91-L94).

Note that it has a [similar implementation to `#persisted?`](https://github.com/rails/rails/blob/75f274c8e42e9df3eeebb557583e1aafd2396047/activerecord/lib/active_record/persistence.rb#L98-L101) which is actually stubbed out in `FactoryGirl::Strategy::Stub`.  It calls `ActiveRecord::Base#sync_with_transaction_state` and then returns a value.  `#sync_with_transaction_state` does not appear to actually access the database in any way.

I'm wondering if I'm either 1) offbase, or 2) just shouldn't be doing what I'm doing in the first place.  If not, would it be possible to remove `#destroyed?` from the list of disabled persistence methods?

bradleypriest:
See #981 for more details

FactoryGirl 4.8.0 updated some logic to make sure to raise on all ActiveRecord::Persistence methods.

However the `destroyed?` method seems like it is easily stubbed on the Stub object and allows the ActiveRecord::AutosaveAssociation#save_belongs_to_association method to work with stubbed objects again

### DIFF
--- a/lib/factory_girl/strategy/stub.rb
+++ b/lib/factory_girl/strategy/stub.rb
@@ -10,7 +10,6 @@ module FactoryGirl
         :delete,
         :destroy!,
         :destroy,
-        :destroyed?,
         :increment!,
         :increment,
         :reload,
@@ -56,6 +55,10 @@ module FactoryGirl
 
           def new_record?
             id.nil?
+          end
+
+          def destroyed?
+            nil
           end
 
           DISABLED_PERSISTENCE_METHODS.each do |write_method|
--- a/spec/factory_girl/strategy/stub_spec.rb
+++ b/spec/factory_girl/strategy/stub_spec.rb
@@ -36,6 +36,7 @@ describe FactoryGirl::Strategy::Stub do
 
     it { expect(subject.result(evaluation)).not_to be_new_record }
     it { expect(subject.result(evaluation)).to be_persisted }
+    it { expect(subject.result(evaluation)).not_to be_destroyed }
 
     it "assigns created_at" do
       created_at = subject.result(evaluation).created_at
@@ -52,7 +53,6 @@ describe FactoryGirl::Strategy::Stub do
     include_examples "disabled persistence method", :delete
     include_examples "disabled persistence method", :destroy
     include_examples "disabled persistence method", :destroy!
-    include_examples "disabled persistence method", :destroyed?
     include_examples "disabled persistence method", :increment
     include_examples "disabled persistence method", :increment!
     include_examples "disabled persistence method", :reload


Joeper214:
I hope will be merged soon.

Jakanapes:
Hit this exact issue when I tried to upgrade the FG gem.  Hoping to see this merged soon.

joshuaclayton:
@urbanautomaton Given this was introduced in https://github.com/thoughtbot/factory_girl/commit/fd3bf889f8ba33ba8cd0a69fd891a1b0a0bae5e8, do you have any insight you could provide here? I've confirmed that `:destroyed?` is in fact present within `ActiveRecord::Persistence.public_instance_methods`, so I'm interested in your thoughts on whether this is the appropriate solution, or if you're aware of other impacts here.

Thanks in advance!

urbanautomaton:
@joshuaclayton Yep, this change makes total sense to me - it's consistent with the existing stubbing and I can't think of any problems it would cause. 👍

joshuaclayton:
Excellent, all set; I've merged this as 168bbf8e346487973d2b50496ed2d93100f83276.

leonelgalan:
@joshuaclayton any plans on releasing this as a new version? 4.8.1 perhaps?

cgriego:
I also ran into this regression with 4.8.0 as well. Mongoid calls `destroyed?` as a part of its ActiveModel `to_key` method, so our decorator secs that use stubbed factories now have a bunch of errors. Any plans for a release?

TheRusskiy:
@joshuaclayton @jferris sorry to ping you guys directly, but that's an easy and very important fix

joshuaclayton:
@TheRusskiy @cgriego @leonelgalan all set - 4.8.1 is up and published on Rubygems!

PR #983, opened 2017-01-28
Improve GETTING_STARTED table of content hierarchy

daguar:
Hello!

The current GETTING_STARTED file has a somewhat confusing table of contents, with the specific names of each test framework as big top-level content, and then the main content (like "Defining factories") as small bullets nested below "minitest-rails":

screen shot 2017-01-28 at 9 38 26 am

This PR makes the test suite names sub-headers of "configure your test suite" and makes what I think is the main content top-level headers:

screen shot 2017-01-28 at 9 36 15 am

I think this just brings the file more in line with the intent, but happy to discuss if I'm wrong about that!

(Also note, this will have to land on Rubygems to be live on the documentation site, since the link to GETTING_STARTED in the README is hard-coded.)

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -28,7 +28,7 @@ Once your Gemfile is updated, you'll want to update your bundle.
 Configure your test suite
 -------------------------
 
-# RSpec
+### RSpec
 
 ```ruby
 # spec/support/factory_girl.rb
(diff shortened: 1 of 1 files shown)


rlue:
:100: This has been bothering me for forever. Any chance we could get this merged?

PR #984, opened 2017-02-01
Update lint example to configure DatabaseCleaner

thorncp:
Not specifying `strategy`/`clean_with` can lead to test data sticking around on CI and cause unique validations to fail while the lint process is running.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -987,6 +987,8 @@ namespace :factory_girl do
   task lint: :environment do
     if Rails.env.test?
       begin
+        DatabaseCleaner.strategy = :transaction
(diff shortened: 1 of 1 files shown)


GETTING_STARTED.md
```diff
@@ -987,6 +987,8 @@ namespace :factory_girl do
   task lint: :environment do
     if Rails.env.test?
       begin
+        DatabaseCleaner.strategy = :transaction
```
composerinteralia: Do we need to specify the strategy? It looks like DatabaseCleaner [tries to infer this](https://github.com/DatabaseCleaner/database_cleaner/blob/14626296ef31b34f7704593667db1d41d81825ab/lib/database_cleaner/base.rb#L179)

PR #991, opened 2017-03-03
Allow instance to be accessed from evaluator

kwstannard:
This will allow for the following:

```ruby
factory :post do
  user
end

factory :user do
  posts { build_list(:post, 2, user: instance) }
end
```

I think this is an improvement because it allows for accessing association objects prior to callbacks, which I feel makes for writing factories that are easier to follow.


kwstannard:
Actually, I have a similar but better idea that will actually use this. I am going to open a new PR some time this week.

PR #994, opened 2017-03-12
Allow selective linting of single factories

dcarral:
Before these changes: `FactoryGirl.lint(FactoryGirl.factories.first)` raises error, since `.lint` expected an Enumerable (`FactoryGirl::Registry` by default).

This change allows to pass single factories to be linted, without the need of adding them to an enumerable collection:

Before: `FactoryGirl.lint([FactoryGirl.factories.first])`
With this change, simply: `FactoryGirl.lint(FactoryGirl.factories.first)`

Next step to achieve greater linting flexibility would be to allow selective linting by factory names, instead of full factory objects.

### DIFF
--- a/lib/factory_girl/linter.rb
+++ b/lib/factory_girl/linter.rb
@@ -2,7 +2,7 @@ module FactoryGirl
   class Linter
 
     def initialize(factories_to_lint, linting_strategy)
-      @factories_to_lint = factories_to_lint
+      @factories_to_lint = force_enumerable(factories_to_lint)
       @linting_method = "lint_#{linting_strategy}"
       @invalid_factories = calculate_invalid_factories
     end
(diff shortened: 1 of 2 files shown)


joshuaclayton:
@dcarral thanks for submitting this!

I'd like to keep the interface complexity down with `lint`, so instead of accepting multiple types of arguments, I think only allowing a list is the right way to go. If a consumer of the library knows only one is going to be returned, manually wrapping in an array should suffice.

dcarral:
@joshuaclayton Thanks a lot for getting back to me soon. Would you folks be interested in allowing selective linting by factory names, instead of full factory objects?

joshuaclayton:
@dcarral for the same reason I'd like to avoid having FG handle the array adapter aspect, I'd instead encourage developers to use Ruby for that too:

```ruby
factories_to_lint = FactoryGirl.factories.select do |factory|
  factory.name.to_s =~ /user|post|comment/
  # or
  %i(user post comment).include? factory.name
end

FactoryGirl.lint(factories_to_lint)
```

dcarral:
Ok, got it. It's a pity though :( 

Thanks for the info and keep up the good work.

PR #995, opened 2017-03-16
Loosen up the stub behavior to work nicely in Rails 5

gkop:
During the migration from Rails 4.2.7 to 5.0.2, I ran into an issue where ActiveRecord tried to autosave my stubbed associations, and called `#destroyed?`. This patch resolved my issue.

I'm not sure what the correct solution is, but thought to open this up in case I'm not the only one that runs into this.

### DIFF
--- a/lib/factory_girl/strategy/stub.rb
+++ b/lib/factory_girl/strategy/stub.rb
@@ -10,7 +10,6 @@ module FactoryGirl
         :delete,
         :destroy!,
         :destroy,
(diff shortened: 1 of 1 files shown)


buehmann:
Duplicate of https://github.com/thoughtbot/factory_girl/pull/982

joshuaclayton:
This behavior was merged in a separate commit: 168bbf8e346487973d2b50496ed2d93100f83276

PR #997, opened 2017-03-17
Do not raise error for increment and decrement in stubs.

michaelkoper:
Right now an error is raised when increment or decrement is called. But those methods don't access the database, increment! and decrement! do. 

I think factory girl should allow these methods.

https://github.com/thoughtbot/factory_girl/issues/996

### DIFF
--- a/lib/factory_girl/strategy/stub.rb
+++ b/lib/factory_girl/strategy/stub.rb
@@ -6,13 +6,11 @@ module FactoryGirl
       DISABLED_PERSISTENCE_METHODS = [
         :connection,
         :decrement!,
-        :decrement,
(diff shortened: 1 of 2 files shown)


joshuaclayton:
@michaelkoper thanks for pushing this up! I just merged another commit, 168bbf8e346487973d2b50496ed2d93100f83276, which addresses `destroyed?` and is conflicting with your branch; mind rebasing to be up-to-date?

Also, can you update the test (`spec/factory_girl/strategy/stub_spec.rb`) to reflect these changes as well?

michaelkoper:
I think it should be good now @joshuaclayton

Issue #968, opened 2016-11-25
Inconsistent and brittle trait inheritance behavior


lordofthelake:
Given this this class and this factory definition:

```ruby
DataClass = Struct.new(:x, :y, :z)

FactoryGirl.define do
  factory :parent_factory, class: DataClass do
    x 1

    trait :y do
      y 1
    end

    trait :z do
      y
      z 1
    end
  end

  factory :child_factory, class: DataClass, parent: :parent_factory do
    x 2

    trait :y do
      y 2
    end
  end
end
```

You observe this behaviour:
```ruby
require 'factory_girl'
FactoryGirl.find_definitions

FactoryGirl.build(:parent_factory, :z)
# => <struct DataClass x=1, y=1, z=1>
FactoryGirl.build(:child_factory, :z)
# => <struct DataClass x=2, y=1, z=1>

FactoryGirl.reload
FactoryGirl.build(:child_factory, :z)
# => <struct DataClass x=2, y=2, z=1>
[8] pry(main)> FactoryGirl.build(:parent_factory, :z)
# => <struct DataClass x=1, y=2, z=1>

FactoryGirl::VERSION
# => '4.7.0'
```

composerinteralia:
This has to do with the order in which things get ["compiled"](https://github.com/thoughtbot/factory_bot/blob/3a4d6f489ccd3bb58674e470c3bfaad642888975/lib/factory_bot/definition.rb#L46) by factory_bot. Part of the compilation process involves [copying over trait definitions](https://github.com/thoughtbot/factory_bot/blob/3a4d6f489ccd3bb58674e470c3bfaad642888975/lib/factory_bot/definition.rb#L50-L53) so we can refer to one trait within another.

In the first example we compile `parent_factory` first, which involves copying the `y` trait over from `parent_factory` to `z`. We later compile `child_factory` and copy that`child_factory`s `y` trait over to `z` as well, but because the `parent_factory`'s `y` was copied over first https://github.com/thoughtbot/factory_bot/blob/3a4d6f489ccd3bb58674e470c3bfaad642888975/lib/factory_bot/definition.rb#L118 always finds that one.

In the second example we copy over `child_factory`s `y` trait first, and so that is the one we will always find.

As mentioned in https://github.com/thoughtbot/factory_bot/pull/1064#issuecomment-448075022, which I think is rather closely related, this is one of the more complex parts of factory_bot, and straightening it out will be no simple task!

composerinteralia:
[...]

PR #999, opened 2017-03-31
Issue #968: Inconsistent and brittle trait inheritance behavior

rbuetzer:
So far, this is a regression test serving as a starting point for anyone audacious enough to tackle Issue #968.


composerinteralia:
I'm going to close this for now, since we have #968 to track the issue. Thanks for your help! If we end up using this spec we will be sure to give you credit.

PR #1001, opened 2017-03-31
More intuitive attribute-association setting

kwstannard:
When building a factory, self in attribute blocks feels like it should
somehow be related to the instance. This commit allows self to be passed
like the instance. This allow for a more fluid definition of factories
in which you can pass self, and then on the other side access attributes
like they already exist.

In the specs you can see that the comment factory uses the post title,
despite the post title being in a delayed execution block which I think
is a huge deal for enabling simpler factories and could allow for 95% of
callback related code to be deleted.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -408,13 +408,11 @@ FactoryBot.define do
         posts_count 5
       end
 
-      # the after(:create) yields two values; the user instance itself and the
-      # evaluator, which stores all values from the factory, including transient
-      # attributes; `create_list`'s second argument is the number of records
+      # within the  the dynamic association block, pass self as the user and
+      # the post factory will know to associate with the generated user.
+      # `create_list`'s second argument is the number of records
       # to create and we make sure the user is associated properly to the post
-      after(:create) do |user, evaluator|
-        create_list(:post, evaluator.posts_count, user: user)
-      end
+      posts { build_list(:post, posts_count, user: self) }
     end
   end
 end
@@ -458,14 +456,11 @@ FactoryBot.define do
         languages_count 5
       end
 
-      # the after(:create) yields two values; the profile instance itself and
-      # the evaluator, which stores all values from the factory, including
-      # ignored attributes; `create_list`'s second argument is the number of
-      # records to create and we make sure the profile is associated properly
-      # to the language
-      after(:create) do |profile, evaluator|
-        create_list(:language, evaluator.languages_count, profiles: [profile])
-      end
+      # within the  the dynamic association block, pass [self] as the users and
+      # the post factory will know to associate with the generated user.
+      # `create_list`'s second argument is the number of records
+      # to create and we make sure the user is associated properly to the post
+      languages { build_list(:language, languages_count, profiles: [self]) }
     end
   end
 end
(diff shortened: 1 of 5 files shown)


pluff:
@kwstannard you can simply use `@instance` in your factory definition

kwstannard:
@pluff This is more than `@instance`. `@instance` would not give access to the data in the builder and if you tried to use it in the tests I wrote they would fail.

kwstannard:
@drapergeek Is the reason no maintainers have touched this PR because of something I did wrong? I have been checking back every couple months since I opened this.

drapergeek:
@kwstannard I apologize for the delay and appreciate your contribution.

The maintainers of the project seem to be backed up on reviews at the moment. I wish I could provide more insight but I'm not one of the maintainers and don't have enough context to properly review your code.

joshuaclayton:
@kwstannard sorry for the delay on this!

This is looking good - I'm wondering, however, if we can keep `instance` unavailable and instead only rely on `self`; that way, it's a smaller interface (and arguably more intuitive!). Thoughts?

mike-burns:
Should this have a documentation change?

lib/factory_girl/attribute_assigner.rb
```diff
@@ -53,7 +53,16 @@ def build_hash
     end
 
     def get(attribute_name)
-      @evaluator.send(attribute_name)
+      case obj = @evaluator.send(attribute_name)
+      when FactoryGirl::Evaluator
+        obj.instance
```
kwstannard: @joshuaclayton I am not sure how we could avoid `attr_reader :instance` and still have this work. The writer could be made private though.

kwstannard:
@mike-burns Yes. Is it just GETTING_STARTED.md?

mike-burns:
@kwstannard yes, `GETTING_STARTED.md`.

kwstannard:
@drapergeek Thanks. I was just worried since other PRs were getting merged that I was missing something.

kwstannard:
@mike-burns updated. I was debating on the choice of deleting the old after_create style. I couldn't articulate a good reason to keep it myself so I just went with removing it for now.

GETTING_STARTED.md
```diff
@@ -408,13 +408,11 @@ FactoryBot.define do
         posts_count 5
       end
 
-      # the after(:create) yields two values; the user instance itself and the
-      # evaluator, which stores all values from the factory, including transient
-      # attributes; `create_list`'s second argument is the number of records
+      # within the  the dynamic association block, pass self as the user and
```
mike-burns: Repeated "the".

mike-burns:
Cool.

I think the only reason to keep the old docs is because the old style will still be in codebases and people might want to know what it does.

lib/factory_bot/attribute_assigner.rb
```diff
@@ -53,7 +53,16 @@ def build_hash
     end
 
     def get(attribute_name)
-      @evaluator.send(attribute_name)
+      case obj = @evaluator.send(attribute_name)
```
joshuaclayton: Thoughts on extracting instance access from mapping over the array? e.g.

```ruby
def extract_instance(obj)
  if obj.is_a?(FactoryBot::Evaluator)
    obj.instance
  else
    obj
  end
end

def get(attribute_name)
  obj = @evaluator.send(attribute_name)

  if obj.is_a?(Array)
    obj.map { |o| extract_instance(o) }
  else
    extract_instance(obj)
  end
end
```

lib/factory_girl/attribute_assigner.rb
joshuaclayton: Right - `instance` (reader) wasn't public previously. I think with the change you've included to call `instance`, a public writer and a protected reader might be a good compromise. Thoughts?

lib/factory_bot/attribute_assigner.rb
kwstannard: I forget if there was a reason I didn't extract already. Maybe I was waiting for a third repitition?

kwstannard:
> I think the only reason to keep the old docs is because the old style will still be in codebases and people might want to know what it does.

There are still plenty of references to the callback system including its own section.

lib/factory_bot/attribute_assigner.rb
kwstannard: I will say, in the ideal world, we would delete the hard check in rails on object type and do duck typing instead, which would negate the need for this if statement.

kwstannard: After thinking for a few days, I want to leave it as is for now as I think the extraction makes the code slightly more confusing and it is unknown if this will be useful in the future.

kwstannard:
> Right - instance (reader) wasn't public previously. I think with the change you've included to call instance, a public writer and a protected reader might be a good compromise. Thoughts?

Just saw this. This call occurs from within a different class so a protected reader wouldn't work either. Is there something specific that you are worried about? Instance is already basically public since it gets passed to callbacks.

kwstannard:
Going to assume this is never going to get merged. If anyone likes this idea, give my makery gem a shot and leave some issues or something: https://github.com/kwstannard/makery

asadakbar:
I would love if this could get another look. It would clean up a lot of callback code.

PR #1002, opened 2017-04-05
Avoid id collisions with build_stubbed

victormours:
Hi there!

Thanks for building FactoryGirl. :) We've recently had failing builds because of id collisions between records created with `FactoryGirl.create` and `FactoryGirl.build_stubbed`.

Without changing the way ids are generated by the stub strategy, this should keep us safe for a while. :)

### DIFF
--- a/lib/factory_girl.rb
+++ b/lib/factory_girl.rb
@@ -84,6 +84,7 @@ module FactoryGirl
              :allow_class_lookup=,
              :use_parent_strategy,
              :use_parent_strategy=,
+             :build_stubbed_starting_id=,
              to: :configuration
   end
 
--- a/lib/factory_girl/configuration.rb
+++ b/lib/factory_girl/configuration.rb
@@ -33,5 +33,9 @@ module FactoryGirl
     def duplicate_attribute_assignment_from_initialize_with=(value)
       ActiveSupport::Deprecation.warn 'Assignment of duplicate_attribute_assignment_from_initialize_with is unnecessary as this is now default behavior in FactoryGirl 4.0; this line can be removed', caller
     end
+
+    def build_stubbed_starting_id=(starting_id)
+      Strategy::Stub.next_id = starting_id - 1
+    end
   end
 end
--- a/lib/factory_girl/strategy/stub.rb
+++ b/lib/factory_girl/strategy/stub.rb
@@ -39,6 +39,10 @@ module FactoryGirl
         end
       end
 
+      def self.next_id=(id)
+        @@next_id = id
+      end
+
       private
 
       def next_id
--- a/spec/acceptance/build_stubbed_spec.rb
+++ b/spec/acceptance/build_stubbed_spec.rb
@@ -206,3 +206,18 @@ describe 'defaulting `id`' do
     expect(FactoryGirl.build_stubbed(:post, id: 12).id).to eq 12
   end
 end
+
+describe "configuring the starting id" do
+  before do
+    define_model("Post")
+
+    FactoryGirl.define do
+      factory :post
+    end
+  end
+
+  it "defines which id build_stubbed instances start with" do
+    FactoryGirl.build_stubbed_starting_id = 3000
+    expect(FactoryGirl.build_stubbed(:post).id).to eq 3000
+  end
+end


joshuaclayton:
While I'm not opposed to this from a code modification standpoint, I'm honestly interested in the issue that you were running into, mostly to try to understand why this hasn't come up before. Since there are collisions, does this indicate that you're using `build_stubbed` very infrequently, but when you are, it's after a total of over 1000 records have been created during the test suite run? Is there padding of the initial id for a given record that's persisted, or perhaps a different incrementing scheme?

Thanks in advance!

victormours:
> does this indicate that you're using build_stubbed very infrequently, but when you are, it's after a total of over 1000 records have been created during the test suite run?

Yes, that's pretty much what happened. We have a test suite for a Rails monolith (a car rental marketplace) that has about 4500 examples, and our CI runs them in four parallel containers that run about 1100 examples each.

We use `FactoryGirl.create` quite liberally (1898 instances in our suite), and just like most Rails applications, most of our records are linked to the `users` table in one way or an other, sometimes creating multiple users for a single record (i.e. an order).
We only use `build_stubbed` on our `users` table in only 74 instances, but some of them started raising one of our validation errors when a trying to book a car, since a user can't rent their own car. We checked the ids, and realized that the user created with `build_stubbed` had the id 1003, which was colliding with the id of the user created along with the car that we needed for that particular spec.

So yeah, nothing really special going on code-wise, we just have a lot of tests, and started running into this issue recently.
For now we added `FactoryGirl::Strategy::Stub.class_variable_set(:@@next_id, 100_000)` to our test setup, but we'd rather use an official fix, and save the headache to other people that might run into this as well (and also we don't mind the street cred that comes with contributing 1 line to a famous gem 😉).

I hope that helps. :)

joshuaclayton:
@victormours thanks for the explanation!

Given all of that, what are your thoughts on making a few more changes and having the starting value be part of the FG configuration but defaulting to the current value `1_000`? While 100k is large, we also felt like 1k was a pretty large starting ID as well, so we'd be future-proofing in this commit 😄 

Thoughts?

victormours:
Yep, that's an great idea. :) I'll push an update doing that in the days to come.

Thanks again for your review!

victormours:
Yep, it's a good idea to add the option in the config. However, when this issue popped up, our main pain point was understanding where it came from, and then finding a (hacky) fix was rather easy.

May I suggest raising the default to `2_000`? This way much fewer teams will run into the issue, while still being a reasonable number.

Regardless of the default, I'll push an update to add the start value to the configuration in the days to come. :-)

Thanks again for your review!

jcw-:
FYI, ran into this as well (it's not that unlikely to have 1000 instances of something in a test database that has a relationship to a mocked object) - we're planning to hack `@@next_id` in the same way for now.

I'd suggest using a larger number by default and letting it be configured back to 1000 if for some reason there is a dependency on mocked id's (there really shouldn't be), otherwise it's really just a matter of time before you run into obscure test failures.

composerinteralia:
I think one reason I have never come across this is that I don't tend to use `build_stubbed` and `create` in the same spec. I tend to use `build_stubbed` only for unit tests that don't touch the database at all.

The setting proposed in this PR may help in some cases, but even with a very large starting id it is still possible to run into this problem if using Spring. When using Spring the database ids do not reset back to 1 from one test run to another. If last `user` record created in one test run had id 2000, the first `user` record of the next test run will have id 2001.

odlp:
I'd be happy to help get this PR across the line, if you're happy for me to do so @victormours? 

Looks like we need to:

- [ ] Rebase & resolve conflicts
- [ ] Increase the default starting point as suggested by Josh in https://github.com/thoughtbot/factory_bot/pull/1002#issuecomment-300171275 (shall we say 100k?)

I think starting at a far greater starting value might help a little bit to distinguish 'real' IDs from autoincrement vs build-stubbed IDs - although it doesn't completely resolve the Spring issue:

> When using Spring the database ids do not reset back to 1 from one test run to another;

oliverswitzer:
also ran into an issue with this recently! Ideally we'd only be using `build_stubbed` in all of our tests but unfortunately with our 10 y/o codebase its a bit of a tall order.

Is there any chance that we'd be able to provide the option to have build_stubbed read from the underlying database's auto-incremented id?

composerinteralia:
Given that this is a minimal code change and seems like it will help a number of folks, I rebased on master, moved a couple things around, added some documentation, and merged as https://github.com/thoughtbot/factory_bot/commit/65e64ec2a313c67e5f0c63c4afb96234f045b92e.

Having build_stubbed read from the underlying database's auto-incremented id is a neat idea, but I think it would at a lot of additional complexity to the library and to some extent would defeat the purpose of `build_stubbed` (i.e. a build strategy that avoids hitting the database), so I don't plan to move forward with that idea.

PR #1009, opened 2017-05-23
Simplify DatabaseCleaner code example for linter

oboxodo:
The `cleaning` method already ensures cleaning and provides for a way to write better looking code.

This should be 100% equivalent to previous example.


joshuaclayton:
Nice find, thanks @oboxodo!

PR #1013, opened 2017-05-25
update Ruby and Rails in travis integration

takkanm:


composerinteralia:
It looks like we ended up doing this work in a few other commits. Thanks for your contribution!

PR #1017, opened 2017-06-06
Failing spec for trait with same name as factory

josephks:
After upgrading from 2.6.4 → 4.8.0 we started getting a "Self-referencing association" error.  The problem is that the trait in question had the same name as a factory, as this spec illustrates.

### DIFF
--- a/spec/acceptance/traits_spec.rb
+++ b/spec/acceptance/traits_spec.rb
@@ -725,3 +725,36 @@ describe "traits used in associations" do
     expect(creator.name).to eq 'Joe Creator'
   end
 end
+
+describe "associations set in traits" do
+  before do
+    define_model("Person", name: :string)
+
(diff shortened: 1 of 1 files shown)


spec/acceptance/traits_spec.rb
```diff
@@ -725,3 +725,36 @@ def initialize(name)
...
+
+      factory :bob, parent: :person do
+        name "Bob"
+      end
+
+      factory :book do
+        association :author, factory: :person
+      end
+
+      trait :bob do
```
rocket-turtle: I stumbled over this pull request and was wondering whether a trait can be defined outside of a factory.

josephks: It may not be good practice but it can be.

composerinteralia:
Oops, that commit should have said `closes #1221` instead of `#1017`.

PR #1018, opened 2017-06-14
Fix test order dependent failures with seeds 50350 + 48550

odlp:
I noticed a couple of tests fail intermittently whilst setting up the project to contribute. 

Using `rspec --bisect` to hunt down the minimal reproductions:

## Seed 50350

```
bundle exec rspec './spec/acceptance/define_child_before_parent_spec.rb[1:1]' './spec/factory_girl/definition_spec.rb[3:1]' --seed 50350
``` 

```

Run options: include {:ids=>{"./spec/acceptance/define_child_before_parent_spec.rb"=>["1:1"], "./spec/factory_girl/definition_spec.rb"=>["3:1"]}}

Randomized with seed 50350
.F

Failures:

  1) defining a child factory before a parent creates admin factories correctly
     Failure/Error: declarations.attributes

     Mocha::ExpectationError:
       unexpected invocation: #<Mock:declaration list>.attributes()
     # ./lib/factory_girl/definition.rb:47:in `compile'
     # ./lib/factory_girl/definition.rb:127:in `aggregate_from_traits_and_self'
     # ./lib/factory_girl/definition.rb:38:in `constructor'
     # ./lib/factory_girl/configuration.rb:22:in `constructor'
     # ./lib/factory_girl.rb:71:in `constructor'
     # ./lib/factory_girl/definition_hierarchy.rb:8:in `constructor'
     # ./lib/factory_girl/factory.rb:136:in `compiled_constructor'
     # ./lib/factory_girl/factory.rb:37:in `run'
     # ./lib/factory_girl/factory_runner.rb:29:in `block in run'
     # ./lib/factory_girl/factory_runner.rb:28:in `run'
     # ./lib/factory_girl/strategy_syntax_method_registrar.rb:20:in `block in define_singular_strategy_method'
     # ./spec/acceptance/define_child_before_parent_spec.rb:19:in `block (2 levels) in <top (required)>'

Finished in 0.19463 seconds (files took 0.40745 seconds to load)
2 examples, 1 failure
```

## Seed 48550

```
bundle exec rspec './spec/acceptance/register_strategies_spec.rb[3:2:1]' './spec/factory_girl/definition_spec.rb[2:1]' --seed 48550
```

```
Run options: include {:ids=>{"./spec/acceptance/register_strategies_spec.rb"=>["3:2:1"], "./spec/factory_girl/definition_spec.rb"=>["2:1"]}}

Randomized with seed 48550
.F

Failures:

[...]

PR #1019, opened 2017-06-16
Linting factory strategy

odlp:
When using FactoryGirl with non-ActiveRecord objects it would be useful to allow the strategy used for linting to be specified:

```ruby
FactoryGirl.lint strategy: :build
```

---

~~Based on #1018 to get a green build. Will point base branch to `master` if the other PR is resolved.~~

### DIFF
--- a/lib/factory_girl.rb
+++ b/lib/factory_girl.rb
@@ -59,12 +59,14 @@ module FactoryGirl
   # Parameters:
   # factories - which factories to lint; omit for all factories
   # options:
-  #   traits : true - to lint traits as well as factories
+  #   traits: true - to lint traits as well as factories
+  #   strategy: :create - to specify the strategy for linting
   def self.lint(*args)
     options = args.extract_options!
     factories_to_lint = args[0] || FactoryGirl.factories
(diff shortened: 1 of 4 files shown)


lib/factory_girl.rb
```diff
@@ -59,12 +59,14 @@ def self.reset_configuration
...
   # options:
-  #   traits : true - to lint traits as well as factories
+  #   traits: true - to lint traits as well as factories
+  #   strategy: :create - to specify the strategy for linting
   def self.lint(*args)
     options = args.extract_options!
     factories_to_lint = args[0] || FactoryGirl.factories
-    strategy = options[:traits] ? :factory_and_traits : :factory
-    Linter.new(factories_to_lint, strategy).lint!
+    linting_strategy = options[:traits] ? :factory_and_traits : :factory
```
odlp: Both this new variable and the previous `strategy` are somewhat confusing because it makes be think of the ["build strategy"](http://www.rubydoc.info/gems/factory_girl/file/GETTING_STARTED.md#Using_factories).

Would `linting_scope` or `linting_approach` be good alternative names?

PR #1021, opened 2017-06-23
Allow opt-in of warning on deprecating static attributes

joshuaclayton:
What?
=====

Static attributes can introduce odd or unexpected behavior, largely because
the value is evaluated once, when the factory definition is loaded. Most
commonly, issues can come up with dates:

```ruby
factory :post do
  published_on Date.yesterday
end
```

This may cause problems based on the time of day the suite is running,
especially if there's already a mystery guest introduced (e.g. the test
callsite assumes `published_on` was always one day ago).

To address this, the first step is letting developers know where their test
suites refer to static attributes. This commit introduces a configuration
option developers can opt into where we trigger a deprecation warning with
the factory name and static attributes.

To opt into this setting:

```ruby
# spec/support/factory_girl.rb
FactoryGirl.configuration.warn_on_static_attributes = true
```


jferris:
If we think this is a good idea, should it be opt out instead of opt-in? And then we remove static attributes in the next major release?

composerinteralia:
See #1135

PR #1025, opened 2017-07-10
Update GETTING_STARTED.md

recepinanc:
Change wrong use of proposition.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -46,7 +46,7 @@ RSpec.configure do |config|
 end
 ```
 
-Remember to require the above file in your rails_helper since the support folder isn't eagerly loaded
+Remember to require the below file in your rails_helper since the support folder isn't eagerly loaded
 
 ```ruby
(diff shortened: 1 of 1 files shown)


GETTING_STARTED.md
```diff
@@ -46,7 +46,7 @@ RSpec.configure do |config|
 end
 ```
 
-Remember to require the above file in your rails_helper since the support folder isn't eagerly loaded
+Remember to require the below file in your rails_helper since the support folder isn't eagerly loaded
```
mxie: I see what you mean here, but we do want `above` in this context. I think what we're trying to say here is something like:

> Remember to `require` the above file since the `support` folder isn't automatically loaded for specs. You can do so by adding the following line in your `rails_helper.rb`:

Does that clarify the intent here?

composerinteralia:
Thanks for your contribution! See 8ad9cb1d716016c18c7fb349b32e474bb0aa4d94 for why I am closing this.

Issue #974, opened 2016-12-21
Throw error when trait was already defined


Dorian:
e.g.:

```
FactoryGirl.define  do
  factory :tree do
    size 2

    trait :big do
      size 3
    end

    trait :big do
      size 2
    end
  end
end
```

For instance, should throw an error (I ended up with some confusing code because of that).

(I'm ready to do a PR for that, just wanted to have it posted as an issue meanwhile)

Epigene:
Alternatively, repeat definitions could __override__ the trait with a warning message.
Currently FactoryGirl is silent about duplicates and uses the first definition, which is _unruby_.

Dorian:
@Epigene I prefer raising an error but feel free to go from my PR to implement the warning

Epigene:
Ideally strategy used should be configurable.
Thanks for putting together a PR, I'll see if I can contribute anything in the following days.

Epigene:
I've put together a PR that implements the proposed additions @ https://github.com/thoughtbot/factory_bot/pull/1064

Since I spend only an hour getting my head around how factory_bot works and found few options for introspecting what traits are defined in a performant manner (currenly my code loops over a factory's defined traits whenever a new one is added, bad performance in terms of Big O), the proposed changes are probably rough around the edges.

Please let me know if I can refactor things.

PR #1050, opened 2017-10-19
Add failing spec for duplicate traits

Dorian:
- [x] add failing spec for duplicate traits
- [ ] fix the spec

From #974


composerinteralia:
I am going to close this for now and continue tracking the issue in #974. Thanks for your help! If we end up using this spec we will be sure to give you credit.

PR #1051, opened 2017-10-20
Replace 'girl' with 'bot' everywhere

Euraldius:
Also: add a deprecation warning to factory_girl, asking users to switch to factory_bot

https://github.com/thoughtbot/factory_girl/issues/921


mike-burns:
I don't know how to review this. Can I safely assume that you programmatically renamed everything?

It looks like you have too many gemspecs.

joshuaclayton:
@mike-burns you're correct; for all non-NEWS references, we updated the name. Additionally, there's a new gemspec file for factory_bot so we can create a new *.gem file.

As for the abundance of lock files, that's due to Appraisal.

README.md
```diff
@@ -1,16 +1,18 @@
...
+factory_bot is a fixtures replacement with a straightforward definition syntax, support for multiple build strategies (saved instances, unsaved instances, attribute hashes, and stubbed objects), and support for multiple factories for the same class (user, admin_user, and so on), including factory inheritance.
 
-If you want to use factory_girl with Rails, see
-[factory_girl_rails](https://github.com/thoughtbot/factory_girl_rails).
+If you want to use factory_bot with Rails, see
+[factory_bot_rails](https://github.com/thoughtbot/factory_bot_rails).
+
+**A historical note:** factory_bot used to be named factory_girl. An explanation of the name change can be found in the [old factory_girl repository](https://github.com/thoughtbot/factory_girl).
 
 _[Interested in the project name?](NAME.md)._
```
pablobm: I think this line (and the associated file) needs to be removed. Maybe replace with an explanation of the name change, which would also tackle my previous comment?

README.md
```diff
@@ -1,16 +1,18 @@
...
 
-factory_girl is a fixtures replacement with a straightforward definition syntax, support for multiple build strategies (saved instances, unsaved instances, attribute hashes, and stubbed objects), and support for multiple factories for the same class (user, admin_user, and so on), including factory inheritance.
+factory_bot is a fixtures replacement with a straightforward definition syntax, support for multiple build strategies (saved instances, unsaved instances, attribute hashes, and stubbed objects), and support for multiple factories for the same class (user, admin_user, and so on), including factory inheritance.
 
-If you want to use factory_girl with Rails, see
-[factory_girl_rails](https://github.com/thoughtbot/factory_girl_rails).
+If you want to use factory_bot with Rails, see
+[factory_bot_rails](https://github.com/thoughtbot/factory_bot_rails).
+
+**A historical note:** factory_bot used to be named factory_girl. An explanation of the name change can be found in the [old factory_girl repository](https://github.com/thoughtbot/factory_girl).
```
pablobm: The old respository links to the new one, so this ends up being a link to the current page.

README.md
```diff
@@ -50,35 +52,37 @@ manually.
 Supported Ruby versions
 -----------------------
 
-The factory_girl 3.x+ series supports MRI Ruby 1.9. Additionally, factory_girl
+The factory_bot 3.x+ series supports MRI Ruby 1.9. Additionally, factory_bot
 3.6+ supports JRuby 1.6.7.2+ while running in 1.9 mode. See [GETTING_STARTED]
 for more information on configuring the JRuby environment.
 
-For versions of Ruby prior to 1.9, please use factory_girl 2.x.
+For versions of Ruby prior to 1.9, please use factory_bot 2.x.
```
bquorning: All changes to this “Supported Ruby versions” section are a bit confusing, since there is currently only a 4.x release of factory_bot.

README.md
mustmodify: That link now redirects from factory_girl to factory_bot... I'd be interested in knowing why it was renamed? I'm hoping it is something other than "girls shouldn't work in factories."

mike-burns: Might be nice to have a link to https://robots.thoughtbot.com/factory_bot in here somewhere.

mustmodify:
I understand this decision has already been made and it's your code, do what you want. But just to play devil's advocate for a second, if Elon Musk is right and true general AI becomes a thing, might this not make machines feel uncomfortable? 'Bot' could easily be seen as pejorative.

PR #1057, opened 2017-10-21
Add transition from factory_girl instructions

Euraldius:


README.md
```diff
@@ -9,6 +9,19 @@ If you want to use factory_bot with Rails, see
...
+
+Transitioning is a two step process:
+
+1) Replace `factory_girl` with `factory_bot` in your Gemfile and `bundle install`
+
+2) Rename all `FactoryGirl` to `FactoryBot` and `factory_girl` to `factory_bot` in your codebase.
+
+This command will perform a `FactoryGirl` to `FactoryBot` rename within your `spec/` directory:
+
+```ag -0l FactoryGirl spec | xargs -0 sed -i 's/FactoryGirl/FactoryBot/g'```
```
mxie: Since there are some people who don't have Silver Searcher, maybe we should use a common tool like `grep`. Alternatively, we can say something like "If you have a search tool like Silver Searcher, this command will...`.

README.md
```diff
@@ -9,6 +9,19 @@ If you want to use factory_bot with Rails, see
 
 _[Interested in the project name?](NAME.md)._
 
+
+### Transitioning from factory\_girl?
+
+Transitioning is a two step process:
+
+1) Replace `factory_girl` with `factory_bot` in your Gemfile and `bundle install`
```
mxie: Add a period at the end.

joshuaclayton:
There's a "living" version of the upgrade document (so any additional changes on the branch will be reflected) available at https://github.com/thoughtbot/factory_bot/blob/4-9-0-stable/UPGRADE_FROM_FACTORY_GIRL.md; thoughts on linking there directly instead?

Euraldius:
@joshuaclayton makes sense! I've made that change

PR #1058, opened 2017-10-22
Allow timestamp assignment in stubbed models

alextsui05:
This is in response to the following issue: https://github.com/thoughtbot/factory_bot_rails/issues/233

Consider the following code snippet:

```
model = build_stubbed(:some_model)
timestamp = 1.month.ago
model.created_at = timestamp
expect(model.created_at).to eq(timestamp)
```

FactoryBot will currently fail because the stub only installs a `created_at` method but not a `created_at=` method. As a result, the stub instance variable doesn't get updated.

Incidentally, `timestamp` still gets assigned to the ActiveRecord attribute - it just gets masked by FactoryBot's `created_at`.

### DIFF
--- a/lib/factory_bot/strategy/stub.rb
+++ b/lib/factory_bot/strategy/stub.rb
@@ -76,6 +76,10 @@ module FactoryBot
             def created_at
               @created_at ||= Time.now.in_time_zone
             end
+
+            def created_at=(date)
+              @created_at = date
+            end
           end
         end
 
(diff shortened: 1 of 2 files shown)


lib/factory_bot/strategy/stub.rb
```diff
@@ -76,6 +76,10 @@ def destroyed?
             def created_at
               @created_at ||= Time.now.in_time_zone
             end
+
+            def created_at=(date)
```
composerinteralia: Is there a way we could deal with arguments of the wrong type, like `stub.created_at = 'nonsense'`? The original `created_at` would type cast that string to `nil`, but this method will preserve whatever we pass in.

composerinteralia: Or I wonder if we could get rid of the custom setter and just have created at return `super` if it is present?

PR #1061, opened 2017-10-23
FactoryGirl 4.9.0

joshuaclayton:


lib/factory_girl.rb
```diff
@@ -154,3 +154,5 @@ def self.register_callback(name)
 
 FactoryGirl.register_default_strategies
 FactoryGirl.register_default_callbacks
+
+ActiveSupport::Deprecation.warn "The factory_girl gem is deprecated. Please upgrade to factory_bot. See https://github.com/thoughtbot/factory_bot/blob/4-9-0/UPGRADE_FROM_FACTORY_GIRL.md for further instructions."
```
jferris: Should this reference the tag in case we delete the branch?

PR #1064, opened 2017-10-25
Duplicate trait handling

Epigene:
Implemented two new behaviors for traits definitions for Issue @ https://github.com/thoughtbot/factory_bot/issues/974

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -764,7 +764,7 @@ factory :order do
 end
 ```
 
-Finally, traits can accept transient attributes.
+Traits can accept transient attributes.
 
 ```ruby
 factory :invoice do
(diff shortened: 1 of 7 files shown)


dmitriy-strukov:
@Epigene Nice pull request, it solve my problem, thank you! Can you please add improvements or if you don't have much time i can do it. WDYT?

Epigene:
If no additions need be made functionality-wise, I am happy to hand over tweaking syntax so that CI is happy to you, @dmitriy-strukov

composerinteralia:
Thanks for working on this!

Unfortunately these new options to not play well with trait inheritance:

```rb
define_model("User", name: :string)

FactoryBot.define do
  factory :user do
    name "Bob"

    trait :common do
      name "James"
    end

    factory :person do
      trait :common do
        name "John"
      end
    end
  end
end

expect(FactoryBot.create(:person, :common).name).to eq("John")
```

This test should pass in all three cases, but it will raise an error for `:raise` and return "James" for `:override`.

This has to do with the way traits are copied over from the parent to the child factory. When defining traits, they are immediately [defined on the factory's definition](https://github.com/thoughtbot/factory_bot/blob/3a4d6f489ccd3bb58674e470c3bfaad642888975/lib/factory_bot/definition_proxy.rb#L173-L175) . Later, when the factory is "compiled", the traits from any parent factories get [copied over to the child's definition](https://github.com/thoughtbot/factory_bot/blob/3a4d6f489ccd3bb58674e470c3bfaad642888975/lib/factory_bot/factory.rb#L86). Since the parent factories are copied over later, they will always win with the `:override` option. This is not what we want.

Traits are one of the more complicated parts of factory_bot, and you are uncovering some of that complexity here. I think in order to fix this we may need to dig into traits even more, and possibly restructure some things. https://github.com/thoughtbot/factory_bot/issues/968 seems closely related, and I think we would want to address it at the same time.

composerinteralia:
Given that there hasn't been much activity on this for a while, and I think solving the underlying issue (#968 is related as well) is going to require some restructuring of how we define traits, I am going to close this for now and continue tracking it in #974. Thanks for your work on this!

PR #1065, opened 2017-10-26
Changed all references of the name John to a random string of charact…

Circuit8:
Changed all references of the name John to a random string of characters.
    
John is a male name and causes many people offense, so I have replaced it with a string with no gender connotations.

In the wake of changing the name to factory_bot I though this would be a great step in the right direction.


joshuaclayton:
Thanks for the suggestion @Circuit8; we won't be merging this.

Circuit8:
Any reason you won't be merging it?

PR #1070, opened 2017-10-27
Added Unix instructions to upgrade guide from FactoryGirl to FactoryBot

roooodcastro:
The new 4.9.0 version issues a deprecation warning when I run my tests, and tell me to go to [UPGRADE_FROM_FACTORY_GIRL.md](https://github.com/thoughtbot/factory_bot/blob/v4.9.0/UPGRADE_FROM_FACTORY_GIRL.md) to see how to upgrade.

This page contains an example grep + sed command to automatically do the job for you. However it didn't work for me running Ubuntu 16.04.

I then fixed the command to work on Unix and made this patch to add the unix example command.

### DIFF
--- a/UPGRADE_FROM_FACTORY_GIRL.md
+++ b/UPGRADE_FROM_FACTORY_GIRL.md
@@ -32,10 +32,18 @@ end
 
 A global find-and-replace of `FactoryGirl` to `FactoryBot` across the codebase
 to replace all references with the new constant should do the trick. For
-example, on OS X:
+example:
+
+### OS X:
 
 ```sh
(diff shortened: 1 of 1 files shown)


jbcden:
For what it's worth I also found that I had to use the fix here for macOS 10.12.5. The suggested returned no results.

jbcden:
Actually just tweaked this again and managed to get the macOS version to work with:
```shell
grep -e FactoryBot **/**/*.rake **/**/*.rb -s -l
```
This seems to also work on Ubuntu 😄

composerinteralia:

PR #1073, opened 2017-10-30
Add global flag to sed command for migration

okuramasafumi:
Now the script will work for lines containing multiple `FactoryGirl`.


odlp:
Thanks for your contribution!

PR #1075, opened 2017-10-31
Add pictures for GUI editor to upgrade guide.

BobbyMcWho:
This is helpful for people who are not comfortable with complicated command line arguments and work better with a GUI like VSCode.


composerinteralia:

PR #1076, opened 2017-11-01
Replace more than one match per line

soulcutter:
When I migrated 50+ repositories, I ran into some that had more than one `FactoryGirl` on the same line, and the script only updated the first match.


odlp:
I've just merged in #1073 which resolves the issue (it was opened earlier), but thanks for your contribution - much appreciated!

PR #1083, opened 2017-11-21

npverni:
The find and replace for FactoryGirl to FactoryBot needs a g to replace multiple occurrences in the same line.

For example. Prior to the change:

FactoryGirl.create(:book, author: FactoryGirl.create(:author)) became FactoryBot.create(:book, author: FactoryGirl.create(:author))

Not the best way to be using factories, but there is probably occurrences like that in legacy apps.


odlp:
I've just merged in #1073 which resolves the issue (it was opened earlier), but thanks for your contribution - much appreciated!

PR #1084, opened 2017-11-27

diei:
At my OS X Yosemite 10.10.5 the grep commands does not work. I have updated them with the versions working for me.

### DIFF
--- a/UPGRADE_FROM_FACTORY_GIRL.md
+++ b/UPGRADE_FROM_FACTORY_GIRL.md
@@ -35,7 +35,7 @@ to replace all references with the new constant should do the trick. For
 example, on OS X:
 
 ```sh
-grep -e FactoryGirl **/*.rake **/*.rb -s -l | xargs sed -i "" "s|FactoryGirl|FactoryBot|"
+grep -rl -e FactoryGirl --include="*.rake" --include="*.rb" . | xargs sed -i "" "s|FactoryGirl|FactoryBot|"
 ```
 
 ## Replace All Path References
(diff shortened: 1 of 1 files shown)


UPGRADE_FROM_FACTORY_GIRL.md
```diff
@@ -35,7 +35,7 @@ to replace all references with the new constant should do the trick. For
 example, on OS X:
 
 ```sh
-grep -e FactoryGirl **/*.rake **/*.rb -s -l | xargs sed -i "" "s|FactoryGirl|FactoryBot|"
+grep -rl -e FactoryGirl --include="*.rake" --include="*.rb" . | xargs sed -i "" "s|FactoryGirl|FactoryBot|"
```
MikeRogers0: The include could also be `--include=\*.{rb,rake}` what do you think?

diei: That works as well.

tf: This removes the `-s` flag in the first but not the second command. It should probably stay.

tf:
On Ubuntu, I also had to remove the first `""` from the `sed` command (and add a `g` to the command, see #1083):

```
grep -rls -e factory_girl --include="*.rb" --include="*.rake" | xargs sed -i "s|factory_girl|factory_bot|g"                      
```
Leaving it in results in 
```
sed: can't read s|factory_girl|factory_bot|: No such file or director
```

composerinteralia:

PR #1085, opened 2017-11-30
Add extra info about requiring helpers in rails

blaszczakphoto:
It does not work in rails if you just add require `'support/factory_bot'` to rails helper.


composerinteralia:
Thanks for your contribution! See 8ad9cb1d716016c18c7fb349b32e474bb0aa4d94 for why I am closing this.

PR #1087, opened 2017-12-08
Fixes for FactoryBot::Evaluator overrides

BrianHawley:
- Makes `FactoryBot::Evaluator#__override_names__` be consistent.
- Makes it so you can detect overrides in factory callbacks.
- Lets you pass in a frozen hash of overrides without raising exceptions.

Discovered while using override detection to handle denormalized stuff.
The test is not a good example of the best way to handle the situation
specified in the test, but it does verify the problem.

### DIFF
--- a/lib/factory_bot/evaluator.rb
+++ b/lib/factory_bot/evaluator.rb
@@ -13,7 +13,7 @@ module FactoryBot
     def initialize(build_strategy, overrides = {})
       @build_strategy = build_strategy
       @overrides = overrides
-      @cached_attributes = overrides
+      @cached_attributes = overrides.dup
       @instance = nil
 
       @overrides.each do |name, value|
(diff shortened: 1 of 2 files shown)


composerinteralia:
> The test is not a good example of the best way to handle the situation
specified in the test, but it does verify the problem.

Is there a way we could write a test that does make it more obvious what the problem is? I'm a little unclear myself.

BrianHawley:
@composerinteralia I'll try to come up with a better test.

composerinteralia:
Since I'm unsure what problem we are trying to solve here, I'm going to close for now. Feel free to reopen with more context if there is something we need to fix here. Looking at the current test, my thought is that we don't officially support using `evaluator.__override_names__`. I would rather add a publicly documented feature than fix an edge case with undocumented usage.

PR #1095, opened 2018-01-07
Add information for ugrading on Ubuntu

fwoelm:
Using Ubuntu, I had to slightly modify the shell script to replace all occurrences of factory girl with factory bot. Thought I would share that with everyone.

For grep:
`-r` for recursive
`**/*` to follow all paths in the current directory
`--include` to search only .rb and .rake files
The reason the OS X script was not working is that it was not searching recursively. It was filtering out all directories (only files with .rb or .rake were matched) and therefore only producing a partial list of matches.

For sed: Had to drop the first (empty) string argument.

PS: Awesome gem :heart:

### DIFF
--- a/UPGRADE_FROM_FACTORY_GIRL.md
+++ b/UPGRADE_FROM_FACTORY_GIRL.md
@@ -38,11 +38,24 @@ example, on OS X:
 grep -e FactoryGirl **/*.rake **/*.rb -s -l | xargs sed -i "" "s|FactoryGirl|FactoryBot|"
 ```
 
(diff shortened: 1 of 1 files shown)


sunnyrjuneja:
just as an fyi, I get:

```zsh
$ grep -i -r FactoryGirl **/* --include=*.rb --include=*.rake -s -l | xargs sed -i "s|FactoryGirl|FactoryBot|"
zsh: argument list too long: grep
sed: no input files
```

composerinteralia:

PR #1101, opened 2018-01-29
Fix greps to do recursive searches during upgrade

mrgordon:
I found the provided `grep` commands to switch FactoryGirl -> FactoryBot weren't finding everything as they didn't search recursively. Hopefully this helps someone else that hasn't updated yet


odlp:
@mrgordon thanks for your contribution. I think this has been resolved by #1073 which updated the `sed` command to update multiple matches. Please feel free to re-open the issue if there are still outstanding problems with the find & replace instruction.

PR #1102, opened 2018-01-31
Fix upgrade from factory girl instructions

BKSpurgeon:
### What is the main proposed change?

**Summary**: the proposed change allows folks to easily edit  instructions.

**Detailed explanation**: The previous commit referred the upgrade instructions to a fixed tagged commit via a link. But then how can folks upgrade or improve upon those instructions? Instructions are living/breathing documents subject to change and editing. Hence these instructions must be included in the current HEAD / commit to allow for easy emendation. That's what this commit seeks to do.

### The second change

I have also edited the instructions re: upgrading. But, given the above change,  you can easily emend those instructions if I've made an error :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,15 @@ _[Interested in the history of the project name?](NAME.md)_
 
 ### Transitioning from factory\_girl?
 
-Check out the [guide](https://github.com/thoughtbot/factory_bot/blob/4-9-0-stable/UPGRADE_FROM_FACTORY_GIRL.md).
+Transitioning is a two step process:
 
+1) Replace `factory_girl` with `factory_bot` in your Gemfile and `bundle install`
+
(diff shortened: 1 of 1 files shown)


composerinteralia:
Thanks for your contribution!

I don't think the upgrade guide belongs at the top of `README.md`. It makes it more difficult for first-time users to find the documentation and install details. I would prefer a PR against the `4-9-0-stable` branch. The commits related to `UPGRADE_FROM_FACTORY_GIRL.md` are already on that branch, and it should be just as easy to edit there as it would have been to edit it on master.

composerinteralia:
I'm going to close this for now. Feel free to open a PR against `4-9-0-stable`. Many thanks!

BKSpurgeon:
I can add to the other branch.

PR #1091, opened 2017-12-26
Update GETTING_STARTED.md

1Mill:
Reduce the chance of misplacing the require statement in `rails_helper.rb` by providing some guidance.

PR #1081, opened 2017-11-10
adds instruction during factory_bot setup in rspec

shivabhusal:

PR #987, opened 2017-02-04
Markdown, formatting, and a couple sentence fixes

rhizome:

PR #1105, opened 2018-02-15
clarify how to include spec/support/factory_bot.rb to avoid confusion

jarvisjohnson:
It's pretty easy to load `spec/support/factory_bot.rb` twice accidentally by including this whole snippet: 

```ruby 
# spec/support/factory_bot.rb
RSpec.configure do |config|
  config.include FactoryBot::Syntax::Methods
end

# RSpec without Rails
RSpec.configure do |config|
  config.include FactoryBot::Syntax::Methods

  config.before(:suite) do
    FactoryBot.find_definitions
  end
end
```

not realizing one is for Rails and one not. This just clarifies the docs to hopefully avoid the mistake I (and [a few others](https://stackoverflow.com/questions/9300231/factory-already-registered-user-factorygirlduplicatedefinitionerror/#answer-41080111)) have made.

Thanks!

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -30,13 +30,19 @@ Configure your test suite
 
 ### RSpec
 
+If you're using Rails:
+
 ```ruby
 # spec/support/factory_bot.rb
 RSpec.configure do |config|
   config.include FactoryBot::Syntax::Methods
 end
+```
 
-# RSpec without Rails
+If you're *not* using Rails:
+
+```ruby
+# spec/support/factory_bot.rb
 RSpec.configure do |config|
   config.include FactoryBot::Syntax::Methods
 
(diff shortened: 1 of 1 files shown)


GETTING_STARTED.md
```diff
@@ -46,7 +52,7 @@ RSpec.configure do |config|
 end
 ```
 
-Remember to require the above file in your rails_helper since the support folder isn't eagerly loaded
```
composerinteralia: What do you think about getting rid of this line and the indication of which file the RSpec configuration belongs in? We don't indicate what file the configuration should go in for any of the other test libraries, and I have seen plenty of people put this configuration directly in the rails_helper or spec_helper. It is worth noting also that you are the sixth person (only looking at open PRs) to suggest an edit to this line (#987, #1025, #1081, #1085, and #1091). Perhaps it is confusing more people than it is helping.

jarvisjohnson: I still think it's necessary to mention that. I feel like a lot of new Ruby developers are pointed to Factory Bot in tutorials and might not be aware where that file would need to be required.  Thoughts?

composerinteralia:
Merged as 8ad9cb1d716016c18c7fb349b32e474bb0aa4d94. I amended your commit to remove the bit I had commented on. You can see in the commit message why I decided to do it that way. Thanks for your contribution!

PR #1107, opened 2018-02-18
Fix when FactoryGirl appears twice on same line

rdunlop:
If a spec has two `FactoryGirl` declarations on the same line, it would only replace the first command. We need to have a `g` to replace all occurrences.


ge-fa:
LGTM, thanks!

rdunlop:
this appears to have been resolved by another commit

PR #1124, opened 2018-04-27
FactoryBot.lint exposes original exceptions

jwoudenberg:
When FactoryBot.lint runs into an exception while testing a factory, only the message of that exception is preserved in the error returned by FactoryBot. Without the stack trace of that exception, it can be hard to figure out where the error breaking the factory is originating.

This PR makes the original errors accessible.


composerinteralia:
Related to #710

Issue #1125, opened 2018-05-01
The build_stubbed strategy adds extra timestamps


Mandaryn:
Even if model does not define `created_at` or `updated_at` the `build_stubbed` creates a stub with those defined.
The attributes should be added only to models that define specific timestamps.

PR #1126, opened 2018-05-01
Do not add missing timestamps when using build stubbed strategy

Mandaryn:
Solves https://github.com/thoughtbot/factory_bot/issues/1125

### DIFF
--- a/lib/factory_bot/strategy/stub.rb
+++ b/lib/factory_bot/strategy/stub.rb
@@ -69,9 +69,8 @@ module FactoryBot
         end
 
         created_at_missing_default = result_instance.respond_to?(:created_at) && !result_instance.created_at
-        result_instance_missing_created_at = !result_instance.respond_to?(:created_at)
 
-        if created_at_missing_default || result_instance_missing_created_at
+        if created_at_missing_default
           result_instance.instance_eval do
             def created_at
               @created_at ||= Time.now.in_time_zone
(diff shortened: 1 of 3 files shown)


spec/acceptance/build_stubbed_spec.rb
```diff
@@ -145,8 +145,9 @@
     ENV['TZ'] = original_timezone
   end
 
-  it "adds created_at to objects who don't have the method" do
-    expect(build_stubbed(:thing_without_timestamp)).to respond_to(:created_at)
+  it "doesn't add created_at to objects who don't have the method" do
```
composerinteralia: Is looks like this behavior was intentional. @joshuaclayton do you remember the context for adding `created_at` to an object that wouldn't normally respond to that? This was added in 4bccbb90d1f4bfe8e55bb48f4e93a760f77d1b8f

joshuaclayton: It's the result of https://github.com/thoughtbot/factory_bot/pull/390, where there's further discussion. Looking at this now, I agree that defining these methods automatically is a bit odd, as it's basing that assumption on FB being used only for ORM-backed classes.

The ability to override `created_at`/`updated_at`, however, is something I think we'd want to retain in cases where the class does inherit from `AR::B`/`ApplicationRecord`/et al, so this should be good to go. I still can't find a reference where we determine a need to set it on objects that don't define `created_at` though.

lib/factory_bot/strategy/stub.rb
```diff
@@ -71,7 +71,7 @@ def destroyed?
         created_at_missing_default = result_instance.respond_to?(:created_at) && !result_instance.created_at
         result_instance_missing_created_at = !result_instance.respond_to?(:created_at)
 
-        if created_at_missing_default || result_instance_missing_created_at
+        if created_at_missing_default
```
joshuaclayton: Because we're removing `result_instance_missing_created_at` (and `updated_at`, below), we should also remove the assignment of these values (on lines 72 and 84)

Mandaryn:
I removed the leftover assignments.

PR #1135, opened 2018-05-25
Print warning about static attribute deprecation

odlp:
On `stderr` a message will be displayed like:

```
DEPRECATION WARNING: Static attributes will be removed in FactoryBot 5.0. Please use dynamic attributes instead. Static attribute=:rockstar value=true (called from block (5 levels) in <top (required)> at /Users/oli/workspace/factory_bot/spec/acceptance/transient_attributes_spec.rb:11)
```

The caller / callsite in the message is set to where the attribute is declared. 

For further background on the motivation behind this change, see this commit: https://github.com/thoughtbot/factory_bot/commit/eaf2b403f35b2b7c9a014675d15d8b2646bd0601

### DIFF
--- a/lib/factory_bot/definition_proxy.rb
+++ b/lib/factory_bot/definition_proxy.rb
@@ -45,16 +45,7 @@ module FactoryBot
       declaration = if block_given?
         Declaration::Dynamic.new(name, @ignore, block)
       else
-        attribute_caller = caller
-
-        if attribute_caller[0].include?("method_missing")
-          attribute_caller = caller(2)
-        end
-
-        ActiveSupport::Deprecation.warn("Static attributes will be removed in "\
-          "FactoryBot 5.0. Please use dynamic attributes instead. Static "\
-          "attribute=#{name.inspect} value=#{value.inspect}", attribute_caller)
-
+        warn_static_attribute_deprecation(name, value)
         Declaration::Static.new(name, value, @ignore)
       end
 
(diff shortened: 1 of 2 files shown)


lib/factory_bot/definition_proxy.rb
```diff
@@ -170,5 +171,19 @@ def trait(name, &block)
...
       @definition.define_constructor(&block)
     end
+
+    private
+
+    def warn_static_attribute_deprecation(name, value)
+      attribute_caller = caller(2)
+
+      if attribute_caller[0].include?("method_missing")
+        attribute_caller = caller(3)
```
odlp: We need to skip further up the callstack depending on whether the attribute was declared explicitly (`add_attribute(:x, "y")`) or implicitly (`x: "y"`)

odlp:
Hound seems rather confused, the current thoughtbot style is double quoted strings:
https://github.com/thoughtbot/guides/blob/65e6f89916887cacb1cbd880a0f0bd4b8acaf68f/style/ruby/.rubocop.yml#L339-L343

This project may need a `hound.yml` / `rubocop.yml` file setup, which I'll address in a following PR.

mikegee:
Based on exactly zero evidence, I suspect there are test suites that will get many screenfuls of this deprecation warning. Do you think this is a legitimate concern that deserves some additional research or mitigation?

odlp:
@mikegee I would also anticipate there will be test suites which will display this deprecation notice a lot - and I think that's ok. The user can either:

1. Fix the warnings
2. Keep `factory_bot` at the previous version without the warnings until they have bandwidth to make the changes
3. Turn off warnings (not advisable, but an option)

I don't think we should hold back in making progress towards the goals of `5.0`, and helping people progressively make changes in preparation is a good thing in my opinion.

mikegee:
Thanks @odlp, those first two options are legit indeed.

Other thoughts (perhaps not worth the effort):

- a single warning explaining the situation for any number of static attributes
- maybe create a tool to convert static to dynamic

composerinteralia:
It looks like Josh had also worked on this a bit in #1021. Do we want to offer some configuration flag to quiet the deprecation warning?

composerinteralia:
I bet we could use something like [this rubocop-rspec cop](https://github.com/rubocop-hq/rubocop-rspec/blob/master/lib/rubocop/cop/rspec/factory_bot/dynamic_attribute_defined_statically.rb) to automatically fix most cases. I could look into it this coming Friday. @koic what do you think about this idea?

composerinteralia:
OK, now that the Cop is ready, I am going to go ahead an merge this. Before releasing 4.11 I will add some documentation to make sure people know the Cop is available.

PR #1137, opened 2018-05-29
Adds require factory_bot

ezuk:
Without requiring factory_bot, on a fresh Rails 5 installation, you get an uninitialized constant FactoryBot error when trying to run rspec as per the docs.

See https://stackoverflow.com/a/48091777/592229 for others experiencing the same issue.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -32,6 +32,7 @@ Configure your test suite
 
 ```ruby
 # spec/support/factory_bot.rb
+require 'factory_bot'
 RSpec.configure do |config|
   config.include FactoryBot::Syntax::Methods
 end


mikegee:
I've never needed this in the various Rails apps I've used FactoryBot(Girl). In my experience, Rails will tell Bundler to `require` all the gems listed in the `Gemfile`. I'm curious why that's not working for you.

ezuk:
I have no idea; I've hunted for a solution for quite a while and then found this Stack Overflow issue where others seemed to be experiencing the same issue. This was on a new Rails 5 installation, nothing custom or funky. :)

mikegee:
I duplicated the problem by generating a Rails app, installing RSpec & FactoryBot. The problem happened for me when I call `require 'support/factory_bot'` too early. It works ok f i move that `require` after line 8 in `spec/rails_helper.rb`:

```
# Add additional requires below this line. Rails is not loaded until this point!
```

ezuk:
Ah, fair point! Thank you, closing.

PR #1144, opened 2018-06-12
Inherit from BasicObject instead of undefining

composerinteralia:
`FactoryBot::DefinitionProxy` currently undefines all but a few methods.
Instead, it can inherit from `BasicObject` and then get any other methods
it needs (`#block_given?` and `#raise` are it as far as I can tell)
directly from `Kernel`.

This way if somebody adds a method to Kernel after loading a factory
that has an attribute of the same name as that method, there is still
some hope their code might work. (Of course nobody is ever adding
methods to Kernel, right?)

### DIFF
--- a/lib/factory_bot/definition_proxy.rb
+++ b/lib/factory_bot/definition_proxy.rb
@@ -1,10 +1,5 @@
 module FactoryBot
-  class DefinitionProxy
-    UNPROXIED_METHODS = %w(__send__ __id__ nil? send object_id extend instance_eval initialize block_given? raise caller method)
-
-    (instance_methods + private_instance_methods).each do |method|
-      undef_method(method) unless UNPROXIED_METHODS.include?(method.to_s)
-    end
+  class DefinitionProxy < BasicObject
 
     delegate :before, :after, :callback, to: :@definition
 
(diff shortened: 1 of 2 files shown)


jferris:
Here are the methods on `Object` that we used to keep that AREN'T defined on `BasicObject`:

* `block_given?`
* `caller`
* `extend`
* `initialize`
* `method`
* `nil?`
* `object_id`
* `raise`
* `send`

It looks to me like `block_given?`, `caller`, and `raise` can be used on `Kernel`.

We don't need `initialize` (it still works on `BasicObject` subclasses) and `send` (we can use `__send__`).

The ones I'm unsure of are `extend`, `method`, `nil?`, and `object_id`, but if the tests are passing, it's unlikely that any of those are used for `DefinitionProxy`.

It's probably worth trying this out in an application to see if there will be any issues with constant resolution or anything like that in attribute blocks.

composerinteralia:
I tried out on a few applications and it seems to work fine.

I supposed somebody could do this:

```rb
module DefinitionProxyExtensions
...
end

FactoryBot.define do
  factory :thing do
    extend DefinitionProxyExtensions
  end
end
```

But I'm not sure we want to support that.

PR #1149, opened 2018-06-22
Add version and semver stability badges to README

greysteil:
Same as https://github.com/thoughtbot/factory_bot_rails/pull/279, but for `factory_bot` instead of `factory_bot_rails`.

[Gem Version](http://badge.fury.io/rb/factory_bot) [SemVer](https://dependabot.com/compatibility-score.html?dependency-name=factory_bot&package-manager=bundler&version-scheme=semver)

Also cleans up the link style to match `factory_bot_rails`.


composerinteralia:
Posted on `factory_bot_rails`

> I went back and forth on this, but I think I am going to pass on the semver stability badge. I have heard good things about Dependabot, but I have concerns about whether that percentage accurately represents what it says it does (CI can fail for all sorts of other reasons), and about how useful it is to people using the library (I would normally just update and trust my test suite to let me know if anything is broken). The gem version badge seems fine to me; if you want to update the PR to include just that I will merge.

greysteil:
Updated to match thoughtbot/factory_bot_rails#279.

Thanks again for the feedback.

PR #1156, opened 2018-07-09
Fixed problem when name of the trait is not a symbol

pacop:
Whereas factories can be declared using both string and symbol, traits were sensitive of the used type. This merge-request addresses the problem converting the name of the trait to a symbol.


spec/acceptance/traits_spec.rb
```diff
@@ -20,7 +20,7 @@
           great "GREAT!!!"
         end
 
-        trait :admin do
+        trait "admin" do
```
composerinteralia: I could image myself "correcting" this back to a symbol to match all the other traits. Could we add a separate example that is a bit more explicit about what we are testing?

pacop:
@composerinteralia let me know if current version is better. Thanks

composerinteralia:
Merged as f4ad4f0c87747b5511288004bb1ba249478e73c9

PR #1167, opened 2018-08-11
Don't include extra files in the gem

composerinteralia:
With `git ls-files` we were still including unneeded dotfiles and
appraisal files. This switches to a whitelist of files included in the
gem, rather than a blacklist. I also tried to make RuboCop happy while I
was in here.

### DIFF
--- a/factory_bot.gemspec
+++ b/factory_bot.gemspec
@@ -10,7 +10,9 @@ Gem::Specification.new do |s|
                   "using factories - less error-prone, more explicit, and "\
                   "all-around easier to work with than fixtures."
 
-  s.files = Dir.glob("lib/**/*") + %w[GETTING_STARTED.md LICENSE README.md]
+  s.files =
+    Dir.glob("lib/**/*") +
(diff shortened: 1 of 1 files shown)


factory_bot.gemspec
```diff
@@ -1,18 +1,18 @@
...
-                      using factories - less error-prone, more explicit, and
-                      all-around easier to work with than fixtures.}
+  s.summary     = "factory_bot provides a framework and DSL for defining and "\
+                  "using model instance factories."
+  s.description = "factory_bot provides a framework and DSL for defining and "\
+                  "using factories - less error-prone, more explicit, and "\
+                  "all-around easier to work with than fixtures."
 
-  s.files         = `git ls-files`.split("\n").reject { |f| f.match(%r{^(spec|features)/}) }
+  s.files = Dir.glob("lib/**/*") + %w[GETTING_STARTED.md LICENSE README.md]
```
composerinteralia: going to include NEWS and CONTRIBUTING.md as well (and in https://github.com/thoughtbot/factory_bot_rails/pull/286)

mike-burns: How about `NAME.md` and `NEWS` too?

Issue #1172, opened 2018-08-19
GETTING_STARTED.md does not exist on RubyDoc


BerkhanBerkdemir:
Hello,

I was looking the documentation, but GETTING_STARTED.md which is on README.md returns not found on RubyDoc.

I want to ask that because you have GETTING_STARTED.md on the codebase; however, you redirect to RubyDoc.

composerinteralia:
Nice catch. It looks like I should have included .yardopts in https://github.com/thoughtbot/factory_bot/pull/1167. Also, I think .yardopts should have included NAME.md. I suspect we have a similar problem in factory_bot_rails. Any interest in opening a PR?

BerkhanBerkdemir:
@composerinteralia, so, you want to add something like that, right?

[`.yardopts`](https://github.com/thoughtbot/factory_bot/blob/dacc167803e87fc1433f91493ff8cf0b8aa583fe/.yardopts)
```diff
CONTRIBUTING.md
+ NAME.md
LICENSE
```

[`factory_bot.gemspec`](https://github.com/thoughtbot/factory_bot/commit/fc451872b9b4e7e17b1d19e8e6f27d918acd3ef8#diff-ff1a92cd12d8179e487b79302e7e9316R15)
```diff
+ %w[CONTRIBUTING.md GETTING_STARTED.md LICENSE NAME.md NEWS README.md .yardopts]
```

composerinteralia:
Yeah, I think so

BerkhanBerkdemir:
Hi @composerinteralia,

I created a PR. Could you check it out?

composerinteralia:
Fixed in #1175

PR #1175, opened 2018-08-23
Fix missing docs (#1172)

BerkhanBerkdemir:
This PR fixes issue #1172. Please see issue as the description of this PR.

### DIFF
--- a/.yardopts
+++ b/.yardopts
@@ -2,4 +2,5 @@ lib/**/*.rb
 -
 GETTING_STARTED.md
 CONTRIBUTING.md
+NAME.md
 LICENSE
--- a/factory_bot.gemspec
+++ b/factory_bot.gemspec
@@ -12,7 +12,7 @@ Gem::Specification.new do |s|
 
   s.files =
     Dir.glob("lib/**/*") +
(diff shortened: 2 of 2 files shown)


composerinteralia:
Looks reasonable to me. I’m out this week, but will try to get this through when I get back next week.

composerinteralia:
The problem with #1183 is that it links to the version of the file on `master`, rather than the version of the file associated with the specific gem version. What you have in this PR should be enough to fix the problem, since `GETTING_STARTED.md` is listed in `.yardopts`. I will be sure to test this out before merging.

composerinteralia:
This change seems good to me. There is still some inconsistency with the way we link to things (see the links for NAME.md and LICENSE as compared to GETTING_STARTED.md), but I think we can clean those up later.

BerkhanBerkdemir:
@composerinteralia [the link](https://www.rubydoc.info/gems/factory_bot/file/GETTING_STARTED.md) still doesn't work :sob:

composerinteralia:
I haven't put together a release yet, so it makes sense that it doesn't work. I'll try to do it soon!

PR #1187, opened 2018-09-07
Working link to GETTING_STARTED

pablobm:
The rubydoc.info link stopped working. I spent a few minutes trying to figure it out (the `.yardopts` file seems correct?), and I can't tell why. This is a quick fix in absence of anything better.


composerinteralia:
Fixed in https://github.com/thoughtbot/factory_bot/pull/1175, but not released yet.

PR #1191, opened 2018-09-10
Raise KeyError instead of ArgumentError in Registry

### LINKED ISSUE #992: Add `did_you_mean` to "Trait not registered"
It would be nice to have:

```
     ArgumentError:
       Trait not registered: in_progess
       Did you mean? in_progress
```

Instead of:

```
     ArgumentError:
       Trait not registered: in_progess
```

ChrisBr:
to make use of did_you_mean gem for better error messages.

This will partially fix #992.

However, there is one issue we might need consider with this solution. The did_you_mean gem only supports three different types of errors:

* NotMethodError
* NameError
* KeyError

``NoMethodError`` does not make sense. 

In #992 it was discussed to use ``NameError``, however, when using a ``NameError`` it is not possible to use a customized error message like 'Factory not registered: :foobar' because did_you_mean initialized the [SpellChecker](https://github.com/yuki24/did_you_mean/blob/master/lib/did_you_mean/spell_checkers/name_error_checkers.rb#L7) depending on the error message. I think that a customized error message is essential! 

This leaves us with ``KeyError`` (which I think also matches the error).

But only raising this ``KeyError`` is not enough as the did_you_mean gem makes heavily use of the ``receiver`` attribute (and for ``KeyError`` also the ``key`` attribute) to calculate the corrections (see e.g. https://github.com/yuki24/did_you_mean/blob/master/lib/did_you_mean/spell_checkers/key_error_checker.rb#L7). Unfortunately in Ruby <= 2.5 it is not possible to set the ``receiver`` and ``key`` attributes. This will be possible from Ruby 2.6 on (see https://bugs.ruby-lang.org/issues/14313).  So the only possibility I currently see is to also monkey patch the ``KeyError`` class :cry: 

See my PR for a possible solution. The solution is far from perfect but a starting point for discussion. I'm also not sure if we should really monkey patch the ``KeyError`` class for better error messages (although I would like to have better error messages).

Here is also an example application to demonstrate the issue:
```
require 'factory_bot'

class User
  attr_accessor :name, :admin

  def save!
  end
end

FactoryBot.define do
  factory :user do
    name { "John" }
    admin { false }

    trait :admin do
      admin { true }
    end

  factory :admin_user,                 :traits => [:admin]
  end
end

def do_with_rescue
  begin
    yield
  rescue KeyError => e
    puts e.message
    puts "==="
  end
end


do_with_rescue { user = FactoryBot.create(:use) }
do_with_rescue { user = FactoryBot.create(:admin_use) }
do_with_rescue { user = FactoryBot.create(:user, :admi) }

# Factory not registered use
# Did you mean?  :user
# ===
# Factory not registered admin_use
# Did you mean?  :admin_user
# ===
# Trait not registered admi
# ===
```

As you can also see, the solution only works for factories so far but not for traits. The problem with traits is that it seems to use a complete different code path and the items hash does not contain the traits but the strategies (like create, build etc.). I would need some hints how to implement this for traits as well.


composerinteralia:
This looks promising! I understand your hesitation about the monkey patch, although I think there is a strong argument for it.

As for the missing did-you-mean message for traits, I think that is because we don't register factory-level traits in the global registry. They are local to the factory definition. It would work for a global trait though:

```ruby
require 'factory_bot'

class User
  attr_accessor :name, :admin

  def save!
  end
end

FactoryBot.define do
  trait :admin do
    admin { true }
  end

  factory :user do
    name { "John" }
    admin { false }

    factory :admin_user, :traits => [:admin]
  end
end

# ===
# Trait not registered admi
# Did you mean?  :admin
# ===
```

We recently started "privately" (with lots of double underscores) registering local sequences in the global register in https://github.com/thoughtbot/factory_bot/pull/1164. I wonder if we could do the same thing for traits. I suggested in https://github.com/thoughtbot/factory_bot/pull/1156#pullrequestreview-143418432 that doing so could also solve some other issues related to traits. Any interest in looking into that? That would be a separate PR, with no specific dependency on this one.

lib/factory_bot/registry.rb
```diff
@@ -1,4 +1,20 @@
...
+    attr_accessor :receiver, :key
+
+    def initialize(msg, key: nil, receiver: nil)
+      @key = key
+      @receiver = receiver
+      super msg
+    end
+  end
+
+  KeyError.prepend KeyErrorExtension
```
composerinteralia: Is there a way to do this only if it isn't already available? The built-in version will probably be a bit faster.
We could maybe check `KeyError.instance_method(:initialize).arity`, although that isn't the prettiest thing.

lib/factory_bot/registry.rb
```diff
@@ -1,4 +1,20 @@
 module FactoryBot
+  # In Ruby <= 2.5 it is necessary to monkey patch
+  # the key error class because it does not expose
+  # receiver and key attributes.
+  # See https://bugs.ruby-lang.org/issues/14313
+  module KeyErrorExtension
+    attr_accessor :receiver, :key
```
composerinteralia: I was under the impression there methods were always available, these just weren't using instance variables before. Do we need to redefine them?

ChrisBr: You're right, patching initialize should be enough. Dropping ...

ChrisBr: Turns out, the attributes exist but are not exposed means we need at least ``attr_reader``.

lib/factory_bot/registry.rb
ChrisBr: What about:
```
Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.6')
```
?

spec/factory_bot/registry_spec.rb
```diff
@@ -17,7 +17,7 @@
   end
 
   it "raises when an object cannot be found" do
-    expect { subject.find(:object_name) }.to raise_error(ArgumentError, "Great thing not registered: object_name")
+    expect { subject.find(:object_name) }.to raise_error(KeyError, "Great thing not registered: object_name")
```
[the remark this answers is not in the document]
ChrisBr: This line was already too long before my patch, should I change it anyway?

spec/acceptance/traits_spec.rb
```diff
@@ -167,7 +167,7 @@
 
   context "factory outside of scope" do
     subject     { FactoryBot.create(:user_without_admin_scoping) }
-    it { expect { subject }.to raise_error(ArgumentError, "Trait not registered: admin_trait") }
+    it { expect { subject }.to raise_error(KeyError, "Trait not registered: admin_trait") }
```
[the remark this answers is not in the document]
ChrisBr: This line was already too long before my patch, should I change it anyway?

ChrisBr:
I will look into https://github.com/thoughtbot/factory_bot/pull/1156, let's see if we can fix it for local traits as well.

lib/factory_bot/registry.rb
```diff
@@ -1,4 +1,22 @@
...
+    attr_reader :receiver, :key
+
+    def initialize(msg, key: nil, receiver: nil)
+      @key = key
+      @receiver = receiver
+      super msg
+    end
+  end
+
+  if Gem::Version.new(RUBY_VERSION) < Gem::Version.new("2.6")
```
composerinteralia: I don't think this would work for jruby or other implementations, but I don't know if factory_bot currently works at all for any other Ruby implementations. We stopped running CI for jruby in https://github.com/thoughtbot/factory_bot/pull/1018, and are rbx build hasn't been passing lately.

ChrisBr:
> We recently started "privately" (with lots of double underscores) registering local sequences in the global registry in #1164. I wonder if we could do the same thing for traits. I suggested in #1156 (review) that doing so could also solve some other issues related to traits. Any interest in looking into that? That would be a separate PR, with no specific dependency on this one.

@composerinteralia so I looked a little bit into https://github.com/thoughtbot/factory_bot/pull/1156. Indeed it would potentially fix at least the string vs. symbol issue as well as duplicate traits. However, we would register to the registry something like:

```
"__#{defintion.name}__#{trait.name}__"
```

right?

That would also mean that then we would print to the e.g. user:

```
# ===
# Trait not registered admi
# Did you mean?  :user_admin
# ===
```

So actually not the 'real' trait name. Not sure if this is what we want?

lib/factory_bot/registry.rb
ChrisBr: Just verified with ``rubinius 3.107 `` and ``jruby 9.2.0.0 (2.5.0)``, works for both. ``RUBY_VERSION`` is in both implementations the actual Ruby version without the platform.

ChrisBr:
@composerinteralia should we keep the KeyError monkey patch in the registry.rb file or should we move it to a dedicated file?

composerinteralia:
I think it would make sense to move it to it to it's own file.

Can you open a PR with whatever work you have done on traits so far and we can continue that discussion there? It make still make sense even if it doesn't solve the KeyError problem.

Thanks for your work on this!

ChrisBr:
> I think it would make sense to move it to it to it's own file.

Moved to dedicated file and fixed hound violations. Is there anything else left to get this merged?

> Can you open a PR with whatever work you have done on traits so far and we can continue that discussion there? It make still make sense even if it doesn't solve the KeyError problem.

Is there a IRC channel or chat the maintainers hang out? I would have some questions first... I just tried some stuff out, it fixed some stuff but broke a lot of other. So I'm not sure about the approach actually.

composerinteralia:
This PR looks good to me. I want to get at least one other set of eyes on it.

I had one other thought on this, which is that we could maybe use Ruby's refinements to keep the monkey patch local to this library. Thoughts?

I don't think there is an IRC channel, but perhaps you could open a new issue and we could have the discussion there.

ChrisBr:
> I had one other thought on this, which is that we could maybe use Ruby's refinements to keep the monkey patch local to this library. Thoughts?

Hm, yeah we could do that. Although I don't think that this monkey patch would have any undesired side effects (famous last words ...).

Anyway, I was playing around with the refinements and it seems it is not possible to overwrite the ``initialize`` method, see example:

```ruby
module FactoryBot
  module CoreExtensions
    module KeyError
      attr_reader :key, :receiver

      def initialize(msg, key: nil, receiver: nil)
        @key = key
        @receiver = receiver
        super msg
      end
    end
  end
end

module FactoryBot
  module CoreExtensions
    refine ::KeyError do
      attr_reader :key, :receiver

      def initialize(msg, key: nil, receiver: nil)
        @key = key
        @receiver = receiver
        super msg
      end
    end
  end
end


using FactoryBot::CoreExtensions
e = KeyError.new("refine", key: :key, receiver: :receiver)
# `initialize': wrong number of arguments (given 2, expected 0..1) (ArgumentError)

KeyError.prepend FactoryBot::CoreExtensions::KeyError
e = KeyError.new("monkey patch", key: :key, receiver: :receiver)
```

The monkey patch works as expected. Do you see sth what we do wrong? Or is this a limitation of refinements? Although I did not see anything in the docs regarding it and I also wouldn't expect that overwriting initialize is not possible ... Anyway, if this is the case we could only manually set key and receiver after initializing the object which would work but is not nice tbh.

composerinteralia:
That's an interesting limitation of refinements. I have no idea whether it is supposed to work that way. Thanks for looking into it!

ChrisBr:
> That's an interesting limitation of refinements. I have no idea whether it is supposed to work that way. Thanks for looking into it!

@composerinteralia I opened an issue for Ruby and it is indeed intended behavior because ``initialize`` is called outside of the scope of the Refinement of course. See https://bugs.ruby-lang.org/issues/15135 :cry: 

Hmm, I still like the idea of refinements but it seems like we need to do some workarounds to get this working which I'm not sure is worthwhile... We could do sth like this:

```
e = KeyError.new(msg)
e.receiver = @items
e.key = name
raise e
```

But we would need use the refinement always, also for Ruby 2.6 because 2.6 does not expose setter / getter for receiver / key attributes.

composerinteralia:
Sorry for the delay. I still think this is probably merge-able as is, but I had one more idea that could work once https://github.com/thoughtbot/factory_bot/pull/1212 gets merged.

Rather than trying to mess around with `KeyError` class, would could just let Ruby raise its own `KeyError`, then grab the message off that (which would already include the did_you_mean text), customize the error message and re-raise the error.

```ruby
def find(@items)
  with_custom_key_error { @items.fetch(name) }
end

private

def with_custom_key_error
  yield
rescue KeyError => e
  message = e.message.sub("key not found", "#{@name} not registered")
  error = KeyError.new(message)
  error.set_backtrace(e.backtrace)
  raise error
end
```

ChrisBr:
Hey @composerinteralia,

have you tried this already? I have some doubt because ``did_you_mean`` will not trigger when there is no receiver & key attribute. I will look into it. But honestly I think I prefer the monkey patch as this will be the official way from Ruby 2.6 on ...

composerinteralia:
I played around with it in the console and it did seem to work, but I haven't put it all together yet. I'll push up a branch that we can play around with.

The monkey patch does still make me nervous. While it is indeed similar to what will be available in 2.6, it is not the same. I fear the differences may come back to bite us. The one that recently occurred to me is that factory_bot will generally be included only in test and development environments, which means that this patch will cause `KeyError#intialize` to have a different arity in production than it had in test or development. I could imagine somebody trying out raising a `KeyError` in their application code and including the new keyword arguments, not realizing those arguments shouldn't be available in the version of Ruby they are on. Everything will work fine until they get a rude surprise in production: `ArgumentError: wrong number of arguments`.

I realize as I am writing this that we also made the `msg` argument required, when it should be optional. If we are going to include the monkey patch, I think we would need to include some more test cases to make sure we don't break `KeyError`. But at this point I am leaning pretty strongly toward exploring alternate solutions.

composerinteralia:
Here is a branch with what I had suggested: https://github.com/thoughtbot/factory_bot/tree/dc-cb-did-you-mean-test. It seems to work. What do you think?

ChrisBr:
@composerinteralia sorry I was travelling the last weeks, I will look into it the next days and come back to you. Thanks :+1:

ChrisBr:
@composerinteralia sorry for the loooong delay! I adapted now your commit (needed to create the message in the find method, otherwise the key in the message would be not correct). I added you as co author, I hope that's fine as this is mostly your idea.

composerinteralia:
Thanks so much! I know there was a lot of back and forth on this issue, but I really appreciate you sticking with it.

I noticed that the final version didn't give us the `did_you_mean` message, which was the original purpose of all this, so I combined what you had with a branch of mine and merged as b86ec13.

I think the change to the way the name shows up is fine, since it makes it more similar to other `KeyError`s.

PR #1193, opened 2018-09-11
Remove deprecated methods

composerinteralia:
Since we have started work on factory_bot 5, we no longer need to keep
these deprecated methods around.


composerinteralia:
Yes, but I haven't decided what is the best way to handle that one yet. I opened https://github.com/thoughtbot/factory_bot/issues/1196 to track it

PR #1163, opened 2018-08-03
Remove static attributes

composerinteralia:
- [x] Autocorrect static attributes to dynamic ones with:
```sh
  rubocop \
    --require rubocop-rspec \
    --only FactoryBot/AttributeDefinedStatically \
    --auto-correct
```
- [x] Remove static attributes from the definition proxy
- [x] Stop referring to static attributes in specs
- [x] Remove static attribute files
- [x] Update documentation (GETTING_STARTED.md and lib/factory_bot/definition_proxy.rb)

Co-authored-by: @georgewambold

PR #1194, opened 2018-09-12
Changing gem install command to use single quotes to appease Rubocop

rambleraptor:
If you're copying + pasting this into a Gemfile, Rubocop will complain about the double quotes.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -7,13 +7,13 @@ Update Your Gemfile
 If you're using Rails, you'll need to change the required version of `factory_bot_rails`:
 
 ```ruby
-gem "factory_bot_rails", "~> 4.0"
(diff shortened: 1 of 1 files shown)


composerinteralia:
Thanks for the PR! I'm going to close this, since the official style on this project is to use double quotes. I'm not exactly sure where we are getting that configuration, so it might be worth adding an explicit .rubocop.yml, but you can see in https://github.com/thoughtbot/factory_bot/pull/1163#discussion_r207628103 that hound will bark at us if we use double quotes. I think it makes sense for our documentation to match the the code style for the project.

ChrisBr:
@composerinteralia similar topic: Is there a way to run rubocop locally? I didnt see a rubocop config file? It is a little tiresome to wait for the hound comments after pushing and then trying to blindly fix the issues again ...

PR #1197, opened 2018-09-15
Add rubocop

### LINKED ISSUE #1195: Add a .rubocop.yml
I can't figure out what .rubocop.yml hound is using. It doesn't seem to be the default RuboCop config (it prefers double quotes to single quotes, for example). We should add a .rubocop.yml with whatever our current preferences are. That way we can run RuboCop locally, rather then waiting for Hound to bark at us. 

I think we can run `rubocop --auto-gen-config` for this.

cc @ChrisBr

ChrisBr:
to run rubocop locally and not to wait for HoundCI to bark at us.

The version is locked to 0.54 because this is the version HoundCI is using.
Using a newer version will not work as config changed from 0.54 to currently 0.59
and rubocop would complain with the old configuration.

The configuration is using the default configuration from version 0.54.

See http://help.houndci.com/configuration/rubocop.

Fix #1195.


composerinteralia:
If what we have in .rubocop.yml are all the defaults for the version we are using, can we just leave that file empty except for the `inherit_from: .rubocop_todo.yml` line?

The `.rubocop_todo.yml` disables `Style/StringLiterals`, but hound is still leaving comments for that Cop. That makes no sense to me. Do you know what is going on there?

I think we should start by getting this in, then we can gradually fix the todos. I don't really want to review one giant PR that changes everything.

ChrisBr:
> If what we have in .rubocop.yml are all the defaults for the version we are using, can we just leave that file empty except for the inherit_from: .rubocop_todo.yml line?

We could but I personally would prefer to have the config for

1) If we update rubocop or Hound updates rubocop, the defaults might change.
2) it's more verbose

But I can live with it if you prefer to have it dropped.

> The .rubocop_todo.yml disables Style/StringLiterals, but hound is still leaving comments for that Cop. That makes no sense to me. Do you know what is going on there?

I run rubocop --auto-gen-config which generates the todo file. However, if there are more than 15 violations of a cop, it gets disabled. See my commit with the string literal fixes, we have many many files which use single quotes and therefore it got disabled again. For instance LineLength and ClassLength are basically also disabled (just have a very high limit). We should get this in and then decide what we want to fix or disable in the config file.

> I think we should start by getting this in, then we can gradually fix the todos. I don't really want to review one giant PR that changes everything.

Yeah, right.

I also configured Hound now to use our rubocop config. However, it seems it does not respect it. Can you maybe check if you have an organisation wide configuration which takes precedence?

composerinteralia:
Alright, I think I found something that will work. Check out https://github.com/thoughtbot/factory_bot_rails/pull/291. Hound does not support `inherited_from`, so we have to put the todos into the main `.rubocop.yml`.

But I also came across this: https://robots.thoughtbot.com/hound-introduces-style-guides-for-organisations. We use the .rubocop.yml from https://github.com/thoughtbot/guides. I'll ask around here at thoughtbot to see if there is a good way to use that organization-wide configuration locally.

ChrisBr:
> Alright, I think I found something that will work. Check out thoughtbot/factory_bot_rails#291. Hound does not support inherited_from, so we have to put the todos into the main .rubocop.yml.

Hmm ok, I will change it!

Commented on your other PR...

composerinteralia:
I made a few edits to your original commit here and then merged as https://github.com/thoughtbot/factory_bot/commit/5b0257363ca4d14466e740b826374337f0c06d9b.  You are still the author of the commit though.

(Then I merged another commit to remove the .rubobo_todo.yml that I didn't mean to merge in 😢)

If anything doesn't work as expected, feel free to make more changes as necessary.

ChrisBr:
Awesome, thanks!

Should we create a follow up issue to fix all the rubocop violations from the todo file and mark it as ``Good first issue`` or something?

composerinteralia:
Good idea. Done!

PR #1199, opened 2018-09-18
Manually raise NoMethodError in definition proxy

composerinteralia:
Inside `super` Ruby calls `inspect`, but since we have undefined
`inspect` on the definition proxy we end up declaring an implicit
`inspect` attribute. We do the same thing with `methods` and
`singleton_methods`, and then `inspect` again. By the time we finally
see the NoMethodError, we see it on a
`FactoryBot::Declaration::Implicit` instead of the definition proxy.

By raising the NoMethodError manually we avoid this nonsense and have
the added benefit of showing the name of the factory where the missing
method came from.

### DIFF
--- a/spec/factory_bot/definition_proxy_spec.rb
+++ b/spec/factory_bot/definition_proxy_spec.rb
@@ -56,9 +56,15 @@ describe FactoryBot::DefinitionProxy, "#method_missing" do
     expect(subject).to have_dynamic_declaration(:attribute_name).with_value(attribute_value)
   end
 
-  it "calls super" do
+  it "raises a NoMethodError" do
+    definition = FactoryBot::Definition.new(:broken)
+    proxy = FactoryBot::DefinitionProxy.new(definition)
+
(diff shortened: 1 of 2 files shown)


spec/factory_bot/definition_proxy_spec.rb
```diff
@@ -56,7 +56,7 @@
     expect(subject).to have_dynamic_declaration(:attribute_name).with_value(attribute_value)
   end
 
-  it "calls super" do
+  it "raises a NoMethodError" do
     invalid_call = -> { proxy.static_attributes_are_gone true }
     expect(invalid_call).to raise_error(NoMethodError)
```
joshuaclayton: Since the benefit of doing this is for readability, what are your thoughts on expanding the `raise_error` expectation to also verify we inform the developer of what's happening?

composerinteralia: Yup, that makes sense

Issue #1202, opened 2018-09-19
Gradually fix RuboCop todos


composerinteralia:
We added a .rubocop.yml in https://github.com/thoughtbot/factory_bot/commit/5b0257363ca4d14466e740b826374337f0c06d9b with a bunch of todos. It would be nice to gradually fix them. I would like to do it one at a time so nobody has to review a giant PR. 😸

colinross:
Thinking like 1 commit per offense type?

composerinteralia:
Yeah, I think so, but use your best judgement. The main thing would be to keep each pull request fairly small.

ChaoticBoredom:
Are there any of these remaining to be addressed? It looks like #1208 hits most remaining things, unless we want to deviate from the primary rubocop.yml from thoughtbot.

composerinteralia:
@ChaoticBoredom I merged #1208, but there are still a few more TODOs in https://github.com/thoughtbot/factory_bot/blob/master/.rubocop.yml

PR #1203, opened 2018-09-21
[Rubocop] Layout Cop Offenses

colinross:
Begins to address #1202 

This is just the first stack of the 'Layout' cops corrected. I broke up the changes into a set of commits that should be able to easily digest. 

I also decided to not move on the `Layout/IndentHeredoc` cop, since I was a bit unsure how you would prefer to have HEREDOC spacing enforced (there is no rubocop default). I didn't notice anything in the greater thoughtbot policy on this either. See: https://github.com/rubocop-hq/rubocop/blob/master/manual/cops_layout.md#layoutindentheredoc


colinross:
Also, I'll be avoiding any @houndci-bot issues unless I _introduced_ them in order to not muck up the waters too much.

gemfiles/4.2.gemfile
```diff
@@ -2,9 +2,9 @@
 
 source "https://rubygems.org"
 
+gem "activerecord", "~> 4.2.5.1"
```
composerinteralia: The files in `gemfiles/` are generated by Appraisal, so we shouldn't edit them. Instead, let's tell RuboCop to exclude this directory.

colinross:
Ignored the `gemfiles/`;  feel free to squash merge.

I'll 'tackle' the HEREDOCs in a later PR. Since we do have a hard dep on activesupport (in the gemspec) I'm leaning towards enforcing the active support style, which would be the `MESSAGE.strip_heredoc` style since it would be rude to bump the ruby version requirement to 2.3 just for enforcing squiggly style `<<~MESSAGE`.

composerinteralia:
fb 5 won't officially support anything before Ruby 2.3, so the squiggly HEREDOC is a possibility.

PR #1207, opened 2018-10-02
[Rubocop] Lint Cop Offenses

goronfreeman:
Continue addressing #1202

Addresses all of the 'Lint' cops. For most of these fixes, I just used the suggested Rubocop fix, but on some others I took them one step further. For example, using `#each` instead of `#each_with_index` when the index wasn't used.

This is my first time contributing, so let me know if I missed any steps.

### DIFF
--- a/spec/acceptance/activesupport_instrumentation_spec.rb
+++ b/spec/acceptance/activesupport_instrumentation_spec.rb
@@ -1,6 +1,6 @@
 unless ActiveSupport::Notifications.respond_to?(:subscribed)
   module SubscribedBehavior
-    def subscribed(callback, *args, &_block)
+    def subscribed(callback, *args)
       subscriber = subscribe(*args, &callback)
       yield
     ensure
--- a/spec/acceptance/activesupport_instrumentation_spec.rb
+++ b/spec/acceptance/activesupport_instrumentation_spec.rb
@@ -1,6 +1,6 @@
 unless ActiveSupport::Notifications.respond_to?(:subscribed)
   module SubscribedBehavior
-    def subscribed(callback, *args, &block)
+    def subscribed(callback, *args, &_block)
       subscriber = subscribe(*args, &callback)
       yield
     ensure
--- a/spec/acceptance/activesupport_instrumentation_spec.rb
+++ b/spec/acceptance/activesupport_instrumentation_spec.rb
@@ -29,7 +29,7 @@ describe "using ActiveSupport::Instrumentation to track factory interaction" do
 
   it "tracks proper time of creating the record" do
     time_to_execute = 0
-    callback = ->(name, start, finish, id, payload) { time_to_execute = finish - start }
+    callback = ->(_name, start, finish, _id, _payload) { time_to_execute = finish - start }
     ActiveSupport::Notifications.subscribed(callback, "factory_bot.run_factory") do
       FactoryBot.build(:slow_user)
     end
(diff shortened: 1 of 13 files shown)


spec/acceptance/sequence_context_spec.rb
```diff
@@ -22,7 +22,7 @@ def awesome
   it 'invokes the correct method on the instance' do
     FactoryBot.define do
       factory :sequence_with_public_method, class: User do
-        sequence(:id) { |n| public_method(:awesome).call }
+        sequence(:id) { |_n| public_method(:awesome).call }
```
mikegee: I think when the block doesn't use the yielded value, it doesn't need to even account for it. This should work the same:

```ruby
sequence(:id) { public_method(:awesome).call }
```

spec/support/macros/define_constant.rb
```diff
@@ -28,7 +28,7 @@ def create_table(table_name, &block)
       connection.create_table(table_name, &block)
       @created_tables << table_name
       connection
-    rescue Exception => exception
+    rescue StandardError => exception
```
mikegee: Tiniest of nitpicks: perhaps rename this local variable to `error` to match?

spec/acceptance/sequence_context_spec.rb
goronfreeman: You're absolutely right. Thanks for pointing that out.

spec/support/macros/define_constant.rb
goronfreeman: 👍

spec/acceptance/activesupport_instrumentation_spec.rb
```diff
@@ -1,6 +1,6 @@
 unless ActiveSupport::Notifications.respond_to?(:subscribed)
   module SubscribedBehavior
-    def subscribed(callback, *args, &block)
+    def subscribed(callback, *args, &_block)
```
composerinteralia: I think we cant just leave of `&block` entirely, since we are using `yeild`

spec/factory_bot/evaluator_class_definer_spec.rb
```diff
@@ -13,7 +13,7 @@
   end
 
   it "evaluates the block in the context of the evaluator" do
-    dependency_attribute = double(
+    double(
```
composerinteralia: I we aren't ever using the variable, I'm guessing we can just get rid of this double entirely

spec/support/macros/define_constant.rb
```diff
@@ -28,9 +28,9 @@ def create_table(table_name, &block)
       connection.create_table(table_name, &block)
       @created_tables << table_name
       connection
-    rescue Exception => exception
+    rescue StandardError => error
```
composerinteralia: I think in this case rescuing`Exception` is appropriate because we are doing some cleanup and then re-raising the exception. If we use `StandardError` instead then terminating our program in the middle of this method might leave an unwanted table hanging around.

goronfreeman: I added a `rubocop:disable` comment for this instance, since I think it's still a good cop to have in most cases.

spec/acceptance/activesupport_instrumentation_spec.rb
goronfreeman: 👍

spec/factory_bot/evaluator_class_definer_spec.rb
goronfreeman: 👍

PR #1208, opened 2018-10-02
[Rubocop] Style Cop Offenses

goronfreeman:
Continue addressing #1202 

Addresses most of the 'Style' cops. I opted to skip a couple cops, because I don't have enough knowledge about them to know how to correctly fix them. The first is the `Style/MethodMissing` cop. I saw [this thread](https://github.com/rubocop-hq/rubocop/issues/3760) and thought that that same issue may apply. If anyone has a suggestion on what route to go on this, I can make the change if requested.

The other is the `Style/StringLiterals` cop. There is a big split on strings that use single and double quotes, so I'm not sure if we want to settle on one and use it. If someone tells me the preference, I can easily fix this as well.

I also had to omit fixing the `Style/RedundantBegin` cop because after making the change, Hound complained about invalid syntax. I'm guessing Hound must be checking for an older Ruby syntax before method blocks counted as an implicit `begin`. This could be fixed by setting a [`TargetRubyVersion`](https://github.com/rubocop-hq/rubocop/blob/master/manual/configuration.md#setting-the-target-ruby-version) in the Rubocop config file.

Last thing I want to point out is that I made a change to the Rubocop config based on the current style in the code. In spec files, curly braces are used for multi-line chained blocks with some `expect` blocks:

```ruby
expect {
  2.times { subject.define_attribute(attribute) }
}.to raise_error(...)
```

I omitted spec files for the `Style/BlockDelimiters` cop because this was used in a few different places in the specs. If a reviewer decides this isn't the desired fix, I can either use `do/end` blocks instead, or convert these to single-lines instead.

### DIFF
--- a/lib/factory_bot/linter.rb
+++ b/lib/factory_bot/linter.rb
@@ -56,7 +56,7 @@ module FactoryBot
       result = []
       begin
         FactoryBot.public_send(factory_strategy, factory.name)
-      rescue => error
+      rescue StandardError => error
         result |= [FactoryError.new(error, factory)]
       end
       result
@@ -67,7 +67,7 @@ module FactoryBot
       factory.definition.defined_traits.map(&:name).each do |trait_name|
         
           FactoryBot.public_send(factory_strategy, factory.name, trait_name)
-        rescue => error
+        rescue StandardError => error
           result |=
             [FactoryTraitError.new(error, factory, trait_name)]
         
--- a/lib/factory_bot/linter.rb
+++ b/lib/factory_bot/linter.rb
@@ -65,12 +65,12 @@ module FactoryBot
     def lint_traits(factory)
       result = []
       factory.definition.defined_traits.map(&:name).each do |trait_name|
-        begin
+        
           FactoryBot.public_send(factory_strategy, factory.name, trait_name)
         rescue => error
           result |=
             [FactoryTraitError.new(error, factory, trait_name)]
-        end
+        
       end
       result
     end
--- a/lib/factory_bot/linter.rb
+++ b/lib/factory_bot/linter.rb
@@ -65,12 +65,10 @@ module FactoryBot
     def lint_traits(factory)
       result = []
       factory.definition.defined_traits.map(&:name).each do |trait_name|
-        
-          FactoryBot.public_send(factory_strategy, factory.name, trait_name)
-        rescue StandardError => error
-          result |=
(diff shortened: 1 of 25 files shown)


spec/support/matchers/declaration.rb
```diff
@@ -48,7 +48,7 @@ def with_options(options)
     def failure_message
       [
         "expected declarations to include declaration of type #{@declaration_type}",
-        @options ? "with options #{options}" : nil
+        @options ? "with options #{options}" : nil,
```
mikegee: Perhaps out of scope for this PR, but I'd prefer if this ternary were replaced with 
```ruby
@options && "with options #{options}"
```

goronfreeman: I agree with you, but I do think it's out of the scope of the PR. Trying to keep it focused to only fixing Rubocop offenses.

lib/factory_bot/linter.rb
```diff
@@ -65,12 +65,12 @@ def lint_factory(factory)
     def lint_traits(factory)
       result = []
       factory.definition.defined_traits.map(&:name).each do |trait_name|
-        begin
+        
           FactoryBot.public_send(factory_strategy, factory.name, trait_name)
-        rescue => error
+        rescue StandardError => error
```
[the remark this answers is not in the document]
goronfreeman: I'm really not sure why it's complaining about this. I thought it was because I removed the `begin` keyword, but now that I added it back, Hound is still complaining. I think this means that I can fix the `Style/RedundantBegin` cop after all.

composerinteralia:
Thanks for keeping this branch up to date! I haven't had a chance to look at this one yet, but I will try to get to it soon.

composerinteralia:
I realized we needed 45bfe43 to get Hound to use our .rubocop.yml. That's why we are getting comments about Style/StringLiterals, for example, even though that cop is currently disabled. 

To answer your other questions:
* I haven't looked into the Style/MethodMissing cops. I think it would be fine to just leave those as they are, at least for now.
* Style/StringLiterals - double quotes. We should already get that from the thoughtbot configuration once we enable the cop again
* 45bfe43 may have fixed the problems you were having with Style/RedundantBegin
* I don't love the multi-line curly braces, but I vaguely recall there being a problem with `do/end` blocks and the `raise_error` expectation. I'll merge as is for now.

PR #1209, opened 2018-10-02
[Rubocop] Performance Cop Offenses

goronfreeman:
Continue addressing #1202 

Addresses all the 'Performance' cops. These were all straightforward to fix.

I originally fixed the `Performance/RegexpMatch` offense by using the suggested fix (`Regexp#match?`), but it turns out that the CI uses Ruby 2.3, which doesn't have that method. I fixed this by setting the `TargetRubyVersion` in `.rubocop.yml` to 2.3. The other way to do this is by adding a `.ruby-version` file to the project, then Rubocop will just use what version is specified there.

I looked into fixing the `Naming/MemoizedInstanceVariableName` cop as well in this PR, but it doesn't really fit into what Rubocop wants. The only thing I could think of would be to add a `rubocop:disable` comment, but I'm not sure if that's what the maintainers want.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,7 @@
 inherit_from:
   - https://raw.githubusercontent.com/thoughtbot/guides/master/style/ruby/.rubocop.yml
 AllCops:
+  TargetRubyVersion: 2.3
   Exclude:
     - 'gemfiles/*'
     - 'tmp/**/*'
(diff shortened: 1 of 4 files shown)


composerinteralia:
I looked into the memoization Cop. What do you think about https://github.com/thoughtbot/factory_bot/pull/1211?

I think setting 2.3 as the `TargetRubyVersion` makes sense. That's what we did over on factory_bot_rails as well.

PR #1210, opened 2018-10-03

viktorsmari:
I have had to look for this a couple of times. 
Would be nice to have a bulletproof Linux script version, as well as a MacOS way of find - replacing.

Answer is borrowed from here:
https://stackoverflow.com/questions/1583219/how-to-do-a-recursive-find-replace-of-a-string-with-awk-or-sed

### DIFF
--- a/UPGRADE_FROM_FACTORY_GIRL.md
+++ b/UPGRADE_FROM_FACTORY_GIRL.md
@@ -38,6 +38,11 @@ example, on macOS:
 grep -e FactoryGirl **/*.rake **/*.rb -s -l | xargs sed -i "" "s|FactoryGirl|FactoryBot|g"
 ```
 
+Linux:
+```sh
+find . -type f -print0| xargs -0 sed -i 's/FactoryGirl/FactoryBot/g'
+```
+
 ## Replace All Path References
 
 If you're requiring files from factory\_girl or factory\_girl\_rails directly,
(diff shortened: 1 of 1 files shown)


composerinteralia:
Many thanks! I appreciate the contribution, but I am going to close this PR with the same comment I left on a few others a while back:

> Thanks for this PR! The example we gave was really intended to be an example, not necessarily a perfect solution that would work for everyone. We have had lots of PRs suggesting alternatives (#1070, #1075, #1084, #1095, #1102), but to reduce churn on this file I have decided to close these issues. I appreciate your help!

viktorsmari:
I think people will continue to to create more pull requests for this issue if nothing is done. What about adding a link in the Readme to your answer including links to the other closed PRs? Just to offer some help

composerinteralia:
Yeah, that seems reasonable to me. Would you be willing to update this PR with that change?

viktorsmari:
Done. However I still don't fully understand what you mean by reducing churn?
Do you mean you were receiving a lot of PRs because the scripts were changing? The unix tools rarely change so if a script works, it should work for a long time. Also it does not make sense to offer a MacOS script, but denying a Linux script at the same time.

composerinteralia:
I think I meant "churn" in the sense of changing rapidly. Every PR opened has been a little different, and I suspect more would come in to cover various other edge cases and tools. I think there are other places (like stack overflow) that are better suited to covering how to do a reliable search and replace for whatever tool one happens to be using.

viktorsmari:
I understand. I however think if we provide the basic Linux script, we won't get so many PRs nitpicking what is the 'best' way to do it.

composerinteralia:
Many thanks. I changed the wording a little bit, squashed your commits into one, and merged as https://github.com/thoughtbot/factory_bot/commit/eb3edeb8b659e82d1332947ec46b9e910c4cd095

PR #1211, opened 2018-10-05
Use dedicated methods for memoized values

composerinteralia:
RuboCop didn't seem to like having memoization in the
`default_constants` method that didn't match the method name. This
satisfies RuboCop, and saves us an array allocation or two when we run
specs that don't use any of these helpers.


goronfreeman:
👍 I think this is a good approach. Thanks for tackling this one!

PR #1215, opened 2018-10-07
Rubocop: Fix Layout/IndentHeredoc issues

ChaoticBoredom:
I removed the exclusionary line from the rubocop file, and then ran `rubocop -a`.
On visually inspecting the resulting output, I manually adjusted the offenses in `lint_spec` as the autocorrect did some weird things with it.
I also took a look at the `Layout/SpaceInsideHashLiteralBraces` TODO, and couldn't find the offense. Was it possibly fixed in a different PR? Can we remove that line from the rubocop file list?

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -16,14 +16,6 @@ AllCops:
 # versions of RuboCop, may require this file to be generated again.
 
 # Offense count: 1
-# Cop supports --auto-correct.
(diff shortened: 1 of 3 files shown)


composerinteralia:
Yeah, that violation was probably removed as part of one of the other RuboCop PRs. Let's remove that TODO from the `.rubocop.yml`. Thanks for your help!

PR #1216, opened 2018-10-07
Rubocop: Fix Style/StringLiterals Offenses

ChaoticBoredom:
I moved this fix into it's own PR due to the noise of autocorrecting String Literals.
I removed the disabling of this cop, and then ran autocorrect.
Visually inspecting the results, everything looks hunky-dory, that being said, if this is too large/noisy, I'm more that happy to split this up further, and group the changes by file/folder etc.


composerinteralia:
Looks good to me. Many thanks!

PR #1217, opened 2018-10-15
Rubocop method missing fixes

krider2010:
More work towards https://github.com/thoughtbot/factory_bot/issues/1202

This adds in the `respond_to_missing?` where appropriate and uses super where possible. Exclusions are handled with an inline ignore where it just does not make sense to implement the rule. It's clear from the rubucop issue that not all rules apply in 100% of cases.

Any issues with the content of this PR, please do shout 😄 

Useful references:
* [Always Define respond_to_missing? When Overriding method_missing](https://robots.thoughtbot.com/always-define-respond-to-missing-when-overriding)
* https://github.com/rubocop-hq/rubocop/issues/3760 which covers how `super` is used when `send`ing the request on


krider2010:
Oops I missed some tests I think - my bad! On it!

lib/factory_bot/decorator/invocation_tracker.rb
```diff
@@ -11,6 +11,10 @@ def method_missing(name, *args, &block)
         super
       end
 
+      def respond_to_missing?(*)
```
composerinteralia: If all we are doing is calling `super`, this seems unnecessary to me and probably slightly worse for performance because of an extra method call. How would you feel about ignoring this one?

lib/factory_bot/definition_proxy.rb
```diff
@@ -84,6 +84,10 @@ def method_missing(name, *args, &block)
       end
     end
 
+    def respond_to_missing?(*)
```
composerinteralia: Given that the DefinitionProxy undefines `respond_to?` and the `method` method, I don't think `respond_to_missing?` makes sense for this class. I think we can just keep the exclude configuration for this file and get rid of the comment on line 75.

lib/factory_bot/decorator.rb
```diff
@@ -6,10 +6,14 @@ def initialize(component)
       @component = component
     end
 
-    def method_missing(name, *args, &block)
+    def method_missing(name, *args, &block) # rubocop:disable Style/MethodMissingSuper
       @component.send(name, *args, &block)
     end
 
+    def respond_to_missing?(name, include_private = false)
+      @component.respond_to?(name) || super
```
composerinteralia: `method_missing` is using `send`, rather than `public_send`, so I think we might need to use `@component.respond_to?(name, true)` here to account for any private or protected methods.

lib/factory_bot/decorator/invocation_tracker.rb
krider2010: I'd be OK with that

lib/factory_bot/decorator.rb
krider2010: I did wonder!

krider2010:
Appreciate there is required cleverness in this file, and as maintainers, you totally have the final say. But the reason I moved to inline `#disable` comments was to make it easier to see the local context and ideally not have the exclusions file. Because then any rule breaking needs to be allowed in situ and helps those newer to the codebase.

I'll fix up the couple of things above and circle back round 👍

krider2010:
Sorry for the delay @composerinteralia - pushed new changes.

composerinteralia:
Many thanks! I made a few changes (because we are on RuboCop 0.54, which doesn't include [this commit](https://github.com/rubocop-hq/rubocop/commit/46a179846480850d65ee2a307c4a2ab6555b971c)) and merged as https://github.com/thoughtbot/factory_bot/commit/c71aded77841e45198efa9932dd5f0af3171b3e8.

PR #1219, opened 2018-10-21
Raise more meaningful error for unregistered trait

composerinteralia:
Closes #970

In this code we are passing an implicit declaration `user`, rather than
the symbol `:user`:

```rb
factory :post do
  author factory: user
end
```

This will raise a confusing error:
`undefined method 'name' for :post:Symbol`.

This is coming from the implicit declaration `#==` method,
which ends up getting called [here](https://github.com/thoughtbot/factory_bot/blob/7925c47653b4e3c06661ffefac192efcf0c48144/lib/factory_bot/attribute_list.rb#L56).
This `#==` method wasn't ever designed to compare against
objects of different classes. I added some tests for all the declaration
classes to handle being compared against other kinds of objects.

Now we will get a more meaningful error `Trait not registered: user`

Co-authored-by: Dusan Orlovic <duleorlovic@gmail.com>

### DIFF
--- a/lib/factory_bot/declaration/association.rb
+++ b/lib/factory_bot/declaration/association.rb
@@ -10,7 +10,8 @@ module FactoryBot
       end
 
       def ==(other)
-        name == other.name &&
+        self.class == other.class &&
+          name == other.name &&
           options == other.options
       end
 
--- a/lib/factory_bot/declaration/dynamic.rb
+++ b/lib/factory_bot/declaration/dynamic.rb
@@ -8,7 +8,8 @@ module FactoryBot
       end
 
       def ==(other)
-        name == other.name &&
+        self.class == other.class &&
+          name == other.name &&
           ignored == other.ignored &&
           block == other.block
       end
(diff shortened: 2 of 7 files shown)


composerinteralia:
```
author factory: user
#=> "Trait not registered: user"
```

@iwz pointed out that `Trait not registered: user` doesn't make it obvious what to do in this case. It is better than `undefined method 'name' for :post:Symbol`, but it doesn't tell us we passed the wrong kind of thing to the `factory` option.

However, this does match the error we currently get if we make the same mistake with the `strategy` option:

```rb
author factory: :user, strategy: build
#=> "Trait not registered: build"
```
or with an override:

```rb
author factory: :user, name: name
#=> "Trait not registered: name"
```

So I am tempted use this same error here, and then maybe open a separate issue to improve the error messages for all of these at once (there might be a way to detect if any FactoryBot::Declarations are being passed around, for example).

composerinteralia:
Opened #1227

PR #1220, opened 2018-10-22
Avoid recursive call to AttributeHash#attributes

### LINKED ISSUE #1155: Stack overflow when referring to `attributes` attribute in `initialize_with`
Defining an `attributes` attribute works fine:

```rb
require 'factory_bot'

class Dog
  attr_writer :attributes
end

FactoryBot.define do
  factory :dog do
    attributes({ name: 'Ella', favorite_food: 'shoes' })
  end
end

FactoryBot.build(:dog)
```

but not when referring to it in `initialize_with`:

```rb
require 'factory_bot'

class Dog
  def initialize(attributes)
    @attributes = attributes
  end
end

FactoryBot.define do
  factory :dog do
    attributes({ name: 'Ella', favorite_food: 'shoes' })
    initialize_with { new(attributes) }
  end
end

FactoryBot.build(:dog)

# => SystemStackError: stack level too deep
```

I think we end up calling https://github.com/thoughtbot/factory_bot/blob/3256842209c87febc5413f16e12b9c0d043da910/lib/factory_bot/decorator/attribute_hash.rb#L9 recursively.

composerinteralia:
Fixes #1155

Defining an attribute called 'attributes', then referring to that
attribute in `initialize_with` was causing `AttributeHash#attributes`
to call itself recursively.

### DIFF
--- a/lib/factory_bot/attribute_assigner.rb
+++ b/lib/factory_bot/attribute_assigner.rb
@@ -31,7 +31,10 @@ module FactoryBot
     private
 
     def method_tracking_evaluator
-      @method_tracking_evaluator ||= Decorator::AttributeHash.new(decorated_evaluator, attribute_names_to_assign)
+      @method_tracking_evaluator ||= Decorator::AttributeHash.new(
+        decorated_evaluator,
+        attribute_names_to_assign,
+      )
     end
 
     def decorated_evaluator
--- a/lib/factory_bot/decorator/attribute_hash.rb
+++ b/lib/factory_bot/decorator/attribute_hash.rb
@@ -8,7 +8,7 @@ module FactoryBot
 
       def attributes
         @attributes.each_with_object({}) do |attribute_name, result|
-          result[attribute_name] = send(attribute_name)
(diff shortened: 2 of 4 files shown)


lib/factory_bot/attribute_assigner.rb
```diff
@@ -31,7 +31,10 @@ def hash
     private
 
     def method_tracking_evaluator
-      @method_tracking_evaluator ||= Decorator::AttributeHash.new(decorated_evaluator, attribute_names_to_assign)
+      @method_tracking_evaluator ||= Decorator::AttributeHash.new(
```
composerinteralia: This change is not strictly related, but this line was long and I happened to be looking here as I was following the page through to the AttributeHash

mike-burns:
LGTM.

Does this mean that we can't name an attribute `initialize`? Not that I want to...

composerinteralia:
Yeah, I don't think adding an `initialize` attribute has ever been possible, and I think I am OK with that.

Issue #1142, opened 2018-06-12
Confusion around aliases for `{attribute}` and `{attribute}_id`


composerinteralia:
We have had several issues over the years related to `FactoryBot.aliases`: #522 #734 #851 #989 #1096 #1138. In most of these issues people do not realize that `factory_bot` assumes`{attribute}_id` will be the foreign key of an `{attribute}` `ActiveRecord` association.

At the very least I think we need better documentation around this. We could probably also offer better configuration for this, and maybe move anything specific to `ActiveRecord` into `factory_bot_rails`.

shyam-habarakada:
Just wasted a few hours because of this. The workaround I applied (based on other issues linked and other research) was as below. 

```ruby
factory :person
  name
  email
  ...

  transient do
    thing_id 0
    thing nil # Thing.to_s or similar, denormalized here
  end

  after :create do |person, options|
    person.thing_id = options.thing_id
    person.thing = options.thing
  end
end
```

Feedback on this would be great. Thanks

composerinteralia:
Yup, that seems to match the solution in #1096. I think you could use `after(:build)` instead of an `after(:create)`, then get rid of the call to `save!`.

agirdler-lumoslabs:
Is there a planned "official" solution to this? While hacking `transient` and `after` callbacks works in most cases I haven't been able to get it to work properly with a hanami/model backed model (the object is [frozen during initialization](https://github.com/hanami/model/blob/2df710ff87bcf67615679114a00a119ab9c5fc0e/lib/hanami/entity.rb#L124)) and instead resorted to hacking an `non_aliasable` method on `FactoryBot` then using that in a monkey patched definition of `FactoryBot.aliases_for`:

```ruby
module FactoryBot
  def self.non_aliasable
    %w[processor processor_id]
  end

  def self.aliases_for(attribute)
    aliases.map do |(pattern, replace)|
      if pattern.match(attribute.to_s) && !non_aliasable.include?(attribute.to_s)
        attribute.to_s.sub(pattern, replace).to_sym
      end
    end.compact << attribute
  end
end
```

This has worked great so far for the project I'm working on, but it doesn't really feel maintainable (as it's a list stored in a module). I'm also not familiar enough with the code to add a method to the DSL to allow defining these on a per model basis but I think that would be the most flexible solution (but may require a rewrite so `FactoryBot.aliases_for` is builder aware?). Are there any opinions on this approach?

composerinteralia:
[...]

PR #1224, opened 2018-10-26
WIP - aliases interfere with fields ending in _id with no association

Dishwasha:
If a field name ends in _id and isn't intended for an association,
the default FactoryBot.aliases interfere.

There doesn't seem to be a clear way to fix this problem.  One idea
would be to have a way in a factory definition to mark an attribute
to not be aliased.

The committed spec will fail with the default aliases.  The user can
always override this with the following to get tests passing:

`FactoryBot.aliases = [[/^((?!item_type).)_id/, '\1']]`

This is feasible but not an ideal situation.

### DIFF
--- a/spec/acceptance/aliases_spec.rb
+++ b/spec/acceptance/aliases_spec.rb
@@ -14,4 +14,23 @@ describe "aliases and overrides" do
   subject { FactoryBot.create(:user, one: "override") }
   its(:one) { should eq "override" }
   its(:two) { should be_nil }
+
+  context "don't alias" do
+    before do
+      # Enable this to get the tests to pass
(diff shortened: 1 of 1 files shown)


Dishwasha:
Looking for feedback on this bug and what direction I might take to fix.

composerinteralia:
It looks like you ran into this: https://github.com/thoughtbot/factory_bot/issues/1142

There is a workaround there that might work for you. I don't have a great permanent solution.

composerinteralia:
I'm going to close this for now, since it is currently expected behavior. We can continue any further discussion on #1142. Thanks!

PR #1225, opened 2018-10-28
Set timestamps before clearing changes

composerinteralia:
## Purpose:
To get timestamps for build_stubbed objects to behave the same as timestamps for real, persisted objects

## Details:

e7f66fb2 fixed one problem and introduced a new one. Setting timestamps
after clearing changes information doesn't work, since using the
created_at and updated_at setters will change the object again. I added
some additional specs for this case (our only spec related to clearing
changes used an object without timestamps).

For the timezone problem, I did a little digging and realized the
casting behavior was actually correct. The reason it was behaving in a
different way than I was expecting is because we had
`time_zone_aware_attributes` set to false, whereas the AR railtie
[sets it to true](https://github.com/rails/rails/blob/54652d886ad1c9ffca36c6f2baadcdd6135ef4dd/activerecord/lib/active_record/railtie.rb#L71). Setting it to true causes date and datetime attributes
[to get decorated with a TimeZoneConverter](https://github.com/rails/rails/blob/54652d886ad1c9ffca36c6f2baadcdd6135ef4dd/activerecord/lib/active_record/attribute_methods/time_zone_conversion.rb#L68..L87)

### DIFF
--- a/spec/acceptance/build_stubbed_spec.rb
+++ b/spec/acceptance/build_stubbed_spec.rb
@@ -184,46 +184,21 @@ describe "defaulting `created_at`" do
     expect(stub.created_at).to eq 3.days.ago
   end
 
-  context "when using time_zone_aware_attributes" do
-    it "defaults created_at to the local time zone" do
-      Time.zone = "Eastern Time (US & Canada)"
-      ActiveRecord::Base.time_zone_aware_attributes = true
-
-      stub = build_stubbed(:thing_with_timestamp)
-
-      expect(stub.created_at.zone).to eq "EST"
+  it "behaves the same as a non-stubbed created_at" do
+    define_model("ThingWithCreatedAt", created_at: :datetime) do
+      def created_at
+        :the_real_created_at
+      end
     end
 
-    it "casts created_at override to local time zone" do
-      Time.zone = "Eastern Time (US & Canada)"
-      ActiveRecord::Base.time_zone_aware_attributes = true
-      created_at = Time.now.in_time_zone("CET")
-
-      stub = build_stubbed(:thing_with_timestamp, created_at: created_at)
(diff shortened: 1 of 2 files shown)


spec/acceptance/build_stubbed_spec.rb
```diff
@@ -172,7 +167,9 @@
   end
 
   it "allows overriding created_at for objects with created_at" do
-    expect(build_stubbed(:thing_with_timestamp, created_at: 3.days.ago).created_at).to eq 3.days.ago
```
composerinteralia: Not totally related, but this line was really long and it was bothering me.

composerinteralia:
One possibility is to replace all the timezone tests with something like:

```rb
it "does not override the original created_at method" do
  define_model("ThingWithCustomTimestampReader", created_at: :datetime) do
    def created_at
      if super.present?
        "Created at was set"
      end
    end
  end

  FactoryBot.define do
    factory :thing_with_custom_timestamp_reader
  end

  stub = build_stubbed(:thing_with_custom_timestamp_reader)

  expect(stub.created_at).to eq("Created at was set")
end
```

This is making sure our build stubbed `created_at` method will return whatever the original `created_at` method returns (which is easy, since we aren't overriding the method anymore).

composerinteralia:
OK, I dropped the timezone tests and added a test that makes sure `created_at` is the same for a stubbed and a non-stubbed object.

mike-burns:
OK.

Do we have an app we can try this on? It's a surprisingly invasive change and I wouldn't want it to break in a public release.

composerinteralia:
Passing tests on upcase, hub, and my current client project. I can also put together a rc for factory_bot 5 in a couple weeks.

PR #1233, opened 2018-11-19
Add option for verbose linting

composerinteralia:
This has come up a few times, and I can see why it might be helpful to
have access to full backtraces when debugging a factory error uncovered
by `FactoryBot.lint`. But since most of the time I don't want the extra
noise from the backtrace, I added this as a verbose option.

The default message is still:

```
The following factories are invalid:
* user - undefined method `save!' for #<User:0x00007ff0cbc89100>
* admin - undefined method `save!' for #<User:0x00007ff0cbc73e40>
```

And with the verbose option (usually with more lines of backtrace):

```
The following factories are invalid:
* user - undefined method `save!' for #<User:0x00007ff0cbc89100>
  /Users/.../factory_bot/lib/factory_bot/evaluation.rb:18:in `create'
  /Users/.../factory_bot/lib/factory_bot/strategy/create.rb:12:in `block in result'

* admin - undefined method `save!' for #<User:0x00007ff0cbc73e40>
  /Users/.../factory_bot/lib/factory_bot/evaluation.rb:18:in `create'
  /Users/.../factory_bot/lib/factory_bot/strategy/create.rb:12:in `block in result'
```

I moved the linting option defaults out of the FactoryBot.lint method
and into keyword argument defaults in Linter#initialize. This seems a
bit cleaner, and now we will get an error if we pass an option we don't
understand (before 6e51159 we had a test that passed in a bogus
option)

Closes #710
Closes #1124

I am opening a new PR since the original PR is years old and it seemed
unkind to request changes after so long. Instead I will list the authors
as co-authors.

Co-authored-by: Jack Kinsella <jack.kinsella@gmail.com>
Co-authored-by: Jasper Woudenberg <mail@jasperwoudenberg.com>

### DIFF
--- a/lib/factory_bot/linter.rb
+++ b/lib/factory_bot/linter.rb
@@ -1,9 +1,10 @@
 module FactoryBot
   class Linter
-    def initialize(factories, strategy: :create, traits: false)
+    def initialize(factories, strategy: :create, traits: false, verbose: false)
       @factories_to_lint = factories
       @factory_strategy = strategy
       @traits = traits
+      @verbose = verbose
       @invalid_factories = calculate_invalid_factories
     end
 
@@ -34,6 +35,13 @@ module FactoryBot
       def message
         message = @wrapped_error.message
         "* #{location} - #{message} (#{@wrapped_error.class.name})"
+      end
+
+      def verbose_message
+        <<~MESSAGE
+          #{message}
+            #{@wrapped_error.backtrace.join("\n  ")}
+        MESSAGE
       end
 
       def location
@@ -85,7 +93,7 @@ module FactoryBot
 
     def error_message
       lines = invalid_factories.map do |_factory, exceptions|
-        exceptions.map(&:message)
+        exceptions.map(&error_message_type)
       end.flatten
 
       <<~ERROR_MESSAGE.strip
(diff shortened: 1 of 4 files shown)


lib/factory_bot/linter.rb
```diff
@@ -52,6 +60,15 @@ def location
       end
     end
 
+    def lint(factory)
+      errors = lint_factory(factory)
+      if @traits
+        errors | lint_traits(factory)
```
joshuaclayton: The `|` operator always throws me - is this intended or would `||` work?

lib/factory_bot/linter.rb
```diff
@@ -1,9 +1,10 @@
 module FactoryBot
   class Linter
-    def initialize(factories, linting_strategy, factory_strategy = :create)
+    def initialize(factories, strategy: :create, traits: false, verbose: false)
```
joshuaclayton: There's a fair amount of logic shift in this PR in addition to the verbose flag. Would it make sense to refactor this to options (sans-`verbose`) in one PR and then add the `verbose` flag as a separate PR?

composerinteralia: Yup, good point

lib/factory_bot/linter.rb
composerinteralia: It was there before (as `|=`) and I didn't think to much about it. I will figure out whether it is necessary.

PR #1234, opened 2018-11-21
Document use_parent_strategy configuration

hmayer00:
Add documentation to the associations section to mention the `use_parent_strategy` configuration, and demonstrate it.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -384,6 +384,28 @@ factory :post do
   author strategy: :build    # <<< this does *not* work; causes author_id to be nil
 ```
 
+To have unspecified associations use the parent's strategy, instead of using :create, you can use the configuration `use_parent_strategy`:
+
+```ruby
+FactoryBot.use_parent_strategy = true
+
+factory :post do
+  # ...
+  author
+end
+
+post = build(:post)
+post.new_record?        # => true
+post.author.new_record? # => true
+
+post = create(:post)
+post.new_record?        # => false
+post.author.new_record? # => false
+```
+
+If you are using rspec, you can set this configuration in your spec_helper.rb (or rails_helper.rb, if using Rails).
+If you want to set it globally, you can use an intializer (if using Rails).
+
 Generating data for a `has_many` relationship is a bit more involved,
 depending on the amount of flexibility desired, but here's a surefire example
 of generating associated data.


composerinteralia:
Thanks! I have been pondering having this set to true by default in factory_bot 5. I have enabled this in several projects without any issue.

hmayer00:
Thanks for accepting so quickly! And 👍 for having that be the default, I think it would save people a lot of thumb-twiddling time.

I actually just found one (fairly minor?) issue with the toggling of it, which has to do with FactoryBot configuration in general. I like using FactoryBot in my dev console, which means I do a certain amount of `FactoryBot.reload`. It looks like this config gets wiped when you do reload - `reload` calls `reset_configuration`, which does what one would expect.

There isn't an obvious quick change to that - `configuration` involves a lot of things that *should* get reset on reload, so we don't want to just make things sticky. There would need to be some separation between the aspects of config that are added programmatically (like factory definitions or callbacks) and those that are added explicitly. It seems like there were recently 3 settable configurations, but with  `duplicate_attribute_assignment_from_initialize_with` and `allow_class_lookup` deprecated and removed, `use_parent_strategy` looks like it's the only one remaining?

Anyway, I could take a swing at switching the way that sort of settable configuration works, if you want. Probably something like:
```ruby
FactoryBot.configure(use_parent_strategy:  true)
```
or:
```ruby
FactoryBot.configure do |config|
  config.use_parent_strategy = true
end
```

That might be useful if you think there might be more user-configurable options in the future; it would deal with the reload problem, and probably make configuration more discoverable. But if you expect user-configuration to be extremely rare and would rather avoid the complexity, then no worries.

Anyway, lemme know if you want me to take a pass at that, and if so, whether you have a preference for one of those approaches over the other (or something completely different!). The first one is probably easier to implement but the second seems to be a more common approach.

hmayer00:
Alternatively, it could just leave the mechanism the same, but store that type of configuration in a separate object which is retained on reload. Something like (in Configuration):
```ruby
@user_configuration = UserConfiguration.new

delegate :use_parent_strategy,
         :use_parent_strategy=,
         to: :user_configuration
```

Just a thought. Thanks again!

composerinteralia:
Thanks! I opened https://github.com/thoughtbot/factory_bot/issues/1236 to track `use_parent_strategy`. I think you are thinking about the user configuration the right way, but we might not need it for now if we get deprecate `use_parent_strategy`.

composerinteralia:
@hmayer00 looking back I realize you had offered to try to address the problem with `use_parent_strategy` and reloading. I ended up merging a quick fix in https://github.com/thoughtbot/factory_bot/pull/1244. That is probably good enough for now, but if we introduce any other configuration options I may ask for you help. Many thanks!

hmayer00:
@composerinteralia Sorry for the delayed response! I'd be very happy to give something back to this extremely useful gem, so if you decide you want to go that way I'll jump in. Thanks for all that you great folks at thoughtbot do for us!

composerinteralia:
Thanks!

PR #961, opened 2016-10-26
Add use_parent_strategy option for building associations

rringler:
### Description

Picking up where https://github.com/thoughtbot/factory_girl/pull/749 left off.

This PR adds a `FactoryGirl.use_parent_strategy` configuration option, allowing factory associations to follow the parent's build strategy.  Previously, all factory associations were
created, regardless of whether the parent was persisted to the database.

The configuration option is off by default, so there is no change to the current behavior.

PR #1240, opened 2018-12-07
Always use parent strategy

### LINKED ISSUE #1236: Maybe deprecate `use_parent_strategy` configuration
https://github.com/thoughtbot/factory_bot/pull/1234#issuecomment-440902652 got me thinking more about this.

> I like using FactoryBot in my dev console, which means I do a certain amount of FactoryBot.reload. It looks like this config gets wiped when you do reload - reload calls reset_configuration, which does what one would expect.

That is a really good point! factory_bot_rails currently has `Spring.after_fork { FactoryBot.reload }` and factory_bot_rails 5 will have https://github.com/thoughtbot/factory_bot_rails/commit/02a0f584876ff5c7a6d6fa4965da58ed70464a73, which will call `FactoryBot.reload` whenever you change a factory definition or whenever you call `reload!` in the rails console. So I suspect this will cause problems at some point.

Right now I am tempted to get rid of the setting altogether for factory_bot 5 and just use the parent strategy all the time. This may cause some test failures for people, but they would be able to fix them by changing `build` to `create` for the failing tests.

composerinteralia:
Closes #1236

Background
---

In issues #64 and #66 (from 2010) people expressed surprise
that using the `build` strategy would still `create` any associations.
I remember being similarly surprised by that behavior when I first
started using factory_bot.

The main reason for this behavior is to ensure the built instance will
be valid if there were any foreign key validations
(like `validates_presence_of :user_id`). If we don't save the
associations, there won't be any ids present.

PR #191 (from 2011) offers a workaround for people who don't want
records to be saved automatically. Passing `strategy: :build`
(originally `method: :build`, but later renamed) when declaring the
association will prevent it from being saved when using `build`.

But #191 isn't really a complete solution (as discussed on the PR
after it was merged). `strategy: :build` may do
the right thing when building objects with the `build` strategy, but it
can cause new problems when using the `create` strategy.

A better solution would be something like:
`strategy: <whatever strategy I was already using>`.
PRs #749 and #961 (merged in 2016) introduce something like that,
with the `use_parent_strategy` configuration option.
With this turned on `build` end up being generally [a little faster][]
than `build_stubbed`, since it no longer needs to hit the database for
each association.

[a little faster]: https://gist.github.com/composerinteralia/d4796df9140f431e36f88dfb6fe9733a

I have set `use_parent_strategy` on several projects now. I also added
it to suspenders in thoughtbot/suspenders#952.
On newer projects I have not run into any problems. On existing
projects I have seen the occasional test failure, which are easy enough
to fix by changing `build` to `create`.

Unfortunately I don't think `use_parent_strategy` is widely known,
since it wasn't documented until #1234 (about a month ago).
I also learned in #1234 that the `use_parent_strategy` setting gets
wiped out with `FactoryBot.reload`, which can be problematic when using
factory_bot_rails.

To summarize, we have been exploring having a `build` strategy that uses
`build` all the way down since 2010, but there still isn't a totally
reliable way to get that.

This PR
---

* Default to using the parent strategy for factory_bot 5, as suggested
in #749.
* Get rid of the `use_parent_strategy` option, since it doesn't play
well with `FactoryBot.reload`

I think this PR will also make the `strategy` option unnecessary, but I
would want to explore deprecating that in a separate PR.

If dealing with foreign key validations is important, maybe we can
rethink that from the ground up.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -347,65 +347,6 @@ factory :post do
 end
 ```
 
-The behavior of the association method varies depending on the build strategy used for the parent object.
-
-```ruby
-# Builds and saves a User and a Post
-post = create(:post)
-post.new_record?        # => false
-post.author.new_record? # => false
-
-# Builds and saves a User, and then builds but does not save a Post
-post = build(:post)
-post.new_record?        # => true
-post.author.new_record? # => false
-```
-
-To not save the associated object, specify strategy: :build in the factory:
-
-```ruby
-factory :post do
-  # ...
-  association :author, factory: :user, strategy: :build
-end
-
-# Builds a User, and then builds a Post, but does not save either
-post = build(:post)
-post.new_record?        # => true
-post.author.new_record? # => true
-```
-
-Please note that the `strategy: :build` option must be passed to an explicit call to `association`,
-and cannot be used with implicit associations:
-
-```ruby
-factory :post do
-  # ...
-  author strategy: :build    # <<< this does *not* work; causes author_id to be nil
-```
-
-To have unspecified associations use the parent's strategy, instead of using :create, you can use the configuration `use_parent_strategy`:
-
-```ruby
-FactoryBot.use_parent_strategy = true
-
-factory :post do
-  # ...
-  author
-end
-
-post = build(:post)
-post.new_record?        # => true
-post.author.new_record? # => true
-
-post = create(:post)
-post.new_record?        # => false
-post.author.new_record? # => false
-```
-
-If you are using rspec, you can set this configuration in your spec_helper.rb (or rails_helper.rb, if using Rails).
-If you want to set it globally, you can use an intializer (if using Rails).
-
 Generating data for a `has_many` relationship is a bit more involved,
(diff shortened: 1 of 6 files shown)


joshuaclayton:
@composerinteralia ah, that's interesting, and great point. My hesitancy was more around introducing the breaking change (changing the default) but also not being able to change that at all, so there's no "halfway." Would it be possible to include a warning if a combination of overridden `use_parent_strategy = false` and `reload` is called instead of removing the option entirely?

This might not also be an issue. In theory major bumps do constitute more work on the developer to upgrade, my approach here in thinking is that there might be some ways to reduce some of that burden (assuming there is any!)

composerinteralia:
@joshuaclayton my new plan is set `use_parent_strategy` on `FactoryBot` directly so it isn't affected by `reload` (see https://github.com/thoughtbot/factory_bot/pull/1244), and then update this PR to simply change the default to `true`.

I will hold of on any deprecations and maybe introduce those in FB 5.1

composerinteralia:
Removed the deprecation and merged as d0208ed

PR #1242, opened 2019-01-02
Easy transition from Factory Girl To Factory Bot

Olli:
I described a way more easy way than find and replace


composerinteralia:
Thanks for the PR. I would rather not offer this as the official solution, since it would preserve references to the old, problematic name.

PR #1243, opened 2019-01-03
 docs: Make doc more clear for spec in rails.

lanzhiheng:
Some new use in factory_bot may forget to add the require to the config file. in addition they don't know the locales of the config file.


lanzhiheng:
@composerinteralia  I place the code in `spec/rails_helper.rb` It works find for me. Thank you very much.  👍

PR #1244, opened 2019-01-03
Do not reset use_parent_strategy on reload

composerinteralia:
Before this PR, `use_parent_strategy` was set on the configuration
instance. Since `FactoryBot.reload` wipes out the configuration, it also
ends up resetting `use_parent_strategy` back to `nil`.

This can cause problems when using factory_bot_rails with Spring, since
factory_bot_rails calls `FactoryBot.reload` each time Spring forks. If
`use_parent_strategy` is set in a file that Spring preloads, like in an
initializer, then the value will get wiped out.

With this PR, we set `use_parent_strategy` directly on FactoryBot,
rather than on the configuration instance, and so reloading no longer
has any effect.

Related to #1236 and #1240

### DIFF
--- a/spec/acceptance/build_spec.rb
+++ b/spec/acceptance/build_spec.rb
@@ -21,8 +21,8 @@ describe "a built instance" do
 
   it { should be_new_record }
 
-  context "when the :use_parent_strategy config option has not been set" do
-    before { FactoryBot.use_parent_strategy = nil }
+  context "when the :use_parent_strategy config option is set to false" do
+    before { FactoryBot.use_parent_strategy = false }
 
     it "assigns and saves associations" do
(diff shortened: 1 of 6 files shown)


spec/acceptance/build_spec.rb
```diff
@@ -21,30 +21,23 @@
...
+  context "when the :use_parent_strategy config option is set to true" do
     before { FactoryBot.use_parent_strategy = true }
 
     it "assigns but does not save associations" do
       expect(subject.user).to be_kind_of(User)
       expect(subject.user).to be_new_record
     end
   end
-
-  it "assigns but does not save associations when using parent strategy" do
```
composerinteralia: This is a duplicate of the test above

PR #1249, opened 2019-01-08
Add documentation and YARD method directives

hjwylde:
> FactoryBot::Syntax::Methods was missing some documentation on the null
strategy and the *_pair methods. This adds in the documentation and
corresponding method directives.
>
> Further, add the missing &block parameter to the existing *_list directives.

This pull request improves the documentation available, so that users don't need to look through the source code to find out that there are `null` and `*_pair` methods available.

My personal interest in this is because RubyMine is able to use YARD documentation to help provide suggestions in auto completion, resolve methods, and to provide method documentation. Adding these missing directives will help improve the user experience.

### DIFF
--- a/lib/factory_bot/syntax/methods.rb
+++ b/lib/factory_bot/syntax/methods.rb
@@ -2,8 +2,8 @@ module FactoryBot
   module Syntax
     ## This module is a container for all strategy methods provided by
     ## FactoryBot. This includes all the default strategies provided ({Methods#build},
-    ## {Methods#create}, {Methods#build_stubbed}, {Methods#attributes_for}, and
-    ## {Methods#null}), as well as the complementary *_list and *_pair methods.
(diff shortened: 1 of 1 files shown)


lib/factory_bot/syntax/methods.rb
```diff
@@ -51,22 +51,51 @@ module Methods
       # Generates a hash of attributes for a registered factory by name.
       # @return [Hash] hash of attributes for the factory
 
-      # @!method build_list(name, amount, *traits_and_overrides)
+      # @!method null(name, *traits_and_overrides, &block)
```
composerinteralia: Is there a use case for using the `null` strategy directly? It always returns `nil`. I am tempted to leave out the documentation for `null`.

hjwylde: Good call, I actually didn't realise it was just for internal purposes. I can't think of a purpose either so I'll remove the documentation for `null`.

hjwylde:
:+1: thanks!

PR #1251, opened 2019-01-11
Set use_parent_strategy default to true explicitly

eugeneius:
Followup to https://github.com/thoughtbot/factory_bot/commit/d0208eda9c65cbc476a02d2f7503234195610005.

The `default:` option of `mattr_accessor` wasn't added until Rails 5.2, which means it has no effect on older versions and the default value of `FactoryBot.use_parent_strategy` in 5.0.0.rc1 is nil on those versions.

### DIFF
--- a/lib/factory_bot.rb
+++ b/lib/factory_bot.rb
@@ -57,7 +57,8 @@ module FactoryBot
     @configuration = nil
   end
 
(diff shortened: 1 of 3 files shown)


composerinteralia:
Thanks, Eugene! Good catch.

PR #1253, opened 2019-01-29
add details of where to paste configuration

BKSpurgeon:
## Why this change?

* It may not altogether be clear for newcomers to know where to paste the relevant lines of code - but the seasoned developer probably would not think twice. This PR will hopefully make it explicitly clear. I hope it helps.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -33,8 +33,10 @@ Configure your test suite
 If you're using Rails:
 
 ```ruby
+#  spec/rails_helper.rb
+
 RSpec.configure do |config|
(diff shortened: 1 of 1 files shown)


composerinteralia:
Thanks for the PR!

I am probably going to close this along the same lines as https://github.com/thoughtbot/factory_bot/pull/1243. We don't specify where to put the configuration for the other testing libraries, and suggesting putting it in `spec/support/factory.rb` (which is what we generally do at thoughtbot) ended up causing some confusion (see https://github.com/thoughtbot/factory_bot/pull/1141).

PR #1271, opened 2019-02-26
Fix Markdown formatting

tysongach:
The list was not enumerating when rendered, because the indentation was only two spaces, not four.


composerinteralia:
Also, factory_bot_rails probably has the same problem

tysongach:
@composerinteralia Yes, good call on both. I’ll merge this and submit PRs to those repos, as well.

Issue #1226, opened 2018-10-31
Improve documentation for polymorphic associations


composerinteralia:
Let's say we have a Video and a Photo model, and both `have_many` Comments:

```rb
class Photo < ApplicationRecord
  has_many :comments, as: :commentable
end 

class Video < ApplicationRecord
  has_many :comments, as: :commentable
end 

class Comment < ApplicationRecord
  belongs_to :commentable, polymorphic: true
end
```

I might write a factory like this:

```rb
FactoryBot.define do
  factory :video
  factory :photo

  factory :comment do
    for_photo

    trait :for_video do
      association(:commentable, factory: :video)
    end

    trait :for_photo do
      association(:commentable, factory: :photo)
    end
  end
end
```

This allows us to create a valid comment for either a video or a photo, and we default to the `for_photo` trait if we don't need a specific type.

Adding an example like this to the documentation might help with questions like https://github.com/thoughtbot/factory_bot/issues/1171 and https://github.com/thoughtbot/factory_bot_rails/issues/232.

joshm1204:
A few questions:

1) do I need to create a separate video.rb factory file and a photo.rb factory file? and does   `factory :video` and` factory :photo` just reference them?  OR do I define my video factory and photo factory as follows

2) Why isn't there a `for_video` method after   `factory :comment do` and before  `for_video`

3) If all the factories are combined into one file, is there a naming convention for the single combined file?

 ```
FactoryBot.define do
  factory :video do

  end 
  factory :photo do
  
  end

  factory :comment do
    for_photo

    trait :for_video do
      association(:commentable, factory: :video)
    end

    trait :for_photo do
      association(:commentable, factory: :photo)
    end

   end
end
```

**Comment by onerinas** [comment]
I have the same question
> 2. Why isn't there a for_video method after factory :comment do and before for_video

**Comment by onerinas** [comment]
@joshm1204 My guess would be: by default, it's set to `for_photo`

ie, `create(:comment)` would create a comment for photo and if you want to create for video, you may use `create(:comment, :for_video)`

**Comment by composerinteralia** [comment]
@joshm1204 https://github.com/thoughtbot/factory_bot/pull/1273 should help with your second question.


For the other questions, you can choose whether to define all your factories in one file or to split them out into separate files. At thoughtbot we tend to use a single spec/factories.rb

By default, factory_bot will look for definitions in:

```
test/factories.rb
spec/factories.rb
test/factories/*.rb
spec/factories/*.rb
[...]

PR #1273, opened 2019-03-15
Code style consistency for Polymorphic association

onerinas:
We have explicit association written as `association :author, factory: :user` everywhere 
except for polymorphic `association(:commentable, factory: :video)`

This PR aims to keep it consistent

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -519,11 +519,11 @@ FactoryBot.define do
     for_photo
 
     trait :for_video do
-      association(:commentable, factory: :video)
+      association :commentable, factory: :video
     end
 
     trait :for_photo do
-      association(:commentable, factory: :photo)
+      association :commentable, factory: :photo
     end
   end
 end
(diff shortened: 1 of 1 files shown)


GETTING_STARTED.md
```diff
@@ -532,7 +532,8 @@ end
 This allows us to do:
 
 ```
-create(:comment)
+# Would be created for :photo by default since we have `for_photo` after :comment factory definition
```
onerinas: I think it's good to have some notes considering I'm not the only one who took some time to understand why `for_video` is missing in the definition (ref: https://github.com/thoughtbot/factory_bot/issues/1226#issuecomment-470272866)
I'm open to reverting this change if needed.

composerinteralia: Could we add this comment inline on line 519? I think we could save some words that way:

```rb
factory :comment do
  for_photo # default to the :for_photo trait if none is specified
```

onerinas: that's way better!

composerinteralia:
Thanks!

PR #1277, opened 2019-03-24
Add setup comment

softwaregravy:
For new users who may trip over setup

I (was moving too fast and) put this in spec_helper and got an error. I ended up on [stackoverflow](https://stackoverflow.com/questions/48091582/ruby-gem-uninitialized-constant-factorybot) to fix the problem. Suggesting this line to help future users.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -30,7 +30,7 @@ Configure your test suite
 
 ### RSpec
 
-If you're using Rails:
+If you're using Rails, add the following configuration to `spec/support/factory_bot.rb` and be sure to require that file in `rails_helper.rb`:
 
 ```ruby
 RSpec.configure do |config|


GETTING_STARTED.md
```diff
@@ -33,6 +33,7 @@ Configure your test suite
 If you're using Rails:
 
 ```ruby
+# in new file spec/support/factory_bot.rb or add the config line to rails_helper.rb
```
composerinteralia: Let's remove the comment and change line 33 to:

```
If you're using Rails, add the following configuration to `spec/support/factory_bot.rb` and be sure to require that file in `rails_helper.rb`:
```

softwaregravy:
Pushed update

composerinteralia:
Thanks!

PR #1278, opened 2019-04-05
Update README.md: remove StackOverflow old tag mention

D-system:
Stackoverflow redirect to the new tag name: "factory-girl" gets redirected to "factory-bot"
No needs to mention the old tag name. The existing tag is line 58 (4 lines above).


composerinteralia:
Thanks!

PR #1284, opened 2019-04-18
Use GH to show the GETTING_STARTED

drapergeek:
This saves the user the need to navigate to another site.


composerinteralia:
Thanks!

If somebody is looking at at the README at https://www.rubydoc.info/gems/factory_bot/file/README.md then this change will have the opposite effect. That said, I am in favor of this change because I am more likely to look at the README on GitHub.

We should update the "Interested in the history of the project name?" and "License" links as well. For the "License" we are using a relative link, which doesn't work at all on rubydoc.

drapergeek:
Thanks for handling this @composerinteralia, sorry for ghosting on you!

Issue #1281, opened 2019-04-12
Move internal methods to FactoryBot::Internal


composerinteralia:
We introduced `FactoryBot::Internal` in https://github.com/thoughtbot/factory_bot/pull/1262 so we could handle inline sequences in https://github.com/thoughtbot/factory_bot/pull/1258 without adding additional public methods on `FactoryBot`.

I suspect there are quite a few class methods on `FactoryBot` that don't need to be public, and could be deprecated and then moved to `FactoryBot::Internal`.

composerinteralia:
@aledustet should we close this now?

PR #1285, opened 2019-04-19
Move and deprecate trait methods from FactoryBot

aledustet:
Why:
These are essentially internal methods that should not be publicly
available from the base namespace.
One thing worth noticing is that the use of these methods internally was
almost exclusively in the `syntax/default` except for one use on the
`factory_bot/definition`.
Also, the deprecation silencing module was referring to the singleton
an instance of the ```ActiveRecord:: Deprecation``` class and not to the new
deprecation instance that was being used in the ```FactoryBot``` module.

This PR:
- Moves the `trait_by_name` and `register_trait` into the
`FactoryBot::Internal` module
- Deprecates uses of `trait_by_name`, `register_trait` and `traits` from
the `FactoryBot` module.
- Rename DEPRECATOR => Deprecation

This is one of the steps towards fixing [this
issue](#1281)

### DIFF
--- a/spec/support/macros/deprecation.rb
+++ b/spec/support/macros/deprecation.rb
@@ -2,10 +2,10 @@ require "active_support"
 
 module SilenceDeprecation
   def silence_deprecation(example)
-    cached_silenced = ActiveSupport::Deprecation.silenced
-    ActiveSupport::Deprecation.silenced = true
+    cached_silenced = FactoryBot::Deprecation.silenced
+    FactoryBot::Deprecation.silenced = true
     example.run
-    ActiveSupport::Deprecation.silenced = cached_silenced
+    FactoryBot::Deprecation.silenced = cached_silenced
   end
 end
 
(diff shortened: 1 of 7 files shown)


spec/support/macros/deprecation.rb
```diff
@@ -13,6 +13,8 @@ def silence_deprecation(example)
   config.include SilenceDeprecation
 
   config.around :example, silence_deprecation: true do |example|
-    silence_deprecation(example)
+    with_silenced_output do
```
composerinteralia: I am not a fan of this `with_silenced_output` method. This is not longer just silencing the deprecation, it is silencing any output, which would make things hard to debug if anything went wrong.

Instead, I think we can just update the `silence_deprecation` to set `FactoryBot::DEPRECATOR.silenced` instead of `ActiveSupport::Deprecation.silenced`.

composerinteralia: We might also consider renaming `FactoryBot::DEPRECATOR` to `FactoryBot::Deprecation` so it better matches `ActiveSupport::Deprecation`.

Issue #1227, opened 2018-10-31
Improve error message for certain typos when defining associations.


composerinteralia:
Specifically typos that lead to `FactoryBot::Declarations` getting passed around, as described in https://github.com/thoughtbot/factory_bot/pull/1219#issuecomment-433433289

PR #1286, opened 2019-04-24
Offer advice for association typo

composerinteralia:
Closes https://github.com/thoughtbot/factory_bot/issues/1227

In this code we are passing an implicit declaration `user`, rather than
the symbol `:user`:

```rb
factory :post do
  author factory: user
end
```

In https://github.com/thoughtbot/factory_bot/issues/1227 we improved the
error message from:

```rb
undefined method 'name' for :post:Symbol
```

to:

```rb
Trait not registered: user
```

But this still doesn't make it totally obvious what the error was.
Why are we trying to register a user trait?

With this PR we update the error message to:

```rb
Association 'author' received an invalid factory argument.
Did you mean? 'factory: :user'
```

### DIFF
--- a/lib/factory_bot/declaration/association.rb
+++ b/lib/factory_bot/declaration/association.rb
@@ -22,8 +22,19 @@ module FactoryBot
       private
 
       def build
+        ensure_factory_is_not_a_declaration!
+
         factory_name = @overrides[:factory] || name
         [Attribute::Association.new(name, factory_name, [@traits, @overrides.except(:factory)].flatten)]
+      end
+
+      def ensure_factory_is_not_a_declaration!
+        if @overrides[:factory].is_a?(Declaration)
+          raise ArgumentError.new(<<~MSG)
+            Association '#{name}' received an invalid factory argument.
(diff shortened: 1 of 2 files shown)


lib/factory_bot/declaration/association.rb
```diff
@@ -22,9 +22,20 @@ def ==(other)
...
 
       def build
+        ensure_factory_is_not_a_declaration!
+
         factory_name = @overrides[:factory] || name
         [Attribute::Association.new(name, factory_name, [@traits, @overrides.except(:factory)].flatten)]
       end
+
+      def ensure_factory_is_not_a_declaration!
+        if Declaration === @overrides[:factory]
```
seanpdoyle: I don't encounter `===` very often. I'm having a hard time using a search engine to find some documentation for it.

Is there an English counterpart that has the same semantics?

Is this line equivalent to `@overrides[:factory].is_a?(Declaration)`?

composerinteralia: https://ruby-doc.org/core-2.6.3/Module.html#method-i-3D-3D-3D

`is_a?` is definitely more expressive. I will change it.

Issue #1272, opened 2019-03-06
Rails 5 - Undefined Method for all factories


nbaki:
When using version `5.0.2` of the `factory_bot_rails` gem, my factories now all give me an `Undefined Method` error for all my registered factories. If I downgrade the version of the gem to `4.11` it works.

`spec_helper.rb`
```
RSpec.configure do |config|

  config.before(:suite) do
    FactoryBot.definition_file_paths = [File.expand_path('../factories', __FILE__)]
    FactoryBot.find_definitions
  end
...
end
```

I'm using Rails 5.2.

Any help would be appreciated.

composerinteralia:
This sounds similar to https://github.com/thoughtbot/factory_bot_rails/issues/313. Any chance this is a result of static attributes being deprecated. Are you getting any deprecation warnings on 4.11?

If it is not that, can you provide a backtrace?

nbaki:
I do get the deprecation warnings on 4.11 yes

composerinteralia:
Got it. You will need to address those deprecation warnings before you can upgrade to factory_bot_rails 5. Let me know if you have any trouble with that.

composerinteralia:
I tried to improve the error message a bit in https://github.com/thoughtbot/factory_bot/pull/1287. Hopefully that will help make it more clear what is going on if somebody tries to update directly from factory_bot <= 4.10 straight to factory_bot >= 5.0 without first going to 4.11

promisepreston:
I think a more explicit answer would have greatly helped.

From the blog by Thoughbot, I came to realize that statically defined attributes have been deprecated from  factory_bot >= 5.0. And so rather than use statically defined attributes for your factories, use dynamically defined attributes

Here's an example

**Use Dynamically defined attribues**
`factory :robot do
  name { "Ralph" }
  email { "ralph@example.com" }
end`

**And not statically defined attributes**

`factory :robot do
  name "Ralph"
email  "ralph@example.com" 
end`

You can read up more about it here
[https://thoughtbot.com/blog/deprecating-static-attributes-in-factory_bot-4-11](url)

tarr11:
That link is broken somehow: https://thoughtbot.com/blog/deprecating-static-attributes-in-factory_bot-4-11

PR #1287, opened 2019-04-24
Improve error for static attribute attempts

composerinteralia:
Some people upgrade straight from factory_bot < 4.11 to factory_bot >=
5.0 and miss the deprecation cycle for static attributes. I have gotten
some feedback that the NoMethodError is confusing. Since we know that in
most cases people are seeing the NoMethodError when trying to define
static attributes, we offer them a "Did you mean?"-style suggestion for
how they might update to dynamic attributes.

I removed the extra test for setter methods. It was a remnant of
something I had removed in #1200. I see no reason to have special
treatment for setters at this point.

Closes https://github.com/thoughtbot/factory_bot/issues/1272

### DIFF
--- a/lib/factory_bot/definition_proxy.rb
+++ b/lib/factory_bot/definition_proxy.rb
@@ -89,14 +89,16 @@ module FactoryBot
     #
     # are equivalent.
     def method_missing(name, *args, &block) # rubocop:disable Style/MethodMissing
-      if args.empty?
+      association_options = args.first
+
+      if association_options.nil?
         __declare_attribute__(name, block)
-      elsif args.first.respond_to?(:has_key?) && args.first.has_key?(:factory)
-        association(name, *args)
+      elsif __valid_association_options?(association_options)
+        association(name, association_options)
       else
         raise NoMethodError.new(<<~MSG)
           undefined method '#{name}' in '#{@definition.name}' factory
-          Did you mean? '#{name} { #{args.first.inspect} }'
+          Did you mean? '#{name} { #{association_options.inspect} }'
         MSG
       end
     end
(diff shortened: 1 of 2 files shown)


lib/factory_bot/definition_proxy.rb
```diff
@@ -94,9 +94,10 @@ def method_missing(name, *args, &block) # rubocop:disable Style/MethodMissing
       elsif args.first.respond_to?(:has_key?) && args.first.has_key?(:factory)
         association(name, *args)
       else
-        raise NoMethodError.new(
-          "undefined method '#{name}' in '#{@definition.name}' factory",
-        )
+        raise NoMethodError.new(<<~MSG)
+          undefined method '#{name}' in '#{@definition.name}' factory
+          Did you mean? '#{name} { #{args.first.inspect} }'
```
seanpdoyle: I see that `args.first` is used a few times in this method. Would it be worth extracting a variable to help describe why `args.first` is significant?

What do you think about:

```rb
def method_missing(name, *args, &block)
  attribute_value = Array(args).first  # coercing with `Array()` might not be necessary with splats

  if attribute_value.nil?
    # ...
  elsif attribute_value.respond_to?(:has_key?) && attribute_value.has_key?(:factory)
    # ...
  else
    # ...
  end
end

Comment (composerinteralia) [reply lib/factory_bot/definition_proxy.rb]: I like it. I adjusted the name a little and pulled out a private predicate method. What do you think:

```rb
def method_missing(name, *args, &block)
  association_options = args.first # AFAIK args will always be an array

  if association_options.nil?
    # ...
  elsif __valid_association_options?__(association_options)
    # ...
  else
    # ...
  end
end

private

def __valid_association_options?(options)
  options.respond_to?(:has_key?) && options.has_key?(:factory)
end
```

PR #1027, opened 2017-07-24
Implement transient associations

faucct:
Fixes https://github.com/thoughtbot/factory_girl/issues/1024

Issue #1024, opened 2017-07-07
Transient attribute is assigned to model when it refers another factory


ypresto:
When there is `user` factory, below code will raise error:

```rb
FactoryGirl.define do
  factory :group do
    transient do
      user_count 0 # passes
      user # Seems trying to assign user attribute to group instance and fails.
    end

    # Use "user" to create some attribute or from after(:save), etc.
  end
end
```

```
    NoMethodError:
       undefined method `user=' for #<Group:...>
     # /usr/local/bundle/gems/factory_girl-4.8.0/lib/factory_girl/attribute_assigner.rb:16:in `public_send'
     # /usr/local/bundle/gems/factory_girl-4.8.0/lib/factory_girl/attribute_assigner.rb:16:in `block (2 levels) in object'
     # /usr/local/bundle/gems/factory_girl-4.8.0/lib/factory_girl/attribute_assigner.rb:15:in `each'
     # /usr/local/bundle/gems/factory_girl-4.8.0/lib/factory_girl/attribute_assigner.rb:15:in `block in object'
     # /usr/local/bundle/gems/factory_girl-4.8.0/lib/factory_girl/attribute_assigner.rb:14:in `tap'
     # /usr/local/bundle/gems/factory_girl-4.8.0/lib/factory_girl/attribute_assigner.rb:14:in `object'
     # /usr/local/bundle/gems/factory_girl-4.8.0/lib/factory_girl/evaluation.rb:12:in `object'
     # /usr/local/bundle/gems/factory_girl-4.8.0/lib/factory_girl/strategy/create.rb:9:in `result'
     # /usr/local/bundle/gems/factory_girl-4.8.0/lib/factory_girl/factory.rb:42:in `run'
     # /usr/local/bundle/gems/factory_girl-4.8.0/lib/factory_girl/factory_runner.rb:29:in `block in run'
     # /usr/local/bundle/gems/factory_girl-4.8.0/lib/factory_girl/factory_runner.rb:28:in `run'
     # /usr/local/bundle/gems/factory_girl-4.8.0/lib/factory_girl/strategy_syntax_method_registrar.rb:20:in `block in define_singular_strategy_method'
```

gnclmorais:
@ypresto, can you provide more code for your issue, please? For example, the model for `Group` and `User`?

composerinteralia:
I am going to close this issue now since #1290 has a clear description of the issue and most of the discussion has happened over there.

PR #1288, opened 2019-04-24
Implement transient associations

jeisen:
From https://github.com/faucct/factory_girl/tree/feature/transient_associations. I claim no credit for the work except to update the code.

I ran into the same problems (Issue #1024) as were addressed in the existing open PR https://github.com/thoughtbot/factory_bot/pull/1027, and made it compatible with current master. If there is a more supported way to do this, I'd be happy to delete this PR and do it the other way.

My use case:
I am using FactoryBot to generate nested Hash structures for use as mock responses from the AWS API. These structures sometimes fully contain other fully-defined structures within them, and sometimes simply echo data that would be well-formed values in those other structures. My strategy has been to do things like this:

```
factory :base_aws_object, :class => Hash do
  initialize_with { attributes.stringify_keys }
  skip_create

  factory :ebs_volume do
    sequence(:volume_id) { |n| "vol-#{n}" }
    size { 8 }
    ....
  end

  factory :ebs_snapshot do
    transient do
      ebs_volume
    end

    sequence(:snapshot_id) { |n| "snap-#{n}" }
    description { "Test Snapshot" }
    volume_id { ebs_volume['volume_id'] }
    volume_size { ebs_volume['size'] }
    ....
  end
end
```

Without this PR, the transient associations are still added to the final Hash when doing `build(:ebs_snapshot)`, and thus are rejected by the AWS API gem as invalid.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,4 +25,4 @@ Style/FormatStringToken:
 # Let's not open a big PR to fix all of these at once - 
 # we can fix gradually if we happen to be editing a file that has a violation
 Metrics/LineLength:
-  Max: 110
+  Max: 115
--- a/lib/factory_bot/attribute/association.rb
+++ b/lib/factory_bot/attribute/association.rb
@@ -4,8 +4,8 @@ module FactoryBot
     class Association < Attribute
       attr_reader :factory
 
-      def initialize(name, factory, overrides)
-        super(name, false)
+      def initialize(name, ignored, factory, overrides)
+        super(name, ignored)
         @factory   = factory
         @overrides = overrides
       end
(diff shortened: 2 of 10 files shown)


composerinteralia:
Thanks for taking a look at this. I will try to take a look later this week.

I appreciate the work you did do get the branch up to date, but I don't want to wipe out the authorship of the original commit. I think the right way to go is to keep the original author on the commit but then add yourself as a co-author by putting:

```
 Co-authored-by: Your Name <youremail@example.com>
```

in the commit message.

jeisen:
Replaced by #1290 to maintain original credit.

PR #1290, opened 2019-04-25
Implement transient associations

jeisen:
Co-authored-by: Jon Eisenstein <jon@zype.com>

(Revision of PR #1288 to maintain proper credit to original author)

From https://github.com/faucct/factory_girl/tree/feature/transient_associations. I claim no credit for the work except to update the code.

I ran into the same problems (Issue #1024) as were addressed in the existing open PR #1027, and made it compatible with current master. If there is a more supported way to do this, I'd be happy to delete this PR and do it the other way.

My use case:
I am using FactoryBot to generate nested Hash structures for use as mock responses from the AWS API. These structures sometimes fully contain other fully-defined structures within them, and sometimes simply echo data that would be well-formed values in those other structures. My strategy has been to do things like this:

```
factory :base_aws_object, :class => Hash do
  initialize_with { attributes.stringify_keys }
  skip_create

  factory :ebs_volume do
    sequence(:volume_id) { |n| "vol-#{n}" }
    size { 8 }
    ....
  end

  factory :ebs_snapshot do
    transient do
      ebs_volume
    end

    sequence(:snapshot_id) { |n| "snap-#{n}" }
    description { "Test Snapshot" }
    volume_id { ebs_volume['volume_id'] }
    volume_size { ebs_volume['size'] }
    ....
  end
end
```

Without this PR, the transient associations are still added to the final Hash when doing build(:ebs_snapshot), and thus are rejected by the AWS API gem as invalid.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,4 +25,4 @@ Style/FormatStringToken:
 # Let's not open a big PR to fix all of these at once - 
 # we can fix gradually if we happen to be editing a file that has a violation
 Metrics/LineLength:
-  Max: 110
+  Max: 115
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -25,4 +25,4 @@ Style/FormatStringToken:
 # Let's not open a big PR to fix all of these at once - 
 # we can fix gradually if we happen to be editing a file that has a violation
 Metrics/LineLength:
-  Max: 115
+  Max: 110
--- a/spec/acceptance/transient_attributes_spec.rb
+++ b/spec/acceptance/transient_attributes_spec.rb
@@ -1,6 +1,6 @@
 describe "transient attributes" do
   before do
-    define_model('Country')
+    define_model("Country")
     define_model("User", name: :string, email: :string)
 
     FactoryBot.define do
--- a/spec/acceptance/transient_attributes_spec.rb
+++ b/spec/acceptance/transient_attributes_spec.rb
@@ -1,15 +1,19 @@
 describe "transient attributes" do
   before do
+    define_model('Country')
     define_model("User", name: :string, email: :string)
 
     FactoryBot.define do
       sequence(:name) { |n| "John #{n}" }
+
+      factory :country
 
       factory :user do
         transient do
           four     { 2 + 2 }
           rockstar { true }
           upcased  { false }
+          country
         end
 
         name  { "#{FactoryBot.generate(:name)}#{' - Rockstar' if rockstar}" }
--- a/spec/acceptance/transient_attributes_spec.rb
+++ b/spec/acceptance/transient_attributes_spec.rb
@@ -1,12 +1,19 @@
 describe "transient attributes" do
   before do
-    define_model("Country")
-    define_model("User", name: :string, email: :string)
+    define_model("Country", country_code: :string)
+    # define_model("Account", account_number: :string)
+    define_model("User", name: :string, email: :string, country_code: :string)
 
     FactoryBot.define do
       sequence(:name) { |n| "John #{n}" }
 
-      factory :country
+      factory :country do
+        country_code { 'US' }
+      end
+
+      # factory :account do
+      #   sequence(:account_number)
+      # end
 
       factory :user do
         transient do
@@ -18,6 +25,7 @@ describe "transient attributes" do
 
         name  { "#{FactoryBot.generate(:name)}#{' - Rockstar' if rockstar}" }
         email { "#{name.downcase}#{four}@example.com" }
+        country_code { country.country_code }
 
         after(:create) do |user, evaluator|
           user.name.upcase! if evaluator.upcased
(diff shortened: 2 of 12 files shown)


.rubocop.yml
```diff
@@ -25,4 +25,4 @@ Style/FormatStringToken:
 # Let's not open a big PR to fix all of these at once - 
 # we can fix gradually if we happen to be editing a file that has a violation
 Metrics/LineLength:
-  Max: 110
+  Max: 115
```
composerinteralia: I would rather not change this. Can we restructure those lines so they aren't as long?

spec/acceptance/transient_attributes_spec.rb
```diff
@@ -1,15 +1,19 @@
...
       sequence(:name) { |n| "John #{n}" }
 
+      factory :country
+
       factory :user do
         transient do
           four     { 2 + 2 }
           rockstar { true }
           upcased  { false }
+          country
```
composerinteralia: This implicitly tests that the transient association doesn't try to set `user.country=`, but I would like to see acceptance tests demonstrating that this really works.

The test cases should cover both explicit associations

```rb
transient do
  association :country
end
```

and implicit ones

```rb
transient do
  country
end
```

and then maybe

jeisen: I'll take a look at this.

jeisen: @composerinteralia: I'm constructing an acceptance test where the transient association is used to construct values in the non-transient attributes, but I'm running into some trouble with the `attributes_for` strategy. It looks like the behavior is that it will always initialize the transient attributes as nil when running this strategy, which causes the dependent attributes to fail:

```
  before do
    define_model("Country", country_code: :string)
    define_model("User", name: :string, email: :string, country_code: :string)

    FactoryBot.define do
      sequence(:name) { |n| "John #{n}" }

      factory :country do
        country_code { 'US' }
      end

      factory :user do
        transient do
          four     { 2 + 2 }
          rockstar { true }
          upcased  { false }
          country
        end

        name  { "#{FactoryBot.generate(:name)}#{' - Rockstar' if rockstar}" }
        email { "#{name.downcase}#{four}@example.com" }
        country_code { country.country_code }

        after(:create) do |user, evaluator|
          user.name.upcase! if evaluator.upcased
        end
      end
    end
  end

  context "returning attributes for a factory" do
    subject { FactoryBot.attributes_for(:user, rockstar: true) }

    it { should_not have_key(:four) }
    it { should_not have_key(:rockstar) }
    it { should_not have_key(:upcased) }
    it { should_not have_key(:country) }
    it { should     have_key(:name) }
    it { should     have_key(:email) }
    it { should     have_key(:country_code) }
  end
end
```

I can get around this issue by passing in `country: Country.new` to the `attributes_for` method, but that's ignoring the implicit association defined in the factory. Is this expected behavior, or a bug?

jeisen:
```
  1) transient attributes returning attributes for a factory 
     Failure/Error: country_code { country.country_code }
     
     NoMethodError:
       undefined method `country_code' for nil:NilClass
```

composerinteralia: I might not have time to look at this in detail right away, but it reminds me of https://github.com/thoughtbot/factory_bot/issues/1292

composerinteralia: I finally got around to looking at this, and it is as I suspected related to #1292. It is pretty well baked in that `attributes_for` will not build any associations.

As I think about it, I don't know what the expected behavior is here. Would we expect when using `attributes_for` that the transient associations are also built with the `attributes_for` strategy? If so, that would also break your test case. Or would we expect the transient associations to be built all the time with the `build` strategy? That would pass the test, but would be a deviation from anything else in factory_bot.

This may extend to other strategies as well: if we are using the `create` strategy, should transient associations also be built with `create`? But then we might be writing records to the database that don't need to be there. Could we use `build` for the transient associations all the time? But what then if we are trying to use the id of the transient record?

These unanswered questions lead me to wonder whether we should proceed with this feature at all. I haven't yet seen a use case for it (https://github.com/thoughtbot/factory_bot/issues/1024 mentions that this doesn't work, but doesn't give an example for why we would want this to work). If this isn't something that solves a problem people are having, then it may not be worth the added complexity.

I don't really have any answers here. I am thinking out loud.

composerinteralia:
Since there are still a number of open questions and there hasn't been any activity here for a while I am going to close this issue.

For the example given in the description, I think you can workaround this by changing:

```rb
    transient do
      ebs_volume
    end
```

to:

```rb
    transient do
      ebs_volume { build(:ebs_volume) }
    end
```

I added some documentation about this in https://github.com/thoughtbot/factory_bot/commit/8ceea2f885bbc776d7b5c759dbd6e1e69037b0e6.

PR #1291, opened 2019-04-26
Deprecate and move to Internal sequence methods

aledustet:
Why:
Another run of internal methods that should not be publicly
available from the base namespace.
This time the sequence involving methods were moved. It's worth noticing
that the ```Internal``` class is getting crowded. Maybe we can start
name-spacing the internal groups into modules under
```internal/sequence.rb``` ```internal/trait.rb``` and so on. Thoughts?

This PR:

- Moves the ```register_sequence```, ```sequence_by_name```,
```sequences``` and ```rewind_sequences``` to the
```FactoryBot::Internal``` module.
- Deprecates uses of ```register_sequence```, ```sequence_by_name```,
```sequences``` and ```rewind_sequences``` from the ```FactoryBot``` module.

This is one of the steps towards fixing [this issue](https://github.com/thoughtbot/factory_bot/pull/1285#1281)

### DIFF
--- a/lib/factory_bot.rb
+++ b/lib/factory_bot.rb
@@ -87,8 +87,11 @@ module FactoryBot
              :constructor,
              to: :configuration
 
-    delegate :trait_by_name,
+    delegate :register_sequence,
              :register_trait,
+             :rewind_sequences,
+             :sequence_by_name,
+             :trait_by_name,
              to: Internal
 
     attr_accessor :allow_class_lookup
@@ -96,6 +99,8 @@ module FactoryBot
     deprecate :allow_class_lookup,
               :allow_class_lookup=,
               :register_trait,
+              :sequence_by_name,
+              :sequences,
               :trait_by_name,
               :traits,
               deprecator: Deprecation
@@ -110,22 +115,6 @@ module FactoryBot
 
   def self.factory_by_name(name)
     factories.find(name)
-  end
-
-  def self.register_sequence(sequence)
-    sequence.names.each do |name|
-      sequences.register(name, sequence)
-    end
-    sequence
-  end
-
-  def self.sequence_by_name(name)
-    sequences.find(name)
-  end
-
-  def self.rewind_sequences
-    sequences.each(&:rewind)
-    Internal.rewind_inline_sequences
   end
 
   def self.register_strategy(strategy_name, strategy_class)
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: lib/factory_bot.rb [hunk 1]
```diff
@@ -87,15 +87,22 @@ class << self
...
+             :sequence_by_name,
+             :trait_by_name,
              to: Internal
 
     attr_accessor :allow_class_lookup
 
     deprecate :allow_class_lookup,
               :allow_class_lookup=,
               :register_trait,
+              :register_sequence,
```
composerinteralia: `register_sequence` is part of the public api (see the last section of [the sequence documentation](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#sequences)), so we shouldn't deprecate it.

spec/acceptance/sequence_resetting_spec.rb
```diff
@@ -1,4 +1,4 @@
-describe "FactoryBot.rewind_sequences" do
+describe "FactoryBot::Internal.rewind_sequences" do
```
composerinteralia: This should remain `FactoryBot.rewind_sequences` since we want to test the public method, not the internal one.

spec/factory_bot/internal_spec.rb
```diff
@@ -17,4 +17,35 @@
...
+    end
+  end
+
+  describe ".rewind_sequences" do
+    it "rewinds the sequences and the internal sequences" do
+      sequence = FactoryBot::Sequence.new(:email)
+      inline_sequence = FactoryBot::Sequence.new(:email)
+      FactoryBot::Internal.register_sequence(sequence)
+      FactoryBot::Internal.register_inline_sequence(inline_sequence)
+      expect(sequence).to receive(:rewind).exactly(:once)
```
composerinteralia: This puts the verify step of the [four-phase test](https://thoughtbot.com/blog/four-phase-test) before the exercise step. Instead, we can use [spies](https://relishapp.com/rspec/rspec-mocks/v/3-8/docs/basics/spies) for the sequences and then `expect(sequence).to have_received` the method

Also, our guides suggest separating those steps with newlines: https://github.com/thoughtbot/guides/tree/master/style/testing#testing

PR #1293, opened 2019-05-03
Deprecate and move to Internal factories methods

aledustet:
Why:
These methods are used internally for the functionality of the library
and are subject to change. Therefore shouldn't be part of the public
interface.

This PR:
- Moves the ```register_factory``` and ```factory_by_name``` methods to
the ```FactoryBot::Internal``` class.
- Deprecates the use of ```register_factory``` and ```factory_by_name```
from the ```FactoryBot``` module.


.gitignore
```diff
@@ -1,3 +1,4 @@
+*.DS_Store
```
composerinteralia: I have always felt that OS and editor specific belong in each person's own global gitignore (checkout the [thoughtbot dotfiles global gitignore](https://github.com/thoughtbot/dotfiles/blob/master/gitignore), for example) rather than in the project's gitignore.

That said, this file already has `*.swp` in it, which is a vim thing.

Totally your call on whether you want to include it.

so I guess I don't feel that strongly one way or another

lib/factory_bot.rb
```diff
@@ -75,19 +75,21 @@ def self.lint(*args)
   end
 
   class << self
-    delegate :factories,
+    delegate :callbacks,
```
composerinteralia: 👍 🔤

lib/factory_bot.rb
```diff
@@ -98,6 +100,8 @@ class << self
 
     deprecate :allow_class_lookup,
```
composerinteralia: Are we also able to deprecate `FactoryBot.factories`?

lib/factory_bot/null_object.rb
```diff
@@ -6,11 +6,8 @@ def initialize(methods_to_respond_to)
     end
 
     def method_missing(name, *args, &block)
-      if respond_to?(name)
```
composerinteralia: This seems unrelated to this PR. Also, I think I like the first version a little better. It is easier to see the conditional.

spec/factory_bot/internal_spec.rb
```diff
@@ -8,6 +8,11 @@
         from(0).
         to(1)
     end
+
+    it "returns the registered trait" do
```
composerinteralia: Is it important that it returns the registered trait? Would it matter if we eventually return something else? If not, we might not need this test.

spec/factory_bot/internal_spec.rb
```diff
@@ -27,6 +32,11 @@
         from(0).
         to(1)
     end
+
+    it "returns the registered sequence" do
```
composerinteralia: Same question here

spec/factory_bot/internal_spec.rb
```diff
@@ -48,4 +58,28 @@
       FactoryBot::Internal.rewind_sequences
     end
   end
+
+  describe ".register_factory" do
```
composerinteralia: As mentioned in https://github.com/thoughtbot/factory_bot/pull/1291#discussion_r283066959, our style guide (and also I 😄 ) would prefer some newlines in these tests.

spec/factory_bot/internal_spec.rb
```diff
@@ -48,4 +58,28 @@
...
+    it "registers the provided factory" do
+      factory = FactoryBot::Factory.new(:object)
+      configuration = FactoryBot::Internal.configuration
+      expect { FactoryBot::Internal.register_factory(factory) }.
+        to change { configuration.factories.count }.
+        from(0).
+        to(1)
+    end
+
+    it "returns the registered factory" do
```
composerinteralia: And here

.gitignore
aledustet: I was wondering why it wasn't there. I have it in my global config but I was editing it home and it slipped I agree that specific settings should belong in your own environment. I'll remove it 👍

spec/factory_bot/internal_spec.rb
aledustet: ```register_trait``` and ```register_sequence``` both have the same signature in the method, I thought it was necessary, that's why I added the test, I only see it used in ```FactoryBot::Syntax::Default::DSL``` and the result of using the block in the DSL is not documented(to the best of my knowledge) to use the return value. Maybe we are safe removing the return statement? There is a test under ```attributes/dynamic_spec``` that uses the return value.

lib/factory_bot.rb
aledustet: There is one mention of using this method in the [docs](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#linting-factories), maybe the use case it's not needed?

Issue #1238, opened 2018-11-26
Fix "Self-referencing association" error for traits


composerinteralia:
This originally came up in https://github.com/thoughtbot/factory_bot/pull/1017, which includes a failing spec.

### Self-referencing association

The "Self-referencing association" error was introduced in e4a3d25711ffcc7945a49bdf67bdf564d1cb27e4 to catch this:

```rb
factory :user do
  user
end
```

This is still working as expected

### Trait referencing association with the same name

When the error was moved from `Factory` into `AttributeList` in e3bf491dd6af2b6c5c110c4a795db65bb458e3da it started catching associations with the same name as a trait as well:

```rb
factory :post

factory :user do
  trait :post do
    association :post
  end
end
```

The `AttributeList` doesn't know when it is part of a trait definition rather than a factory definition.
It just raises an error if its name matches the name of the association:
https://github.com/thoughtbot/factory_bot/blob/cffa56e4c8c55bb1467e14e6fb885a6b6fafc54f/lib/factory_bot/attribute_list.rb#L56

While I think naming the factory and trait the same thing is confusing, I am not convinced it should raise an error.

### Self-referencing trait

What should definitely raise an error is a self-referential trait, but this will currently give us a trait that calls itself recursively:

```rb
factory :user do
  trait :bad_trait do
    bad_trait
  end
end
```

I think it would make sense to catch this and 

```rb
raise TraitDefinitionError, "Self-referencing trait..."
```

swetagsanghavi:
Just merged https://github.com/thoughtbot/factory_bot/pull/1294 which addresses this issue.

PR #1294, opened 2019-05-03
Raise error when a trait is self-referencing

swetagsanghavi:
We want to check if any attributes of a new trait definition matches the name of
the trait. Raise an error if so, to avoid a recursive call to the trait. A fix to:
https://github.com/thoughtbot/factory_bot/issues/1238

### DIFF
--- a/lib/factory_bot/trait.rb
+++ b/lib/factory_bot/trait.rb
@@ -7,9 +7,11 @@ module FactoryBot
       @name = name.to_s
       @block = block
       @definition = Definition.new(@name)
-
       proxy = FactoryBot::DefinitionProxy.new(@definition)
-      proxy.instance_eval(&@block) if block_given?
+
+      if block_given?
+        proxy.instance_eval(&@block)
+      end
     end
 
     delegate :add_callback, :declare_attribute, :to_create, :define_trait, :constructor,
--- a/lib/factory_bot/declaration/implicit.rb
+++ b/lib/factory_bot/declaration/implicit.rb
@@ -25,6 +25,9 @@ module FactoryBot
           [Attribute::Association.new(name, name, {})]
         elsif FactoryBot::Internal.sequences.registered?(name)
           [Attribute::Sequence.new(name, name, @ignored)]
+        elsif @factory.name.to_s == name.to_s
+          message = "Self-referencing trait '#{@name}'"
+          raise TraitDefinitionError, message
         else
           @factory.inherit_traits([name])
           []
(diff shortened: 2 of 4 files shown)


lib/factory_bot/trait.rb
```diff
@@ -26,6 +29,13 @@ def ==(other)
 
     protected
 
+    def ensure_trait_not_self_referencing(evaulated_block)
+      if evaulated_block.respond_to?(:name) && evaulated_block.name.to_s == @name
```
swetagsanghavi: I've worked myself into the wild meta-ness of factory bot. I'm not sure what this should be and where this method should be.

lib/factory_bot/trait.rb
```diff
@@ -7,9 +7,12 @@ def initialize(name, &block)
       @name = name.to_s
       @block = block
       @definition = Definition.new(@name)
-
       proxy = FactoryBot::DefinitionProxy.new(@definition)
```
swetagsanghavi: Should the two lines above only happen `if block_given?`?

lib/factory_bot/trait.rb
```diff
@@ -26,6 +29,13 @@ def ==(other)
 
     protected
 
+    def ensure_trait_not_self_referencing(evaulated_block)
```
aledustet: Maybe this is ```evaluated_block```

lib/factory_bot/trait.rb
```diff
@@ -7,9 +7,12 @@ def initialize(name, &block)
       @name = name.to_s
       @block = block
       @definition = Definition.new(@name)
-
       proxy = FactoryBot::DefinitionProxy.new(@definition)
-      proxy.instance_eval(&@block) if block_given?
+
+      if block_given?
+        evaulated_block = proxy.instance_eval(&@block)
+        ensure_trait_not_self_referencing(evaulated_block)
```
aledustet: It seems exactly what it needs, I may be completely out of sync here, but this seems possible:
```ruby
  factory :user do
    trait :admin do
      admin {true}
    end
end
```
And shouldn't that syntax be valid?

lib/factory_bot/trait.rb
aledustet: maybe this should be ```evaluated_block```

swetagsanghavi: hahaha 😬 thank you!

lib/factory_bot/trait.rb
swetagsanghavi: It seems from the issue https://github.com/thoughtbot/factory_bot/issues/1238 that when you have a self-referencing trait, it will call itself recursively.

lib/factory_bot/trait.rb
swetagsanghavi: I think not

PR #1297, opened 2019-05-07
Deprecate/Move strategies and callback methods

aledustet:
Why:
These methods are used internally for the functionality of the library
and are subject to change. Therefore shouldn't be part of the public
interface.

This PR:
- Moves the ```register_strategy```, ```register_callback```,
```register_default_factories```, ```register_default_callbacks```
and ```strategy_by_name``` methods to the ```FactoryBot::Internal```
class.
- Deprecates the use of ```register_callback``` from the ```FactoryBot```
module.

### DIFF
--- a/lib/factory_bot.rb
+++ b/lib/factory_bot.rb
@@ -88,11 +88,16 @@ module FactoryBot
              to: :configuration
 
     delegate :factory_by_name,
+             :register_callback,
+             :register_default_callbacks,
+             :register_default_strategies,
              :register_factory,
              :register_sequence,
+             :register_strategy,
              :register_trait,
              :rewind_sequences,
              :sequence_by_name,
+             :strategy_by_name,
              :trait_by_name,
              to: Internal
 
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: lib/factory_bot/internal.rb [hunk 1]
```diff
@@ -57,6 +75,30 @@ def register_factory(factory)
...
+
+      def strategy_by_name(name)
+        strategies.find(name)
+      end
+
+      def register_default_strategies
+        DEFAULT_STRATEGIES.each do |strategy_name, strategy_class|
+          register_strategy(strategy_name, strategy_class)
+        end
+      end
```
swetagsanghavi: Nice refactor!

swetagsanghavi:
Should we link https://github.com/thoughtbot/factory_bot/issues/1281 to this PR?

lib/factory_bot/internal.rb
```diff
@@ -57,6 +75,30 @@ def register_factory(factory)
...
+        strategies.register(strategy_name, strategy_class)
+        StrategySyntaxMethodRegistrar.new(strategy_name).define_strategy_methods
+      end
+
+      def strategy_by_name(name)
+        strategies.find(name)
+      end
+
+      def register_default_strategies
+        DEFAULT_STRATEGIES.each do |strategy_name, strategy_class|
```
composerinteralia: Is it possible to do `DEFAULT_STRATEGIES.each(&method(:register_strategy))` like you did with the callbacks?

lib/factory_bot/syntax/methods.rb
```diff
@@ -30,7 +30,7 @@ module Syntax
     ##   # factory with traits and attribute override
     ##   build_stubbed_list(:user, 15, :admin, :male, name: "John Doe")
     module Methods
-      # @!parse FactoryBot.register_default_strategies
+      # @!parse FactoryBot::Internal.register_default_strategies
```
composerinteralia: My yard doc knowledge is limited. What does `@!parse` do?

spec/acceptance/callbacks_spec.rb
```diff
@@ -138,9 +138,9 @@ def name
       end
     end
 
-    FactoryBot.register_strategy(:custom_before, custom_before)
-    FactoryBot.register_strategy(:custom_after, custom_after)
-    FactoryBot.register_strategy(:totally_custom, totally_custom)
+    FactoryBot::Internal.register_strategy(:custom_before, custom_before)
```
composerinteralia: `FactoryBot.register_strategy` is still around as a non-internal method, so I think we can leave a lot of these specs as they were.

spec/factory_bot_spec.rb
```diff
@@ -18,6 +18,12 @@
     expect(FactoryBot.trait_by_name(trait.name)).to eq trait
   end
 
+  it "finds a registered strategy" do
```
composerinteralia: Should this be `FactoryBot` instead of `FactoryBot::Internal`?

lib/factory_bot.rb
```diff
@@ -109,36 +113,7 @@ class << self
...
-  def self.register_strategy(strategy_name, strategy_class)
-    strategies.register(strategy_name, strategy_class)
-    StrategySyntaxMethodRegistrar.new(strategy_name).define_strategy_methods
-  end
-
-  def self.strategy_by_name(name)
-    strategies.find(name)
-  end
-
-  def self.register_default_strategies
```
composerinteralia: We should delegate and deprecate `register_default_strategies` and `register_default_callbacks` too

lib/factory_bot/syntax/methods.rb
aledustet: It says that it parses the code as if it were present in that location. I think this was added since that happens explicitly to have it as part of the declaration of the method and to have yard doc add it as part of the class.

lib/factory_bot.rb
```diff
@@ -101,6 +104,7 @@ class << self
     deprecate :allow_class_lookup,
               :allow_class_lookup=,
               :factory_by_name,
+              :register_callback,
```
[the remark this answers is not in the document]
aledustet: Calling ```register_strategy``` or ```register_callback``` interacts directly with it, and we get a lot of deprecation warnings. Maybe we can but I think it's better to do it in a separate PR, because we need to change the strategy delegating those.

PR #1299, opened 2019-05-13
Upgrade to rubocop 0.68

natesholland:
This pushes rubocop to the latest version. They have since extracted out
the performance cops into the rubocop-performance gem. Those cops were
originally used so I pulled the gem in and required it.

The only changes here are the automatic cop changes (which are in their own commit in 
29cc687) and then to fix the cops for method missing based on the changes in this PR: rubocop-hq/rubocop#5811

Changelog can be found here: https://github.com/rubocop-hq/rubocop/blob/master/CHANGELOG.md

PS. Tell @frasermince hi if he's around.


natesholland:
One question I have for the maintainers is: Should I update the gemfiles in `gemfiles` and if so, what is the best approach to go about that?

composerinteralia:
Thanks for the PR!

If I recall correctly, we have been stuck on rubocop 0.54 because our organization-wide hound config is https://raw.githubusercontent.com/thoughtbot/guides/master/style/ruby/.rubocop.yml, which uses an older syntax. I opened https://github.com/thoughtbot/guides/issues/529, but nobody has worked on it yet. I suspect we may need to update that before we can merge this PR.

natesholland:
@composerinteralia thanks for the feedback. I'll see if I can find time to look into that.

composerinteralia:
It looks like we made some updates to our shared rubocop.yml a while back, so I am going to merge this and see what happens.

composerinteralia:
Rebased and merged from the command line so Github didn't catch that this PR is merged. See https://github.com/thoughtbot/factory_bot/commits?author=natesholland. Closing.

natesholland:
Thanks for resurrecting this 👍

Issue #1268, opened 2019-02-22
Document Evaluator#association


composerinteralia:
https://github.com/thoughtbot/factory_bot/issues/1063
https://github.com/thoughtbot/factory_bot/issues/1255

The tests in https://github.com/thoughtbot/factory_bot/pull/1419 may serve as a starting point for some code examples as part of this documentation.

hmayer00:
It's worth noting one oddity with this usage. If you have an association whose factory has a different name, in general one does
```ruby
association :association_name, factory: :factory_name
```
However, if you want to use the evaluator syntax, it appears that you need to do:
```ruby
association_name { association(:factory_name) }
```

That still works fine (and I'm stoked to have it!), but I would have expected this to work, and it doesn't for me:
```ruby
association_name { association(:association_name, factory: :factory_name) }
```

Probably fine just to document that, but I figured I'd throw that out there since it's not obvious, and would be easy to miss while documenting.

composerinteralia:
Good find. I wonder if we can fix that. I won't have time to look into this for the couple weeks, but the method in question is [here](https://github.com/thoughtbot/factory_bot/blob/99ac02400fd56bd1872fc3ed84f81ea5e8f27737/lib/factory_bot/evaluator.rb#L24-L34) if you or anybody else wants to explore that.

hmayer00:
I'll take a swing at it in the next few days and see what I can come up
with.

hmayer00:
@composerinteralia no rush but since I never explicitly commented on it here I figured I should mention that I opened PR #1300 to resolve the oddity I noted above. I'm happy to also write the docs requested in this issue, pending your thoughts on the suggestion in that PR.

composerinteralia:
Thanks! I am hoping to have some time next week to take a look.

hmayer00:
No problem, thank y'all for the gems! And really no rush, none of this is urgent - just wanted to make sure it didn't get lost in what I assume are a large number of notifications.

composerinteralia:
Now documented in:

* https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#inline-definition
* https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#has_many-associations
* https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#has_and_belongs_to_many-associations
[...]

PR #1300, opened 2019-05-18
Change the evaluator association syntax to allow :factory option.

hmayer00:
This addresses the discussion of #1268 (though it does not create documentation).

This supports the syntax I had initially expected, with the :factory option:
```ruby
association_name { association(:association_name, factory: :factory_name) }
```

It continues to support the original syntax:
```ruby
association_name { association(:factory_name) }
```

It also adds specs for both usages.

This seems like the least impactful change that still works the way I had expected it to. I have some alternative thoughts which I'll address in comments.

### DIFF
--- a/lib/factory_bot/evaluator.rb
+++ b/lib/factory_bot/evaluator.rb
@@ -21,13 +21,15 @@ module FactoryBot
       end
     end
 
-    def association(factory_name, *traits_and_overrides)
+    # If `factory` option is not provided, `association_name` will be used
+    def association(association_name, *traits_and_overrides)
       overrides = traits_and_overrides.extract_options!
       strategy_override = overrides.fetch(:strategy) do
         FactoryBot.use_parent_strategy ? @build_strategy.class : :create
       end
+      factory_name = overrides[:factory] || association_name
 
-      traits_and_overrides += [overrides.except(:strategy)]
+      traits_and_overrides += [overrides.except(:strategy, :factory)]
 
       runner = FactoryRunner.new(factory_name, strategy_override, traits_and_overrides)
       @build_strategy.association(runner)
--- a/spec/acceptance/associations_spec.rb
+++ b/spec/acceptance/associations_spec.rb
@@ -16,4 +16,86 @@ describe "associations" do
       )
     end
   end
+
+  context "when building an association through the evaluator" do
+    before do
+      define_model("User", name: :string)
+
+      define_model("Thumbnail", user_id: :integer) do
+        belongs_to :user
+      end
+
+      FactoryBot.define do
+        factory :user do
+          name { "John Doe" }
+        end
+
+        factory :thumbnail do
+          user
+        end
+      end
+    end
+
+    context "when the association and factory name match" do
+      before do
+        define_model("Post", user_id: :integer, thumbnail_id: :integer) do
+          belongs_to :user
+          belongs_to :thumbnail
+        end
+
+        FactoryBot.define do
+          factory :post do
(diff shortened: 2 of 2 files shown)


hmayer00:
So I still don't love the original syntax:
```ruby
association_name { association(:factory_name) }
```

It seems like outside of the evaluator context, `association(:factory_name)` wouldn't work, and also the name of the method suggests that what should follow is the association name. So I changed the method signature such that the argument is named association_name, but still used that if no factory was provided. This feature wasn't documented anywhere, so dropping that support might be legit, but from conversations in other issues it seems likely that people are using it, so probably makes sense to leave it as is since it's not doing any harm.

However, regardless of whether the `Evaluator#association` method continues to support that, I think it would be nice to explicitly support that shorthand using its own explicit method. I'm thinking something like:

```ruby
class Evaluator
  def manufacture(factory_name, *traits_and_overrides)
    overrides = traits_and_overrides.extract_options!
    strategy_override = overrides.fetch(:strategy) do
      FactoryBot.use_parent_strategy ? @build_strategy.class : :create
    end

    traits_and_overrides += [overrides.except(:strategy)]

    runner = FactoryRunner.new(factory_name, strategy_override, traits_and_overrides)
    @build_strategy.association(runner)
  end

  def association(association_name, *traits_and_overrides)
    overrides = traits_and_overrides.extract_options!
    factory_name = overrides[:factory] || association_name
    traits_and_overrides += [overrides.except(:factory)]

    manufacture(factory_name, traits_and_overrides)
  end
end
```

This would allow the association syntax that I think makes sense:
```ruby
association_name { association(:association_name, factory: :factory_name) }
```
But also a shorthand version without the (unused, but consistent) repetition of :association_name
```ruby
association_name { manufacture(:factory_name) }
```

It would also continue to implicitly support the current syntax, in case it's in use:
```ruby
association_name { association(:factory_name) }
```

Thoughts? I'm happy to set up a different PR if you like that direction.

hmayer00:
Also, since I'm now fairly well versed in this I'm happy handle documentation as well. I just wanted to settle on these details before I wrote anything up.

composerinteralia:
@hmayer00 thanks for working on this, and sorry it has taken me so long to respond. I have looked at this PR every so often and put off commenting because I was unsure what to do. I am still unsure what to do, but maybe we can have a productive conversation about that fact.

I agree that `association_name { association(:factory_name) }` is potentially confusing because it is inconsistent with the documented association method. But `association_name { association(:unused_argument, factory: :factory_name) }` is awkward to me as well, because we end up throwing away the first argument.

Is there an ideal API here? Would it be `association_name { association(:factory_name) }`? If we have clear ideas here, I would be open to doing something like: introduce an `__association__` method that does whatever we need it to internally, deprecate the existing `association` method, introduce a new public method with our ideal public API.

Let me know if it would be easier to chat about this by phone or video (and then summarize what we come up with back in this issue), or if you would like to pair program on this.

Many thanks for all your work on factory_bot!

jopotts:
I have to say I quite like the `manufacture` suggestion. It's a nice progression and distinctive from the original `association` keyword first syntax.

```ruby
association_name { manufacture(:factory_name) }
```

composerinteralia:
I'm going to close this for now. I would prefer to introduce a new method that has the ideal API and then deprecate the existing method, rather than tack on extra not-quite-ideal behavior to a method that isn't quite right.

Thank you for your work on this! I have documentation for the `Evaluation#association` method in the works, and am hoping to get that out in the next week or so. I will document the method as-is for now.

Issue #1049, opened 2017-10-16
Feature question: Boilerplate for enum traits


Epigene:
I work with Rails and encounter models with enum fields.

```rb
class Task
  enum status: {queued: 0, started: 1, finished: 2}
end
```

Then in factory I define traits for these statuses in a programmatic way
```rb
FactoryGirl.define do
  factory :task, class: 'Task' do
    Task.statuses.each do |k, _|
      trait(k) { status: k }
    end
  end
end
```

This allows me to do `build_stubbed(:task, :finished)`.

I want to simplify this process to be able to write

```rb
FactoryGirl.define do
  factory :task, class: 'Task' do
    define_enum_traits_for(:status)    
  end
end
```

How would I go about defining a `define_enum_traits_for` method that is accessible in `FactoryGirl.define` block in all factory files of a project?

I had a look at https://github.com/thoughtbot/factory_girl/issues/564 but that does not seem quite what I'm after.

composerinteralia:
@lancejjohnson and I took a first pass at this in https://github.com/thoughtbot/factory_bot/commit/03e9a87773465e20f2276bbf9677a0c5d1e63035, and I think it is definitely doable. @seanpdoyle and I had also discussed maybe defining these automatically by looking at the object's [`defined_enums`](https://github.com/rails/rails/blob/00f7dbab7590c6ffd188da186067eb269e11580f/activerecord/lib/active_record/enum.rb#L100).

connorshea:
This'd be great!

eLod:
it would be nice to be able to use it without the `enum` field type, e.g. with a provided set of values (so it would work with `:string` or even other column types): `traits_for_enum :status, %w[enabled locked whatever]`

connorshea:
I just ran into this problem again, I think?

So I have a `GamePurchase` model with an enum that can be used to show whether you've completed a game, like this:

```ruby
enum completion_status: {
  unplayed: 0,
  in_progress: 1,
  dropped: 2,
  completed: 3,
  fully_completed: 4,
  not_applicable: 5
}
```

And then I made a factory with a trait like so, to test randomized values when running the test suite:

```ruby
# spec/factories/game_purchases.rb
trait :completion_status do
  completion_status { rand(0..5) }
end
```

The problem is that, when I went to add a new option to the enum I forgot to update this trait, so the new option wasn't ever being used in the test suite.

[...]

PR #1301, opened 2019-05-20
Enum traits

eLod:
See #1049.


lib/factory_bot/definition.rb
```diff
@@ -57,6 +57,13 @@ def compile
       end
     end
 
+    def register_defined_enums(klass)
+      return unless klass.respond_to?(:defined_enums)
+      klass.defined_enums.each_key do |name|
+        register_enum(Enum.new(name))
+      end
+    end
+
```
eLod: not sure if this is the right place to add this

spec/acceptance/enum_traits_spec.rb
```diff
@@ -3,6 +3,7 @@
     define_model("Task", status: :integer) do
       enum status: { queued: 0, started: 1, finished: 2 }
     end
+    Task.reset_column_information
```
eLod: i had to add these, otherwise AR cached the column information and the tests fail (depending on the test execution order the integers are converted to strings or vice versa)

spec/acceptance/enum_traits_spec.rb
```diff
@@ -16,4 +17,59 @@
...
+    FactoryBot.define do
+      factory :task
+    end
+
+    Task.statuses.each_key do |trait_name|
+      task = FactoryBot.build(:task, trait_name)
+
+      expect(task.status).to eq(trait_name)
+    end
+  end
```
eLod: this test makes the first one somewhat redundant

composerinteralia:
Thanks so much for this work! This all looks really good. @fridaland and I are trying to put the finishing touches on this over in #1380. We will be sure to include you as a co-author of the final commit to ensure you get credit for your work here.

eLod:
thanks, sorry for the long reply, busy days. let me know if you need anything else from me, or i can help with anything.

composerinteralia:
@eLod https://github.com/thoughtbot/factory_bot/pull/1380 could use a final review, if you are interested.

PR #1303, opened 2019-05-23
[factory_girl => factory_bot] Change grep command for OSX in the migration guide

kwbtdisk:
I just changed `grep` command for OSX
- from  `grep -e FactoryGirl **/*.rake **/*.rb -s -l` 
- to `grep -lr FactoryGirl --include="*.rake" --include="*.rb" ./*`

Just for applying replacements to the files under the nested dirs

### DIFF
--- a/UPGRADE_FROM_FACTORY_GIRL.md
+++ b/UPGRADE_FROM_FACTORY_GIRL.md
@@ -35,7 +35,7 @@ to replace all references with the new constant should do the trick. For
 example, on macOS:
 
 ```sh
-grep -e FactoryGirl **/*.rake **/*.rb -s -l | xargs sed -i "" "s|FactoryGirl|FactoryBot|g"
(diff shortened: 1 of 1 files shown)


composerinteralia:
Thanks for the PR. I am going to close this because we already have this note in the upgrade guide:

> If these examples don't work for you, various other approaches
have been suggested in pull requests #1070, #1075, #1084, #1095, and #1102.

At least one of those PR's already mentions something like your solution.

We have had a lot of PRs to try and improve the replacement command, but the specific command we included was meant as an example, and not as an official solution that would work in every case.

Thanks!

kwbtdisk:
@composerinteralia Oh, I had completely missed the lines 😅

> We have had a lot of PRs to try and improve the replacement command, but the specific command we included was meant as an example, and not as an official solution that would work in every case.
I understood, yeah there are various situation especially in the env of shell commands.

Thank you for your quick & kind reply!

composerinteralia:
Thank you for taking the time to open a pull request!

PR #1314, opened 2019-07-12
Pass the index of the object to a block when using *_list methods

mcountis:
When using one of the `*_list` methods, allow the block to receive the object's index.

```ruby
posts = build_list(:post, 10, title: 'Post number') do |post, i|
  post.title = "#{post.title} #{i + 1}"
end

posts.first.title # => "Post number 1"
# ...
posts.last.title  # => "Post number 10"
```

### DIFF
--- a/lib/factory_bot/strategy_syntax_method_registrar.rb
+++ b/lib/factory_bot/strategy_syntax_method_registrar.rb
@@ -29,10 +29,10 @@ module FactoryBot
           raise ArgumentError, "count missing for #{strategy_name}_list"
         end
 
-        Array.new(amount) { |i|
-          curried_block = block.nil? ? block : ->(*args) { block.call(*args<<i) }
+        Array.new(amount) do |i|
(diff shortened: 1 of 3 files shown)


composerinteralia:
I had played around with something similar a while back (see https://github.com/thoughtbot/factory_bot/tree/dc-list-index-play). One thing to consider is that calling the block with a different number of arguments could be a breaking change if anyone is passing in a lambda (lambdas raise an ArgumentError if you pass the wrong number of arguments).

@aledustet 's comment about introducing a new `*_list_with_index` method rather than changing the existing methods would solve this. That might be the cleanest approach. On my branch I was doing an arity check.

We should also add some documentation for this so people can find and use it.

composerinteralia:
Closing since there hasn't been activity for a while. I am not opposed to this change, and if anybody wants to take this to the finish line please feel free to open a separate PR (preserving the commit of the original author).

emmanuelgautier1994:
I'd love this change, and would like to get it done! Here's the new PR for it ⬆

PR #1315, opened 2019-07-12
Update yard from 0.9.19 to 0.9.20

aledustet:
Why:
A security vulnerability has been identified with the 0.9.19 version of
the yard gem. More details
[here](https://github.com/lsegal/yard/security/advisories/GHSA-xfhh-rx56-rxcr)

This PR:
Updates the yard gem version.


composerinteralia:
@aledustet I generally merge dependency updates once the tests pass, without getting any review. You are welcome to do the same if you like. It probably also makes sense to add Dependabot to this repo, with the option enabled to auto-merge once the tests pass.

PR #1316, opened 2019-07-12
Avoid stubbing id for records without primary key

### LINKED ISSUE #1305: ActiveModel::MissingAttributeError when using `build_stubbed`, primary_key is not set (Rails 6)
I am trying out Rails 6 RC1. It seems that when using `build_stubbed` I get 

```
ActiveModel::MissingAttributeError:
        can't write unknown attribute ``
```

I have been looking into it and it seems that `primary_key` isn't setup on the stub instance. 

In the stub strategy in `stub_database_interaction_on_result` the ID of the stub instance is set (https://github.com/thoughtbot/factory_bot/blob/master/lib/factory_bot/strategy/stub.rb#L47), however this fails with the above Error as `@primary_key` is `nil`.

Im not sure why this issue has started to occur,  Im assuming it was due to https://github.com/rails/rails/commit/b6828fc91531ae0cc0a0f216705dd19112596301 

I guess the solution is to explicitly call `primary_key=` when doing `build_stubbed`?

factory_bot version: 5.0.2 
Rails version : 6 RC1

composerinteralia:
Fixes #1305

Before Rails 6 when the build_stubbed strategy assigned an id on a
record without a primary key column, the `id=` method would no-op, and
the record would end up looking like a new_record (i.e. `new_record?`
would return true because the id was nil).

This problem surfaced because of a [change in Rails 6][rails], which
caused `id=` to raise a potentially confusing
`ActiveModel::MissingAttributeError: can't write unknown attribute ''`
for records without a primary key column.

Since build_stubbed stubbed was calling `id=` for all instances,
regardless of whether they had primary keys, it was raising the above
error.

To avoid this error, we check whether the instance has a primary_key
defined before setting the id.

We also changed `persisted?` and `new_record?` to be less dependent on
the id. That way those methods will work as expected for records without
primary keys.

[rails]: https://github.com/rails/rails/commit/b6828fc91531ae0cc0a0f216705dd19112596301

### DIFF
--- a/lib/factory_bot/strategy/stub.rb
+++ b/lib/factory_bot/strategy/stub.rb
@@ -44,15 +44,17 @@ module FactoryBot
       end
 
       def stub_database_interaction_on_result(result_instance)
-        result_instance.id ||= next_id
+        if has_settable_id?(result_instance)
+          result_instance.id ||= next_id
+        end
 
         result_instance.instance_eval do
           def persisted?
-            !new_record?
+            true
           end
 
           def new_record?
-            id.nil?
+            false
           end
 
           def destroyed?
@@ -66,6 +68,11 @@ module FactoryBot
             end
           end
         end
+      end
+
+      def has_settable_id?(result_instance)
+        !result_instance.class.respond_to?(:primary_key) ||
+          result_instance.class.primary_key
       end
 
       def clear_changes_information(result_instance)
--- a/spec/acceptance/stub_spec.rb
+++ b/spec/acceptance/stub_spec.rb
@@ -58,3 +58,44 @@ describe "a stubbed instance overriding strategy" do
     expect(subject.user).not_to be_new_record
   end
 end
+
+describe "a stubbed instance with no primary key" do
+  it "builds a stubbed instance" do
+    using_model_without_pk do
+      FactoryBot.define do
+        factory :model_without_pk
+      end
+
+      model = FactoryBot.build_stubbed(:model_without_pk)
+      expect(model).to be_truthy
+    end
+  end
+
+  it "behaves like a persisted record" do
+    using_model_without_pk do
+      FactoryBot.define do
+        factory :model_without_pk
+      end
+
+      model = FactoryBot.build_stubbed(:model_without_pk)
+      expect(model).to be_persisted
+      expect(model).not_to be_new_record
+    end
+  end
+
+  def using_model_without_pk
+    define_class("ModelWithoutPk", ActiveRecord::Base)
+
+    connection = ActiveRecord::Base.connection
+    begin
+      clear_generated_table("model_without_pks")
+      connection.create_table("model_without_pks", id: false) do |t|
+        t.column :updated_at, :datetime
+      end
+
+      yield
+    ensure
+      clear_generated_table("model_without_pks")
+    end
+  end
+end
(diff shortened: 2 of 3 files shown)


lib/factory_bot/strategy/stub.rb
```diff
@@ -44,15 +44,17 @@ def next_id
       end
 
       def stub_database_interaction_on_result(result_instance)
-        result_instance.id ||= next_id
+        if has_settable_id?(result_instance)
+          result_instance.id ||= next_id
```
aledustet: Maybe it's not "doable" but we might want to persist the old behavior. Like assigning the ```primary_key``` the value. That way old versions would stay relevant and even if the primary key is not id it could still be set. we could do:
```ruby
result_instance.send("#{result_instance.class.primary_key}=", next_id)
``` 
Would that cover the case where the primary key is overridden to a field other than id? Or does AR aliases the method to respond to id no matter what?

lib/factory_bot/strategy/stub.rb
```diff
@@ -44,15 +44,17 @@ def next_id
...
 
         result_instance.instance_eval do
           def persisted?
-            !new_record?
+            true
           end
 
           def new_record?
-            id.nil?
+            false
```
aledustet: for completeness maybe we could do the second part of the ```has_settable_id``` method here?

spec/acceptance/stub_spec.rb
```diff
@@ -58,3 +58,44 @@
     expect(subject.user).not_to be_new_record
   end
 end
+
+describe "a stubbed instance with no primary key" do
+  it "builds a stubbed instance" do
+    using_model_without_pk do
```
aledustet: this definitely addresses the case without a primary key, I think my question would be: Do we even wanna support cases where the primary key is not the id field?

lib/factory_bot/strategy/stub.rb
composerinteralia: Do you mean `result_instance.class.primary_key`? I don't follow how that relates to this method.

lib/factory_bot/strategy/stub.rb
composerinteralia: I think `id=` already does the correct thing here for primary keys with a non-standard name. [Here is the relevant rails code](https://github.com/rails/rails/blob/98a57aa5f610bc66af31af409c72173cdeeb3c9e/activerecord/lib/active_record/attribute_methods/primary_key.rb#L22-L25). If we did need to change that behavior, I think it would be beyond the scope of this PR.

lib/factory_bot/strategy/stub.rb
aledustet: sorry, I meant the instance method primary key instead of ```id.nil?``` ```primary_key.nil?```

lib/factory_bot/strategy/stub.rb
composerinteralia: I tested this out and we shouldn't need any special handling of custom primary keys names in factory_bot.

lib/factory_bot/strategy/stub.rb
composerinteralia: `build_stubbed` objects should always act like they are `persisted?`, so I think returning `false` is OK here. Checking whether the primary_key is `nil` still wouldn't work if there is no primary key at all.

Issue #1282, opened 2019-04-12
Improve experience with file uploads


composerinteralia:

composerinteralia:
https://stackoverflow.com/questions/55141549/attatching-activestorage-files-in-factorybot

odlp:
+1 to making this an easier experience.

Would this make sense as a feature / DSL-addon in Factory Bot Rails, rather than plain Factory Bot, since it's Rails specific functionality? Although we may then find it's less 'dicoverable' because I find myself reaching for the FB docs, not the FBR docs.

composerinteralia:
I had the same thought about this maybe being in factory_bot_rails in https://github.com/thoughtbot/factory_bot/pull/1317#issuecomment-513345549, although I agree with you that might be less discoverable.

composerinteralia:
Draft PR https://github.com/thoughtbot/factory_bot/pull/1327

composerinteralia:
I played around with a branch that uses `Rack::Test::UploadedFile`: https://github.com/thoughtbot/factory_bot/commit/8d9bcda8fa22e410b09c38a42ca5dd153466533e.

Trying this out on a real project I ended up with this diff:

```diff
-def photo(photo_name)
-  Rack::Test::UploadedFile.new(Rails.root.join("spec", "fixtures", "photos", photo_name))
-end
+FactoryBot.file_fixture_path = Rails.root.join("spec", "fixtures")
 
 FactoryBot.define do
   factory :client_membership do
@@ -303,7 +301,7 @@ FactoryBot.define do
     end
 
     trait :with_image do
-      image { photo("ralph-bot.jpg") }
+      image { file_fixture("photos/ralph-bot.jpg") }
     end
```

In an ideal world I wouldn't have to specify the `file_fixture_path`. But given that it is common to use factory_bot in the Rails console, where `ActiveSupport::TestCase.file_fixture_path` would not available, this may be the best option.

If we go this route it probably makes sense to allow setting the `file_fixture_path` in the DSL as well, like:

```rb
FactoryBot.define do
  file_fixture_path { "path/to/fixtures" }
end

dorianmariefr:
still an issue, would be nice to have `file_fixture` in factories ❤️

doutatsu:
Bump this

agrobbin:
For those who are still looking for a potential solution to this problem, and are also using Rails and RSpec, I took this approach:

```ruby
module FactoryBotFileFixtureHelpers
[...]

PR #1317, opened 2019-07-12
feature Add file_fixture helper to syntax methods

aledustet:
Why:
This is to add simple support to files on factories. This is the
original [issue](https://github.com/thoughtbot/factory_bot/issues/1282).

What:
It is a common behavior to interact with files when creating factories,
and it is used heavily, as demonstrated with the many examples in the
issue, this is a first draft and I would love to see more use-cases to
add to the acceptance test. For the time being we have at least the base
case added.

### DIFF
--- a/lib/factory_bot/syntax/methods.rb
+++ b/lib/factory_bot/syntax/methods.rb
@@ -129,6 +129,20 @@ module FactoryBot
           Internal.sequence_by_name(name).next
         end
       end
+
+      # Creates a copy of the file passed to attach to the attribute
+      #
+      # Arguments:
+      #   filepath: (String)
+      #     The path for the file that will be loaded. Should be a valid path
+      #     for a file that exists at that location.
+      #
+      # Returns:
+      #   An instance of the file loader class pointing to the file at the
+      #   provided path.
+      def file_fixture(file_path)
+        FileLoader.new(file_path)
+      end
     end
   end
 end
--- a/lib/factory_bot/syntax/methods.rb
+++ b/lib/factory_bot/syntax/methods.rb
@@ -129,20 +129,6 @@ module FactoryBot
           Internal.sequence_by_name(name).next
         end
       end
-
-      # Creates a copy of the file passed to attach to the attribute
-      #
-      # Arguments:
-      #   filepath: (String)
-      #     The path for the file that will be loaded. Should be a valid path
-      #     for a file that exists at that location.
-      #
-      # Returns:
-      #   An instance of the file loader class pointing to the file at the
-      #   provided path.
-      def file_fixture(file_path)
-        FileLoader.new(file_path)
-      end
     end
   end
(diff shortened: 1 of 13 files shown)


composerinteralia:
The primary use case for this feature is for use with ActiveStorage-backed models. It doesn't currently work there, so I think we need  some tests cases for that.

I have a model with `has_one_attached :video`, and I was able to create a factory with an attached video using:

```rb
after(:build) do |instance|
  instance.video.attach(
    io: File.open(Rails.root.join('spec', 'fixtures', 'files', 'video.mp4')),
    filename: 'video.mp4'
  )
end
```

or, using this Rack helper:

```rb
video { Rack::Test::UploadedFile.new("spec/fixtures/files/video.mp4") }
```

but I was unable to attach a video using the new method in this PR:

```rb
video { file_fixture("spec/fixtures/files/video.mp4") }
```

This didn't raise any error, but it silently failed to attach a video.

---

I think it is fine if this works for just the ActiveStorage case for now. We can always extend it other use cases as they emerge. We may even decide to move this into factory_bot_rails if we want to make it clear it is only meant to work with Rails.

That would allow us to be opinionated about expecting the fixture files to live in `ActiveSupport::TestCase.file_fixture_path`, ([defaults to "spec/fixtures/files" in rspec-rails](https://github.com/rspec/rspec-rails/blob/8c6c9590b94916199950dc8a91a9741d3be30c7c/lib/rspec/rails/configuration.rb#L81-L82)) and so we could specify the filename without the full path:

```rb
video { file_fixture_upload("video.mp4") }
```

This would match the [helper method used in the ActiveStorage test suite](https://github.com/rails/rails/blob/611339015fcc6056215edc6a95c20673a6245f7f/activestorage/test/test_helper.rb#L84-L86)

lib/factory_bot/syntax/methods.rb
```diff
@@ -129,6 +129,20 @@ def generate_list(name, count)
...
+      #
+      # Arguments:
+      #   filepath: (String)
+      #     The path for the file that will be loaded. Should be a valid path
+      #     for a file that exists at that location.
+      #
+      # Returns:
+      #   An instance of the file loader class pointing to the file at the
+      #   provided path.
+      def file_fixture(file_path)
```
composerinteralia: This should go on the definition proxy, since it only needs to be available inside the attribute block. Otherwise when people include these syntax methods into their test suite it will end up overwriting the original ActiveSupport `file_fixture` method.

aledustet:
Added a new

lib/factory_bot/syntax/methods.rb
TylerRick: Good idea, we don't want to overwrite the original ActiveSupport file_fixture method.

But we also need to be able to call file_fixture from transient attributes and callbacks...

Issue #1332, opened 2019-09-29
Performance issues with traits


jeffcarbs:
I'm using `FactoryBot` to test a small ruby gem and when I started using traits I noticed a pretty significant uptick in test runtime. I put together a script to show the performance impact even a few traits can have on factories.

I've been investigating a bit and have found what's likely the biggest culprit ([using a `detect` when we probably want a hash lookup](https://github.com/thoughtbot/factory_bot/blob/master/lib/factory_bot/definition.rb#L113-L115)). I'll send over a PR to improve that. However, it looks like even after fixing that there are more inefficiencies when using traits that get worse the more traits you have or the more traits you use on a given `build` or `create` call.

Also I wanted to call out that while I found this testing a small ruby gem my company's main application test suite has 500+ usages of `traits` (across ~100 different factories). One of our core models has over 25 traits on it.

### Steps to reproduce
Running the [reproduction script](https://gist.github.com/jeffcarbs/bc988c4b08aa81e013d22e8feb11c476) will show benchmark results of various scenarios.

### Expected behavior
Traits would have negligible performance overhead.

### Actual behavior
Using a single trait on a factory with one trait defined more than doubles the runtime of `build`. As you define more traits and/or use more of the defined traits performance gets worse.

Raw results:
```
Test one (n=10000): not setting anything
       user     system      total        real
Has  0 traits  0.630000   0.000000   0.630000 (  0.637583)
Has  1 trait   0.630000   0.000000   0.630000 (  0.631442)
Has  2 traits  0.710000   0.010000   0.720000 (  0.734025)
Has  3 traits  0.620000   0.010000   0.630000 (  0.622784)
Has 50 traits  0.610000   0.000000   0.610000 (  0.624279)

Test two (n=10000): passing attributes without using traits
       user     system      total        real
Has  0 traits  1.070000   0.010000   1.080000 (  1.079745)
Has  1 trait   1.090000   0.010000   1.100000 (  1.106030)
Has  2 traits  1.120000   0.000000   1.120000 (  1.126111)
Has  3 traits  1.220000   0.030000   1.250000 (  1.253315)
Has 50 traits  1.140000   0.000000   1.140000 (  1.152264)

Test three (n=10000): passing attributes and using traits
       user     system      total        real
Has  0 traits, using 0  1.160000   0.020000   1.180000 (  1.190228)
Has  1 trait,  using 1   2.600000   0.020000   2.620000 (  2.636136)
[...]

PR #1333, opened 2019-09-29
Add memoization to `trait_for`

jeffcarbs:
Doesn't entirely fix but partially addresses #1332. Looking at the results below you can still see that even after this PR the following are true (although less glaringly):
- the more traits you have, the slower it will be to use any traits in a `build` or `create`
- the more traits you use on a single `build` or `create` the slower it will be.

Using the performance testing script I put together: https://gist.github.com/jeffcarbs/bc988c4b08aa81e013d22e8feb11c476

**Before**
```
Test five (n=10000): using different amounts of traits on factory with a 3 traits without passing attributes
       user     system      total        real
Has 3 traits, using 1  1.890000   0.000000   1.890000 (  1.899632)
Has 3 traits, using 2  2.310000   0.010000   2.320000 (  2.318839)
Has 3 traits, using 3  2.690000   0.000000   2.690000 (  2.699934)

Test six (n=10000): using different amounts of traits on factory with 50 traits without passing attributes
       user     system      total        real
Has 50 traits, using 1  6.720000   0.010000   6.730000 (  6.735540)
Has 50 traits, using 2  8.920000   0.020000   8.940000 (  8.958390)
Has 50 traits, using 3 10.700000   0.010000  10.710000 ( 10.725242)
Has 50 traits, using 4 13.410000   0.010000  13.420000 ( 13.439983)
```

**After**
```
Test five (n=10000): using different amounts of traits on factory with a 3 traits without passing attributes
       user     system      total        real
Has 3 traits, using 1  1.770000   0.000000   1.770000 (  1.765303)
Has 3 traits, using 2  1.950000   0.000000   1.950000 (  1.961503)
Has 3 traits, using 3  2.160000   0.000000   2.160000 (  2.160803)

Test six (n=10000): using different amounts of traits on factory with 50 traits without passing attributes
       user     system      total        real
Has 50 traits, using 1  2.590000   0.010000   2.600000 (  2.588208)
Has 50 traits, using 2  3.150000   0.000000   3.150000 (  3.157104)
Has 50 traits, using 3  3.900000   0.010000   3.910000 (  3.926030)
Has 50 traits, using 4  4.450000   0.010000   4.460000 (  4.467921)
```


lib/factory_bot/definition.rb
```diff
@@ -114,14 +114,28 @@ def trait_by_name(name)
...
     end
 
+    # This is a hot path when using traits so:
+    # - Memoize the defined_traits by name so we can lookup in a hash instead of array
+    # - Store both string/symbol names to avoid needing to cast on lookup
+    # - Use plain Hash since HashWithIndifferentAccess incurs significant overhead
     def trait_for(name)
-      defined_traits.detect { |trait| trait.name == name.to_s }
+      @defined_traits_by_name ||= defined_traits.each_with_object({}) do |trait, memo|
+        # This method was previously using `detect` which means if there were
```
composerinteralia: I think this comment makes more sense in the body of the commit message. The specific spec and line number are likely to change, which would make the comment inaccurate.

lib/factory_bot/definition.rb
```diff
@@ -114,14 +114,28 @@ def trait_by_name(name)
...
+        # multiple traits the first was returned. It doesn't seem like you should be able
+        # to register duplicate traits but without this logic a test was breaking:
+        # spec/acceptance/traits_spec.rb:136 (the one with female_admin_judy)
+        next if memo[trait.name]
+        memo[trait.name] = trait
+        memo[trait.name.to_sym] = trait
+      end
+      @defined_traits_by_name[name]
```
ngan: Any reason to not to_s both the storage of the memorized trait name and the passed in name?

lib/factory_bot/definition.rb
```diff
@@ -114,14 +114,28 @@ def trait_by_name(name)
...
+    # This is a hot path when using traits so:
+    # - Memoize the defined_traits by name so we can lookup in a hash instead of array
+    # - Store both string/symbol names to avoid needing to cast on lookup
+    # - Use plain Hash since HashWithIndifferentAccess incurs significant overhead
     def trait_for(name)
-      defined_traits.detect { |trait| trait.name == name.to_s }
+      @defined_traits_by_name ||= defined_traits.each_with_object({}) do |trait, memo|
+        # This method was previously using `detect` which means if there were
+        # multiple traits the first was returned. It doesn't seem like you should be able
+        # to register duplicate traits but without this logic a test was breaking:
```
composerinteralia: This reminds me of https://github.com/thoughtbot/factory_bot/issues/968. I think there was talk at once point of using inheritance for traits instead of copying them all over to the child factories.

lib/factory_bot/definition.rb
composerinteralia: They mention "- Store both string/symbol names to avoid needing to cast on lookup" in the comments. I wouldn't assume we get a huge performance win from this, but we do save an allocation each time we look up a trait.

It does occur to me that this could be a breaking change if anyone is passing in objects that don't respond to `to_sym` as trait names. I don't imagine people are passing in much beyond strings and symbols, but it is still worth thinking about.

jeffcarbs: Trait names are cast to a string on initialization so we should be ok on the latter concern: https://github.com/thoughtbot/factory_bot/blob/master/lib/factory_bot/trait.rb#L7

lib/factory_bot/definition.rb
jeffcarbs: Good call out. My intention was to explicitly call out behavior that’s a bit implicit but I may try adding a test for this purpose instead.

lib/factory_bot/definition.rb
jeffcarbs: Actually I might have a case where we would need to cast on lookup, going to try to repro with a test.

jeffcarbs: OK so you can technically define a trait with any value and the name is cast `to_s` on creation so we need to also cast to a string on lookup. For example these are technically valid traits:
```ruby
trait 42 do
  some_field { nil }
end

trait [1,2,3] do
  some_field { nil }
end
```

I backfilled specs to exercise this behavior and updated the memoization/lookup to be simpler.

lib/factory_bot/definition.rb
jeffcarbs: I added tests to more explicitly test this behavior. It ensures that traits defined multiple times in the same factory will use the first one and you're able to redefine the trait in child factories. For the former (defining a trait twice in same factory) seems like it's something that should raise an exception although that would be a breaking change and definitely outside of the scope of this PR.

jeffcarbs:
OK this PR is ready for review again. I separated out the changed into two commits:
- the first backfills specs. I wanted to split this out to ensure the specs pass on the existing codebase
- the second adds memoization to the `traits_for` method.

spec/acceptance/traits_spec.rb
```diff
@@ -20,6 +20,10 @@
           great { "GREAT!!!" }
         end
 
+        trait :great do
```
composerinteralia: I agree with you that defining a trait twice in the same factory should probably raise an error. Somebody started down that path in https://github.com/thoughtbot/factory_bot/pull/1064. I doesn't seem trivial given the ways traits currently work.

spec/acceptance/traits_spec.rb
```diff
@@ -182,41 +204,71 @@
 
 describe "trait indifferent access" do
   context "when trait is defined as a string" do
-    it "can be invoked with a string" do
-      build_user_factory_with_admin_trait("admin")
+    before { build_user_factory_with_admin_trait("admin") }
```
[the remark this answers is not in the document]
jeffcarbs: Sounds good, I'll defer to the style of codebase. Just reverted this and the rest of the usages of `before` in favor of inlining the test setup.

jeffcarbs:
@composerinteralia - Thanks for the review! I addressed your one suggestion for avoiding `before` and the build is green so I think this is good to go whenever you're ready 👍

composerinteralia:
Many thanks for your well written issue, and for taking the time to improve factory_bot!

Issue #1331, opened 2019-09-25
factory_bot 5.1.0 does not play nicely with jruby (lambdas don't splat the way they should)


dsimmsatsquare:
### Steps to reproduce
```
~/Development/scrach master * bundle console
jruby-9.1.17.0 :001 > require 'factory_bot'
ArgumentError: wrong number of arguments (1 for 2)
	from /Users/dsimms/.rvm/gems/jruby-9.1.17.0/gems/factory_bot-5.1.0/lib/factory_bot/internal.rb:89:in `register_strategy'
	from org/jruby/RubyHash.java:1350:in `each'
	from /Users/dsimms/.rvm/gems/jruby-9.1.17.0/gems/factory_bot-5.1.0/lib/factory_bot/internal.rb:89:in `register_default_strategies'
	from /Users/dsimms/.rvm/gems/jruby-9.1.17.0/gems/factory_bot-5.1.0/lib/factory_bot.rb:124:in `<main>'
	from org/jruby/RubyKernel.java:956:in `require'
	from (irb):1:in `<eval>'
	from org/jruby/RubyKernel.java:995:in `eval'
	from (irb):1:in `evaluate'
	from org/jruby/RubyKernel.java:1316:in `loop'
	from org/jruby/RubyKernel.java:1138:in `catch'
	from org/jruby/RubyKernel.java:1138:in `catch'
	from /Users/dsimms/.rvm/gems/jruby-9.1.17.0/gems/bundler-2.0.2/lib/bundler/cli/console.rb:19:in `run'
	from /Users/dsimms/.rvm/gems/jruby-9.1.17.0/gems/bundler-2.0.2/lib/bundler/cli.rb:496:in `console'
	from /Users/dsimms/.rvm/gems/jruby-9.1.17.0/gems/bundler-2.0.2/lib/bundler/vendor/thor/lib/thor/command.rb:27:in `run'
	from /Users/dsimms/.rvm/gems/jruby-9.1.17.0/gems/bundler-2.0.2/lib/bundler/vendor/thor/lib/thor/invocation.rb:126:in `invoke_command'
	from /Users/dsimms/.rvm/gems/jruby-9.1.17.0/gems/bundler-2.0.2/lib/bundler/vendor/thor/lib/thor.rb:387:in `dispatch'
	from /Users/dsimms/.rvm/gems/jruby-9.1.17.0/gems/bundler-2.0.2/lib/bundler/cli.rb:27:in `dispatch'
[...]

PR #1334, opened 2019-09-30
Move sqlite3 loading to Gemfile

newellista:
which allows us to load the `activerecord-jdbcsqlite3-adapter` when
running on JRuby.

This, combined with pinning the version to > 50.0 allows most of the
specs to pass under JRuby.

Additionally, I changed the way `DEFAULT_STRATEGIES` are registered, working around a JRuby issue.

The final failing spec has to do with how JRuby handles hashes.  I
wasn't able to completely isolate the problem, or resolve it
satisfactorily.  However, I made a small change to the expectation in
the spec that accommodated both MRI and JRuby.

Resolves: https://github.com/thoughtbot/factory_bot/issues/1331


newellista:
Closing in favor of https://github.com/thoughtbot/factory_bot/pull/1335

PR #1338, opened 2019-10-11
Refactor aliases_spec.rb to more closely adhere to the Let's Not school of RSpec

richiethomas:
For reference:

TL;DR- there is consensus within thoughtbot that RSpec's `subject`, `before`, and `let` features are over-used in the Ruby community, and result in relatively hard-to-read tests.

This is the first of several PRs which will in-line many of the declarations previously made using these idioms.

### DIFF
--- a/spec/factory_bot/aliases_spec.rb
+++ b/spec/factory_bot/aliases_spec.rb
@@ -1,19 +1,19 @@
 describe FactoryBot, "aliases" do
-  it "aliases for an attribute should include the original attribute and a version suffixed with '_id'" do
+  it "for an attribute should include the original attribute and a version suffixed with '_id'" do
     aliases = FactoryBot.aliases_for(:test)
 
     expect(aliases).to include(:test)
     expect(aliases).to include(:test_id)
   end
 
-  it "aliases for a foreign key should include both the suffixed and un-suffixed variants" do
+  it "for a foreign key should include both the suffixed and un-suffixed variants" do
     aliases = FactoryBot.aliases_for(:test_id)
 
     expect(aliases).to include(:test)
     expect(aliases).to include(:test_id)
   end
 
-  it "aliases for an attribute which starts with an underscore should not include a non-underscored version" do
+  it "for an attribute which starts with an underscore should not include a non-underscored version" do
     aliases = FactoryBot.aliases_for(:_id)
 
     expect(aliases).not_to include(:id)
--- a/spec/factory_bot/aliases_spec.rb
+++ b/spec/factory_bot/aliases_spec.rb
@@ -1,29 +1,42 @@
 describe FactoryBot, "aliases" do
-  context "aliases for an attribute" do
-    subject { FactoryBot.aliases_for(:test) }
-    it      { should include(:test) }
-    it      { should include(:test_id) }
+  it "aliases for an attribute should include the original attribute and a version suffixed with '_id'" do
+    aliases = FactoryBot.aliases_for(:test)
+
+    expect(aliases).to include(:test)
+    expect(aliases).to include(:test_id)
   end
 
-  context "aliases for a foreign key" do
-    subject { FactoryBot.aliases_for(:test_id) }
-    it      { should include(:test) }
-    it      { should include(:test_id) }
+  it "aliases for a foreign key should include the un-suffixed attribute name" do
+    aliases = FactoryBot.aliases_for(:test_id)
+
+    expect(aliases).to include(:test)
   end
 
-  context "aliases for an attribute starting with an underscore" do
-    subject { FactoryBot.aliases_for(:_id) }
-    it      { should_not include(:id) }
+  it "aliases for a foreign key should include the original attribute name" do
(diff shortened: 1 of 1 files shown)


mike-burns:
A lot of these have the same setup. How about either combining them or abstracting that setup away? I'd probably combine them.

There might be a better name than `subject`. For the first group, maybe `aliases`?

spec/factory_bot/aliases_spec.rb
```diff
@@ -1,29 +1,47 @@
 describe FactoryBot, "aliases" do
-  context "aliases for an attribute" do
-    subject { FactoryBot.aliases_for(:test) }
-    it      { should include(:test) }
-    it      { should include(:test_id) }
+  it "aliases for an attribute should include the original attribute" do
+    subject = FactoryBot.aliases_for(:test)
```
eebs: What do you think of using a more descriptive name than `subject`? I find naming in specs really helps me understand what the implementation does.

```ruby
    aliases = FactoryBot.aliases_for(:test)

    expect(aliases).to include(:test)
```

richiethomas:
> A lot of these have the same setup. How about either combining them or abstracting that setup away? I'd probably combine them.

@mike-burns something like the following?

```
  it "aliases for an attribute should include the original attribute and a version suffixed with '_id'" do
    aliases = FactoryBot.aliases_for(:test)

    expect(aliases).to include(:test)
    expect(aliases).to include(:test_id)
  end
```

spec/factory_bot/aliases_spec.rb
```diff
@@ -1,29 +1,47 @@
...
-  before do
+  it "the list of aliases should include an un-suffixed equivalent" do
     FactoryBot.aliases << [/(.*)_suffix/, '\1']
+    subject = FactoryBot.aliases_for(:test_suffix)
+
+    expect(subject).to include(:test)
   end
 
-  subject { FactoryBot.aliases_for(:test_suffix) }
+  it "the list of aliases should include the original name plus an '_id' suffix" do
```
eebs: This description is very similar to the one on line  8. Is this testing something else? I'm a little confused by this test.

mike-burns:
That seems right. 

The test description doesn't quite read right. When run with `-fd`it'll print something like "FactoryBot aliases aliases for an attribute should include the original attribute and a version suffixed with '_id'".

Maybe introduce a context for `.aliases_for`, and then describe it in the `it`?

```ruby
context ".aliases_for" do
  it "includes the original attribute plus an _id-suffixed variant"
end
```

spec/factory_bot/aliases_spec.rb
richiethomas: Thanks @eebs !  I just re-named it to `aliases`.

richiethomas:
When I run the tests with the `-fd` flag, I see the following:

Screen Shot 2019-10-11 at 10 56 24 AM

@mike-burns do you mean the presence of the word "aliases" in both the white text and green text is confusing?

I've been coached (by other thoughtbots) to avoid `context` blocks as well if possible.  The feedback I got was that `context` can inhibit readability.  I originally removed it from this spec but happy to add it back if it results in a middle ground that everyone can live with.  Or maybe it's worth a wider discussion?

mike-burns:
Oh I also never use `context`, but yes that's what I mean.

mike-burns:
Oh I also never use `context`, but yes, that double "alias" is what I mean.

Might be worth having a larger discussion with others here. I don't know if we have a standard around things. I know we got rid of the word "should" in lieu of describing how it acts ("frobs the blobs" instead of "should frob the blob").

eliza-jane:
Echoing those who have also reviewed, I do really like where this is going. :) Will check back in to join the discussion as this progresses.

spec/factory_bot/aliases_spec.rb
```diff
@@ -1,29 +1,31 @@
 describe FactoryBot, "aliases" do
-  context "aliases for an attribute" do
-    subject { FactoryBot.aliases_for(:test) }
-    it      { should include(:test) }
-    it      { should include(:test_id) }
+  it "for an attribute should include the original attribute and a version suffixed with '_id'" do
+    aliases = FactoryBot.aliases_for(:test)
+
+    expect(aliases).to include(:test)
+    expect(aliases).to include(:test_id)
```
richiethomas: The thing I *don't* like about this is that, when assertions are combined like this, if one of them fails, you'd have to look up file name and line number from the backtrace to know which one it was (as opposed to just reading the test description at the CLI).  Thoughts?

mike-burns: How about this?

```suggestion
    expect(aliases).to include(:test, :test_id)
```

Does that give a useful failure message?

richiethomas: Screen Shot 2019-10-11 at 2 30 22 PM


I feel like it does.  Anyone feel different?

richiethomas: I changed it to `.not_to` for the purposes of generating a failure, but I'd imagine a `.to` failure would look similar.

PR #1339, opened 2019-10-15
Refactor attribute_spec.rb to conform with Let's Not style

richiethomas:
For reference:

### DIFF
--- a/spec/factory_bot/attribute_spec.rb
+++ b/spec/factory_bot/attribute_spec.rb
@@ -1,7 +1,15 @@
 describe FactoryBot::Attribute do
-  let(:name)  { "user" }
-  subject     { FactoryBot::Attribute.new(name, false) }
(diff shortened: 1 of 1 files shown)


spec/factory_bot/attribute_spec.rb
```diff
@@ -1,7 +1,15 @@
 describe FactoryBot::Attribute do
-  let(:name)  { "user" }
-  subject     { FactoryBot::Attribute.new(name, false) }
+  it "should convert the name attribute to a symbol" do
```
composerinteralia: I don't have strong feelings about this at all, but I think it is fun that somebody was opinionated enough about removing the word "should" that they created a gem for it: https://github.com/should-not/should_not

richiethomas: 😆

PR #1340, opened 2019-10-15
Refactor declaration_list_spec.rb to conform to Let's Not style

richiethomas:
For reference:

### DIFF
--- a/spec/factory_bot/declaration_list_spec.rb
+++ b/spec/factory_bot/declaration_list_spec.rb
@@ -1,30 +1,25 @@
 describe FactoryBot::DeclarationList, "#attributes" do
-  let(:attribute_1) { double("attribute 1") }
-  let(:attribute_2) { double("attribute 2") }
-  let(:attribute_3) { double("attribute 3") }
-  let(:declaration_1) do
-    double(
-      "declaration 1",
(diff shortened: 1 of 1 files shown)


spec/factory_bot/declaration_list_spec.rb
```diff
@@ -1,74 +1,72 @@
...
+    expect(declaration_list.to_a).to eq [declaration1]
+
+    declaration_list.declare_attribute(declaration2)
+    expect(declaration_list.to_a).to eq [declaration1, declaration2]
+  end
 
-      subject.declare_attribute(declaration_2)
-      expect(subject.to_a).to eq [declaration_1, declaration_2]
-    end
+  it "deletes declarations with the same name when overridable" do
```
composerinteralia: It seems to me there is a test case missing here: `it "adds declarations with the same name to the list when NOT overridable"`. Do you want to add it?

richiethomas: Done.  I *kinda* wish the `.overridable?` method was public instead of private, so I could assert against it in this new test and make the intention a bit more clear.  Unfortunately that's not really advisable if that public method only gets called in the test suite.

PR #1341, opened 2019-10-15
Refactor attribute_list_spec.rb to conform to Let's Not style

richiethomas:
For reference:

### DIFF
--- a/spec/factory_bot/attribute_list_spec.rb
+++ b/spec/factory_bot/attribute_list_spec.rb
@@ -2,25 +2,28 @@ describe FactoryBot::AttributeList, "#define_attribute" do
   it "maintains a list of attributes" do
     attribute = double(:attribute, name: :attribute_name)
     another_attribute = double(:attribute, name: :another_attribute_name)
+    list = FactoryBot::AttributeList.new
 
-    subject.define_attribute(attribute)
-    expect(subject.to_a).to eq [attribute]
+    list.define_attribute(attribute)
+    expect(list.to_a).to eq [attribute]
 
-    subject.define_attribute(another_attribute)
-    expect(subject.to_a).to eq [attribute, another_attribute]
+    list.define_attribute(another_attribute)
+    expect(list.to_a).to eq [attribute, another_attribute]
   end
 
   it "returns the attribute" do
     attribute = double(:attribute, name: :attribute_name)
+    list = FactoryBot::AttributeList.new
 
-    expect(subject.define_attribute(attribute)).to eq attribute
+    expect(list.define_attribute(attribute)).to eq attribute
   end
 
   it "raises if an attribute has already been defined" do
     attribute = double(:attribute, name: :attribute_name)
+    list = FactoryBot::AttributeList.new
 
     expect do
-      2.times { subject.define_attribute(attribute) }
+      2.times { list.define_attribute(attribute) }
     end.to raise_error(
       FactoryBot::AttributeDefinitionError,
       "Attribute already defined: attribute_name",
@@ -29,14 +32,12 @@ describe FactoryBot::AttributeList, "#define_attribute" do
 end
 
 describe FactoryBot::AttributeList, "#define_attribute with a named attribute list" do
-  subject { FactoryBot::AttributeList.new(:author) }
-
-  let(:association_with_same_name)      { FactoryBot::Attribute::Association.new(:author, :author, {}) }
-  let(:association_with_different_name) { FactoryBot::Attribute::Association.new(:author, :post, {}) }
-
   it "raises when the attribute is a self-referencing association" do
+    association_with_same_name = FactoryBot::Attribute::Association.new(:author, :author, {})
+    list = FactoryBot::AttributeList.new(:author)
+
     expect do
-      subject.define_attribute(association_with_same_name)
+      list.define_attribute(association_with_same_name)
     end.to raise_error(
       FactoryBot::AssociationDefinitionError,
       "Self-referencing association 'author' in 'author'",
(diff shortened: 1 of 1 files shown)


spec/factory_bot/attribute_list_spec.rb
```diff
@@ -59,28 +63,30 @@ def list(*attributes)
...
-  let(:profile_attribute)   { FactoryBot::Attribute::Association.new(:profile, :profile, {}) }
+  it "returns associations" do
+    email_attribute = FactoryBot::Attribute::Dynamic.new(
+      :email,
+      false,
+      ->(u) { "#{u.full_name}@example.com" },
+    )
+    author_attribute = FactoryBot::Attribute::Association.new(:author, :user, {})
+    profile_attribute = FactoryBot::Attribute::Association.new(:profile, :profile, {})
+    list = FactoryBot::AttributeList.new
```
composerinteralia: I think we could use the `build_attribute_list` function here and in the specs below as well:

```rb
list = build_attribute_list(
  email_attribute,
  author_attribute,
  profile_attribute,
)
```

spec/factory_bot/attribute_list_spec.rb
```diff
@@ -123,28 +135,51 @@ def build_association(name)
...
-    subject.define_attribute(build_non_ignored_attribute(:email))
-    subject.define_attribute(build_non_ignored_attribute(:first_name))
-    subject.define_attribute(build_non_ignored_attribute(:last_name))
-    subject.define_attribute(build_association(:avatar))
-  end
-
   it "knows all its #names" do
-    expect(subject.names).to eq [:comments_count, :posts_count, :email, :first_name, :last_name, :avatar]
+    list = FactoryBot::AttributeList.new
+    list.define_attribute(build_ignored_attribute(:comments_count))
```
composerinteralia: Is there a reason to have more than one of each type in these tests?

spec/factory_bot/attribute_list_spec.rb
```diff
@@ -93,20 +99,26 @@ def build_non_ignored_attribute(name)
...
   it "filters #ignored attributes" do
-    expect(subject.ignored.map(&:name)).to eq [:comments_count, :posts_count]
+    list = FactoryBot::AttributeList.new
+    list.define_attribute(build_ignored_attribute(:comments_count))
+    list.define_attribute(build_ignored_attribute(:posts_count))
+    list.define_attribute(build_non_ignored_attribute(:email))
+    list.define_attribute(build_non_ignored_attribute(:first_name))
+    list.define_attribute(build_non_ignored_attribute(:last_name))
+
+    expect(list.ignored.map(&:name)).to eq [:comments_count, :posts_count]
```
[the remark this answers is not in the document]
richiethomas: Will fix

spec/factory_bot/attribute_list_spec.rb
```diff
@@ -93,20 +99,26 @@ def build_non_ignored_attribute(name)
...
   it "filters #non_ignored attributes" do
-    expect(subject.non_ignored.map(&:name)).to eq [:email, :first_name, :last_name]
+    list = FactoryBot::AttributeList.new
+    list.define_attribute(build_ignored_attribute(:comments_count))
+    list.define_attribute(build_ignored_attribute(:posts_count))
+    list.define_attribute(build_non_ignored_attribute(:email))
+    list.define_attribute(build_non_ignored_attribute(:first_name))
+    list.define_attribute(build_non_ignored_attribute(:last_name))
+
+    expect(list.non_ignored.map(&:name)).to eq [:email, :first_name, :last_name]
```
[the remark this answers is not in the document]
richiethomas: Will fix

spec/factory_bot/attribute_list_spec.rb
```diff
@@ -59,28 +63,30 @@ def list(*attributes)
     attribute1 = double(:attribute1, name: :attribute1)
     attribute2 = double(:attribute2, name: :attribute2)
     attribute3 = double(:attribute3, name: :attribute3)
+    list = FactoryBot::AttributeList.new
 
-    subject.define_attribute(attribute1)
-    subject.apply_attributes(list(attribute2, attribute3))
-    expect(subject.to_a).to eq [attribute1, attribute2, attribute3]
+    list.define_attribute(attribute1)
+    list.apply_attributes(list(attribute2, attribute3))
```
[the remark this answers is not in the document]
richiethomas: Fixed

spec/factory_bot/attribute_list_spec.rb
richiethomas: Fixed

spec/factory_bot/attribute_list_spec.rb
richiethomas: Probably not.  I suppose this was a stylistic choice made by the original test author.  I noticed this kind of duplication in a couple of tests.  I'll remove it, might make the test easier to grok.

spec/factory_bot/attribute_list_spec.rb
```diff
@@ -1,26 +1,35 @@
+def build_attribute_list(*attributes)
+  FactoryBot::AttributeList.new.tap do |list|
+    attributes.each { |attribute| list.define_attribute(attribute) }
+  end
+end
+
 describe FactoryBot::AttributeList, "#define_attribute" do
```
richiethomas: I noticed there are separate top-level describe blocks used to test each method in the AttributeList class (as opposed to one top-level describe block and multiple inner blocks, which is what I'm used to seeing in other code bases).  Does the former choice have any benefits (or drawbacks) which don't occur in the latter?

spec/factory_bot/attribute_list_spec.rb
```diff
@@ -1,26 +1,35 @@
+def build_attribute_list(*attributes)
```
composerinteralia: Fun Ruby fact: if we put this here we end up defining a private `build_attribute_list` method on `Object`.

If we did move to having one top-level describe, instead of separate describe blocks per method, we could move `build_attribute_list` inside of that describe block and avoid defining it globally. Another approach would be to put the method inside a module and include wherever we need it.

richiethomas: Good to know.  I went with the module option. 👍

PR #1343, opened 2019-10-16
Refactor definition_spec.rb to conform to Let's Not style

richiethomas:
For reference:

### DIFF
--- a/spec/factory_bot/definition_spec.rb
+++ b/spec/factory_bot/definition_spec.rb
@@ -1,75 +1,75 @@
 describe FactoryBot::Definition do
-  subject { described_class.new(:name) }
+  it "delegates :declare_attribute to declarations" do
+    definition = described_class.new(:name)
 
-  it { should delegate(:declare_attribute).to(:declarations) }
-
-  describe "with a name" do
-    it "creates a new attribute list with the name passed" do
-      name = "great name"
-      allow(FactoryBot::DeclarationList).to receive(:new)
-
-      FactoryBot::Definition.new(name)
-
-      expect(FactoryBot::DeclarationList).to have_received(:new).with(name)
-    end
+    expect(definition).to delegate(:declare_attribute).to(:declarations)
   end
 
-  describe "#name" do
-    it "returns the name" do
-      name = "factory name"
-      definition = described_class.new(name)
+  it "creates a new attribute list with the name passed when given a name" do
+    name = "great name"
+    allow(FactoryBot::DeclarationList).to receive(:new)
 
-      expect(definition.name).to eq(name)
-    end
+    FactoryBot::Definition.new(name)
+
+    expect(FactoryBot::DeclarationList).to have_received(:new).with(name)
   end
 
-  describe "#overridable" do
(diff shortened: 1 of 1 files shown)


spec/factory_bot/definition_spec.rb
```diff
@@ -1,75 +1,75 @@
...
+    definition.add_callback(callback1)
+    definition.add_callback(callback2)
+
+    expect(definition.callbacks).to eq [callback1, callback2]
   end
 
-  describe "adding callbacks" do
-    let(:callback_1) { "callback1" }
-    let(:callback_2) { "callback2" }
+  it "#to_create is equal to nil when not specified" do
```
composerinteralia: `it "#to_create` reads a bit funny. Should these two tests be in a `describe "#to_create"` block?

richiethomas: I'm also a fan of making the descriptions in an `it` block as readable as possible, and I agree that this definition doesn't pass that test.

In order to improve it, I'd like to understand what the `#to_create` method and its associated instance variable do.  I was unable to figure it out on my own, as I found this class a bit harder to read and understand than many of the others whose tests I worked on.  Can you shed any light on that?

composerinteralia: `to_create` is ultimately used to [determine how `FactoryBot.create` should behave](https://github.com/thoughtbot/factory_bot/blob/de024de9b233a6791ca5b2dc3db85df2c3624608/GETTING_STARTED.md#custom-methods-to-persist-objects). The [default behavior](https://github.com/thoughtbot/factory_bot/blob/de024de9b233a6791ca5b2dc3db85df2c3624608/lib/factory_bot/configuration.rb#L22) is to call `save!` on the object, but you can override that either at the [top level](https://github.com/thoughtbot/factory_bot/blob/de024de9b233a6791ca5b2dc3db85df2c3624608/lib/factory_bot/syntax/default.rb#L36-L38) or inside of a specific factory by setting it on that factory's [definition](https://github.com/thoughtbot/factory_bot/blob/de024de9b233a6791ca5b2dc3db85df2c3624608/lib/factory_bot/definition.rb#L30-L36) (through the [definition proxy](https://github.com/thoughtbot/factory_bot/blob/de024de9b233a6791ca5b2dc3db85df2c3624608/lib/factory_bot/definition_proxy.rb#L163-L165)). This code is a bit confusing, and involves factories inheriting the [DefinitionHeirarchy](https://github.com/thoughtbot/factory_bot/blob/de024de9b233a6791ca5b2dc3db85df2c3624608/lib/factory_bot/definition_hierarchy.rb) from their parent factories. I would be happy to walk you through it if you like.

richiethomas: @composerinteralia a walkthrough would probably be helpful.  Let's sync up on a time for that.

richiethomas: I struggled with this a bit and pushed up a new description.  I'm not in love with it, but I think it's an improvement.  Open to discussion, of course.

PR #1344, opened 2019-10-16
Refactor disallows_duplicates_registry_spec.rb to conform to Let's Not style

richiethomas:
For reference:

### DIFF
--- a/spec/factory_bot/disallows_duplicates_registry_spec.rb
+++ b/spec/factory_bot/disallows_duplicates_registry_spec.rb
@@ -1,7 +1,7 @@
 describe FactoryBot::Decorator::DisallowsDuplicatesRegistry do
   it "delegates #register to the registry when not registered" do
     registry = double("registry", name: "Great thing", register: true)
-    decorator = described_class.new(registry)
+    decorator = FactoryBot::Decorator::DisallowsDuplicatesRegistry.new(registry)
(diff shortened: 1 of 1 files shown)


spec/factory_bot/disallows_duplicates_registry_spec.rb
```diff
@@ -1,17 +1,19 @@
 describe FactoryBot::Decorator::DisallowsDuplicatesRegistry do
-  let(:registry) { double("registry", name: "Great thing", register: true) }
-
-  subject { described_class.new(registry) }
-
   it "delegates #register to the registry when not registered" do
+    registry = double("registry", name: "Great thing", register: true)
+    decorator = described_class.new(registry)
```
composerinteralia: I remember @purinkle having negative opinions about `described_class` although I can't remember the exact reasoning why.

richiethomas: Looks like there's some discussion around deprecating it, due to some people being confused by its API:

https://blog.rubystyle.guide/rspec/2019/08/04/rspec-described_class.html

If true, that could be a big reason not to use it.

richiethomas: I've been trying to replace it with inlined classes in other refactorings I've done (forgot to do so here but will change it).  I'll probably double down on doing so after reading the above links.

PR #1348, opened 2019-10-18
Refactor null_object_spec.rb to conform to Let's Not style

richiethomas:
For reference:


spec/factory_bot/null_object_spec.rb
```diff
@@ -1,20 +1,42 @@
 describe FactoryBot::NullObject do
-  let(:methods_to_respond_to)     { %w[id age name admin?] }
-  let(:methods_to_not_respond_to) { %w[email date_of_birth title] }
+  it "responds to the given methods" do
```
sallyhall: It looks like there are three identical blocks here.

richiethomas: Good catch.  Fixed, thanks.

PR #1350, opened 2019-10-18
Refactor sequence_spec.rb to conform to Let's Not style

richiethomas:
For reference:

### DIFF
--- a/spec/factory_bot/sequence_spec.rb
+++ b/spec/factory_bot/sequence_spec.rb
@@ -45,7 +45,12 @@ describe FactoryBot::Sequence do
     end
 
     it "has the expected names as its names" do
-      expect(subject.names).to eq [:test, :alias, :other]
+      names = [:foo, :bar, :baz]
+      sequence = FactoryBot::Sequence.new(names.first, aliases: names.last(2)) do
+        "=#{n}"
+      end
+
+      expect(sequence.names).to eq names
     end
 
     it_behaves_like "a sequence", first_value: "=1", second_value: "=2"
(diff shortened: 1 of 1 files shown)


richiethomas:
This file makes heavy use of shared examples, which (according to our blog post) is one of the few cases where syntax like `subject` makes sense.  I did manage to get rid of some of the `its` calls, fwiw.

spec/factory_bot/sequence_spec.rb
```diff
@@ -2,20 +2,21 @@
...
+  it "has a next value equal to its first value" do
+    expect(subject.next).to eq first_value
+  end
+
+  it "has a next value equal to the 2nd value after being incremented" do
+    subject.next
 
-  context "when incrementing" do
-    before     { subject.next }
-    its(:next) { should eq second_value }
```
composerinteralia: That was fairly confusing before. Nice improvement.

spec/factory_bot/sequence_spec.rb
```diff
@@ -43,7 +44,9 @@
       end
     end
 
-    its(:names) { should eq [:test, :alias, :other] }
+    it "has the expected names as its names" do
+      expect(subject.names).to eq [:test, :alias, :other]
```
composerinteralia: The names are still a bit of a mystery guest (GitHub wasn't showing me thing lines directly above this at first, so I wasn't immediately sure where they came from). If you want to address that in this PR, you could have the subject for the shared example but then avoid using it for this particular spec.

spec/factory_bot/sequence_spec.rb
```diff
@@ -55,7 +58,9 @@
...
```
composerinteralia: Same here

richiethomas: Fixed

spec/factory_bot/sequence_spec.rb
richiethomas: Good catch, fixed in latest commit.

PR #1353, opened 2019-10-31
Add a block about advanced association overrides

jume-dev:
As I took ages today to find out how you can transfer attributes from a factory to an association, I think it is pretty good to add that feature to the docs.


composerinteralia:
Thanks! I am going to hold off merging for now because there is still some conversation about the best way to expose this method (it is not yet part of the officially documented public API) happening in https://github.com/thoughtbot/factory_bot/pull/1300 and https://github.com/thoughtbot/factory_bot/issues/1268. Once that is sorted out, I plan to merge this.

vindvaki:
@composerinteralia is there any chance the method will get renamed or made otherwise incompatible with this example? If not, then do we need to wait for #1300 and #1268?

composerinteralia:
We are going to keep the method as-is, so we can merge this now. I merged 'master' into this branch to get it up to date and had to resolve some merge conflicts. I moved a couple things around and expanded it a little bit while I was in there. Thank you for this!

PR #1361, opened 2019-12-28
Use already installed version of bundler

composerinteralia:
These before install steps were introduced quite a while back in
850116da, 3c0547fa, and b3509f81. It looks like they were introduced as
workarounds to no-longer-relevant issues. Now that bundler is included
in Ruby by default, this is causing builds to fail while waiting on a
prompt about where it is OK to overwrite the "bundle" executable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,9 +6,7 @@ rvm:
   - jruby-9.2.8.0
   - ruby-head
 before_install:
-  - gem update --system
-  - gem update bundler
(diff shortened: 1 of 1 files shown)


composerinteralia:
This doesn't quite work. jruby seems to need the gem installed bundler.

Based on https://github.com/rubygems/rubygems/pull/3040, I think this may sort itself out on its own, so I am going to close this for now.

Issue #903, opened 2016-04-15
Remove ActiveSupport dependency?


cllns:
Here are the ActiveSupport dependencies in FactoryGirl:
```
In lib/factory_girl/evaluator.rb
1:require 'active_support/core_ext/hash/except'
2:require 'active_support/core_ext/class/attribute'

In lib/factory_girl/factory.rb
1:require 'active_support/core_ext/hash/keys'
2:require 'active_support/inflector'

In lib/factory_girl.rb
2:require 'active_support/core_ext/module/delegation'
3:require 'active_support/deprecation'
4:require 'active_support/notifications'

In spec/support/macros/deprecation.rb
1:require "active_support"
```

For Rails apps, using ActiveSupport is a no-brainer since it's already required. 
For other Ruby apps, ActiveSupport is a pretty big dependency. (Granted, only for testing, but it's still there)

I can work to remove the dependency on ActiveSupport. Is that a worthwhile effort? Are you interested in a PR that does that?

Things like ActiveSupport::Notifications can still be supported, by just checking if they're defined at the top of the file.

(The concrete use-case for this is several of us in the [Hanami](https://github.com/hanami/hanami) Gitter chat would like to use FactoryGirl but don't want to add ActiveSupport, since it monkey-patches)

Thanks!

joshuaclayton:
@cllns thanks for opening this issue.

For a while, we'd avoided including AS because of the same reasons you'd mentioned. We've obviously  become more lax on this, and I think you bring up a good point about assessing whether it's a necessary dependency.

Looking through the source for some of those files, I don't think it makes sense right now to introduce similar behavior while avoiding AS - that's a significant amount of code that has the risk of being outdated, buggy, or otherwise needs maintenance over time. The overhead cost of doing that versus leveraging an existing gem is very high.

Were there a community-adopted alternative that didn't rely on reopening classes, it makes sense to *consider* introducing it given it was actively-maintained and widely-used, across all the versions of Ruby we support.

cllns:
Thanks for the quick and thoughtful response @joshuaclayton. 

[Hanami::Utils](https://github.com/hanami/utils) offers some of the functionality that's needed (like inflectors), but not all of it. Also, it only supports MRI 2.2, MRI 2.3 and JRuby.

joshuaclayton:
[...]

PR #1362, opened 2019-12-28
Prepare for factory_bot 6

### LINKED ISSUE #1336: factory_bot 6 checklist
- [x] Remove Rails 4.2 from test matrix
- [x] Remove EOL ruby versions from test matrix
- [x] Remove Rails 4.2 logic from features/support/factories.rb
- [x] Remove deprecated methods from lib/factory.rb
- [x] Bump FactoryBot::Deprecation to 7.0

composerinteralia:
Closes #1336

- removes all the deprecated methods
- removes Ruby 2.3, 2.4 and Rails 4.2 from travis
- `bundle update`s the test gemfiles
- Removes some pre-5.0 logic from a test helper
- Targets Ruby 2.5 with rubocop and fixes violations

We could also remove support for Rails 5.0 and 5.1, which are now EOL, but I don't see a strong reason to do that. We don't seem to have to do anything special to support those versions.


composerinteralia:
I expect CI to fail because of https://github.com/rubygems/rubygems/pull/3040

composerinteralia:
That’s definitely possible. Do you know if there is anything we need to do in particular besides add 2.7 to Travis? Are we getting keyword argument warnings or anything like that?

composerinteralia:
True. I was thinking about leaving those in for now because there doesn't seem to be any additional maintenance required to do so.

nijikon:
@composerinteralia would you consider dropping `activesupport` dependency in version 6?

composerinteralia:
@nijikon this was discussed in https://github.com/thoughtbot/factory_bot/issues/903. If things have changed significantly since then, we can reopen that issue to discuss again.

PR #1364, opened 2020-01-06
fixes deprecation warning on linter.rb

RicardoTrindade:
With ruby 2.7 we start seeing the following warnings when running tests
```
/factory_bot/lib/factory_bot.rb:74: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/factory_bot/lib/factory_bot/linter.rb:3: warning: The called method `initialize' is defined here
```
I'm proposing the following changes so that the message is supressed.


composerinteralia:
Looks good. Thanks!

PR #1379, opened 2020-04-10
Raise error on undefined callback

composerinteralia:
While reading through some code I noticed that `DEFAULT_CALLBACKS` had a
duplicate `:after_create` and was missing `:before_create`. This was
changed in [f82e40c8], but it didn't cause any test failures. It has
also been release for a while now and we didn't see any bug reports.
Hm...

[f82e40c8]: https://github.com/thoughtbot/factory_bot/commit/f82e40c8c50c538c0fba0362c0f6b7f6f96f9a3c

It turns out we have always had the line:
`FactoryBot::Internal.register_callback(name)`,
which defines callbacks on the fly whenever you refer to them in your
factories.

We have a [callback unit test] to ensure the `Callback` class raises an
error on initialize for invalid callback names, but nobody will have
ever seen this error since our code has been registering the callback on
the fly immediately before initializing the `Callback`.

[callback unit test]: https://github.com/thoughtbot/factory_bot/blob/0bcb6151b884d421ef4e7d840c1b098e6ca74d97/spec/factory_bot/callback_spec.rb#L35-L38

I wrote an acceptance test to capture what I think the original intended
behavior was, and that spec led me to remove the questionable on-the-fly
callback registration.

Removing that gave me the failing test for `DEFAULT_CALLBACKS` that I
had been expecting.

I also fixed a couple of other tests that were using custom callbacks
without registering them.

NOTE: This is a breaking change, so I plan to release this as part of FactoryBot 6.

### DIFF
--- a/lib/factory_bot/definition.rb
+++ b/lib/factory_bot/definition.rb
@@ -95,7 +95,6 @@ module FactoryBot
 
     def callback(*names, &block)
       names.each do |name|
-        FactoryBot::Internal.register_callback(name)
         add_callback(Callback.new(name, block))
       end
     end
--- a/lib/factory_bot/internal.rb
+++ b/lib/factory_bot/internal.rb
@@ -10,7 +10,7 @@ module FactoryBot
     }.freeze
 
     DEFAULT_CALLBACKS = [
-      :after_create, :after_build, :after_stub, :after_create
+      :after_create, :after_build, :after_stub, :before_create
     ].freeze
 
     class << self
--- a/spec/acceptance/callbacks_spec.rb
+++ b/spec/acceptance/callbacks_spec.rb
@@ -142,6 +142,10 @@ describe "custom callbacks" do
     FactoryBot.register_strategy(:custom_after, custom_after)
     FactoryBot.register_strategy(:totally_custom, totally_custom)
 
+    FactoryBot.register_callback(:before_custom)
+    FactoryBot.register_callback(:after_custom)
+    FactoryBot.register_callback(:totally_custom)
+
     FactoryBot.define do
       factory :user do
         first_name { "John" }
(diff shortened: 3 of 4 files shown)


lib/factory_bot/definition.rb
```diff
@@ -95,7 +95,6 @@ def after(*names, &block)
 
     def callback(*names, &block)
       names.each do |name|
-        FactoryBot::Internal.register_callback(name)
```
aledustet, line 98: Oh I see this was also duplicated

composerinteralia:
@aledustet thanks for the review!

 I hadn't thought about the fact that the default callbacks use `FactoryBot::Internal.register_callback` and whereas any custom ones would use `FactoryBot.register_callback` (which then delegates to `FactoryBot::Internal.register_callback`).

I don't think there should be any difference between the default callbacks and any custom ones. I would say the same about custom strategies. Treating them differently would add additional complexity, and I don't think would be worth it.

This leaves me wondering whether we should be using the public methods to register all the defaults, rather than skipping the delegation straight to the internal methods. At the moment it shouldn't make any difference, but I don't know if that will always be the case.

As you said, this is probably beyond the scope of this diff, so we can revisit that separately.

composerinteralia:
Thinking about this a bit more, I may go in another direction. The only reason to register_callbacks was to be able to raise `FactoryBot::InvalidCallbackNameError` for unknown callbacks. Given the current code, nobody will have ever seen that error (although if they have a callback around that factory_bot doesn't recognize it will be a no-op).

Introducing this error now could break people's code, and I am not sure it is worth it. Instead of introducing this error, another option would be to delete the `register_callback` method and the `FactoryBot::InvalidCallbackNameError` entirely. This would preserve the current behavior and would involve less code.

composerinteralia:
Closing this in favor of https://github.com/thoughtbot/factory_bot/pull/1390, unless I get strong feedback otherwise.

PR #1380, opened 2020-04-10
Add functionality for enum traits

### LINKED ISSUE #1049: Feature question: Boilerplate for enum traits

composerinteralia:
## Enum traits

This PR adds additional test coverage and finalizes the work started in #1301.

Given a Rails model with an enum attribute:

```rb
class Task < ActiveRecord::Base
  enum status: {queued: 0, started: 1, finished: 2}
end
```

It is common for people to define traits for each possible value of the enum:

```rb
FactoryBot.define do
  factory :task do
    trait :queued do
      status { :queued }
    end

    trait :started do
      status { :started }
    end

    trait :finished do
      status { :finished }
    end
  end
end
```

With this commit, those trait definitions are no longer necessary—they are defined automatically by factory_bot.

If automatically defining traits for enum attributes on every factory is not desired, it is possible to disable the feature by setting `FactoryBot.automatically_define_enum_traits = false` (see commit:  [Allow opting out of automatically defining traits](https://github.com/thoughtbot/factory_bot/pull/1380/commits/5a20017351b08ce2ec9918d799e187e9eaa3ec32)).

In that case, it is still possible to explicitly define traits for an enum attribute in a particular factory:

```rb
FactoryBot.automatically_define_enum_traits = false

FactoryBot.define do
  factory :task do
    traits_for_enum(:status)
  end
end
```

It is also possible to use this feature for other enumerable values, not specifically tied to ActiveRecord enum attributes:

```rb
class Task
  attr_accessor :status
end

FactoryBot.define do
  factory :task do
    traits_for_enum(:status, ["queued", "started", "finished"])
  end
end
```

The second argument here can be an enumerable object, including a Hash or Array.

Closes #1049


lib/factory_bot/definition_proxy.rb
```diff
@@ -176,6 +176,10 @@ def trait(name, &block)
       @definition.define_trait(Trait.new(name, &block))
     end
 
+    def traits_for_enum(attribute_name, values = nil)
```
composerinteralia, line 179: TODO: Add documentation for this method

lib/factory_bot/factory.rb
```diff
@@ -84,6 +84,7 @@ def compile
       unless @compiled
         parent.compile
         parent.defined_traits.each { |trait| define_trait(trait) }
+        @definition.expand_enum_traits(build_class)
         @definition.compile
```
composerinteralia, line 88: In an ideal world we could pass the `build_class` to `@definition.compile` and let the definition itself handle expanding the traits, but we call `@definition.compile` in other places where a `build_class` is not available.

lib/factory_bot/definition_proxy.rb
fridaland: @composerinteralia Added some documentation, feel free to give me some feedback if you see anything that can be improved!

lib/factory_bot/definition_proxy.rb
```diff
@@ -176,6 +176,48 @@ def trait(name, &block)
...
+    # Allows for creation of traits for enumerable values. The instance can be
+    # built with the following allowed syntax.
+    # ----
+    # Default Example:
+    #   factory :task do
+    #     traits_for_enum :status
+    #   end
+    #
+    # Arguments:
+    # * name_of_enum_type_attribute: +Symbol+
```
composerinteralia, line 188: I don't think we use this name anywhere else. I think we should stick to the argument names we use in the method signature itself.

lib/factory_bot/definition_proxy.rb
```diff
@@ -176,6 +176,48 @@ def trait(name, &block)
...
+    # * Where the hash has enum values
+    #   status: { queued: 0, started: 1, finished: 2 }
+    #
+    # ----
+    # Example using a list of values as an Array, or custom enum:
+    #   factory :task do
+    #     traits_for_enum :status, statuses
+    #   end
+    #
+    # Arguments:
```
composerinteralia, line 200: I am finding it a little confusing to have multiple argument lists. Elsewhere I think we have consolidated them into one.

I am thinking:

```
# Arguments:
#   attribute_name: +Symbol+ or +String+ 
#     the name of the attribute these traits will set the value of
#   values: +Array+, +Hash+, or other +Enumerable+
#     An array of trait names, or a mapping of trait names to values for those traits. When this argument is not
#     provided, factory_bot will attempt to get the values from the class of the object being built by calling the
#     pluralized `attribute_name` class method.

File: lib/factory_bot/definition_proxy.rb [hunk 5]
```diff
@@ -176,6 +176,48 @@ def trait(name, &block)
...
+    # * name_of_enum_type_attribute: +Symbol+
+    #   The name of this attribute.
+    #
+    # * argument given: +Array+ or +CustomEnumClass+
+    #
+    # * Where the Array has enum values
+    #   statuses = %w[queued started finished]
+    #
+    # * Where the CustomEnumClass has enum values
```
composerinteralia, line 209: I don't think we need this. This is teaching people how to build custom enumerables, which is beyond the scope of this documentation.

composerinteralia:
@fridaland I pushed up a commit to move some things around in the documentation based on my comments above. Let me know what you think!

composerinteralia:
I still need to work on adding some documentation to the GETTING_STARTED guided, but I was waiting until https://github.com/thoughtbot/factory_bot/pull/1381 gets reviewed.

lib/factory_bot/definition_proxy.rb
fridaland: Fair point!

lib/factory_bot/definition_proxy.rb
fridaland: It makes sense to consolidate, I was questioning whether or not having multiple argument lists would be valuable.

fridaland:
@composerinteralia updates look good. Thanks for clarifying the documentation!

lib/factory_bot/factory.rb
eLod: i agree with placing the triggering logic in `@definition.compile` would be better, but i also agree that it may not worth it to refactor other parts largely only for this.

spec/acceptance/enum_traits_spec.rb
```diff
@@ -0,0 +1,161 @@
...
+        factory :task
+      end
+
+      Task.statuses.each_key do |trait_name|
+        task = FactoryBot.build(:task, trait_name)
+
+        expect(task.status).to eq(trait_name)
+      end
+
+      Task.reset_column_information
```
eLod, line 18: these may be better in an after hook so it's more readable (obviously not high prio)

spec/acceptance/enum_traits_spec.rb
```diff
@@ -0,0 +1,161 @@
+describe "enum traits" do
+  context "when automatically_define_enum_traits is true" do
+    it "builds traits automatically for model enum field" do
+      define_model("Task", status: :integer) do
+        enum status: { queued: 0, started: 1, finished: 2 }
```
eLod, line 5: these are somewhat also redundant, so they are used in every `it` the same or just the keys as a string array, it may make it more readable to do a single `let` for them for the whole `describe` block (again obviously not high prio)

spec/acceptance/enum_traits_spec.rb
```diff
@@ -0,0 +1,161 @@
...
+      end
+
+      Task.statuses.each_key do |trait_name|
+        task = FactoryBot.build(:task, trait_name)
+
+        expect(task.status).to eq("finished")
+      end
+
+      Task.reset_column_information
+    end
```
eLod, line 49: to some extent the examples could be refactored to use a shared example, but i am not sure at what point it is just "clever" overengineering vs making it more readable. generally all the tests are doing

- define the "same" model with integer or string column, optionally adding an (AR) enum
- define the factory with or without explicit traits, obviously this is the main part, not strictly the subject of the tests but close (so the less code share available)
- assert the factory's traits building with the right attribute value

i would argue the first and last steps are very easy to share between the examples (but again, at some point it will be counter-productive), obviously low prio & more debatable

lib/factory_bot/factory.rb
composerinteralia: I pushed up 9c4d7c5, which solves this but adds an extra conditional to `Definition#compile`. Do you think that is worth it?

spec/acceptance/enum_traits_spec.rb
composerinteralia: This made me realize that not all of these tests need models, so I changed some in 1f70160 to use plain old Ruby objects instead. It no longer makes sense to call `Task.reset_column_information` in an after hook since `Task` won't have that method in every test.

spec/acceptance/enum_traits_spec.rb
composerinteralia: I am generally not a fan of `let` for the reasons outlined in this post: https://thoughtbot.com/blog/lets-not.

spec/acceptance/enum_traits_spec.rb
composerinteralia: There might be some room to pull out some shared methods here, but lately I have been preferring repetition in my tests over clever shared behavior. I think https://thoughtbot.com/blog/the-case-for-wet-tests covers a lot of what I am thinking. I also talk a little bit about my preference for duplication in tests in https://www.bikeshed.fm/186.

composerinteralia:
@eLod thanks so much for taking a look. I left some comments on your comments and pushed up two additional commits.

PR #1383, opened 2020-04-20
Change string to User class reference in Getting Started doc

psnjs:
`User` was quoted in the example that presumably shows how to refer to a class directly, which makes the subsequent usage example:

> you can also pass a symbol or string,

confusing.


composerinteralia:
Good call. Thanks!

PR #1384, opened 2020-04-21
Allow list methods blocks to receive index

emmanuelgautier1994:
Picking up from https://github.com/thoughtbot/factory_bot/pull/1314 to get this to master!

I've kept @mcountis's commits, addressed the pending comments, and added some documentation.

I haven't been able to run the acceptance tests on my machine because of conflicts with my system gems, but I'd be happy to have another try if spec changes are needed.

Will be more than happy to address anything else I've missed.


GETTING_STARTED.md
```diff
@@ -1018,6 +1018,14 @@ To set the attributes for each of the factories, you can pass in a hash as you n
 twenty_year_olds = build_list(:user, 25, date_of_birth: 20.years.ago)
 ```
 
+There are `*_list_with_index` methods for using the index to set different attribute values for each created item:
+
+```ruby
+twenty_somethings = build_list(:user, 10) do |user, i|
```
composerinteralia, line 1024: The text says `*_list_with_index`, but the code example uses `build_list`. I think I like what it is in the code example a bit better (see the next comment).

lib/factory_bot/strategy_syntax_method_registrar.rb
```diff
@@ -33,6 +33,21 @@ def define_list_strategy_method
       end
     end
 
+    def define_list_with_index_strategy_method
```
composerinteralia, line 36: I think your tests are failing because you aren't calling this anywhere. You would need to call it inside of `define_strategy_methods`.

That said, I am leaning toward extending the existing `*_list` method rather than introducing a new `*_list_with_index` method. We would extend the existing method by doing an arity check to decide whether to pass along the index. This is more in line with what we have done in other place in factory bot (e.g. [passing the evaluator to dynamic attributes](https://github.com/thoughtbot/factory_bot/blob/6d8bea1670ef10daf0f15680288334916e031430/lib/factory_bot/attribute/dynamic.rb#L14-L17), [passing the instance to to_create](https://github.com/thoughtbot/factory_bot/blob/c22c9ab0524a00ede1f09a1c996ed77dfc1afa73/lib/factory_bot/evaluation.rb#L16-L19), or [passing both the instance and the evaluator to callbacks](https://github.com/thoughtbot/factory_bot/blob/f82e40c8c50c538c0fba0362c0f6b7f6f96f9a3c/lib/factory_bot/callback.rb#L12-L16))

I had played around with this in https://github.com/thoughtbot/factory_bot/commit/6996d8287510aa2f415c8ddf5745a1022fdb837d. I don't remember how far I got, but it should give you the general idea of what I was thinking about.

composerinteralia:
Thanks for picking this up!

lib/factory_bot/strategy_syntax_method_registrar.rb
emmanuelgautier1994: Thanks for the feedback! Extending the method sounds great, I gave it a try in c4443cb1d905552a39475e3acf2e26040c58f9c6, it looks like it's working well.

spec/acceptance/create_list_spec.rb
```diff
@@ -52,6 +52,20 @@
     end
   end
 
+  context "with a block that receives both the object and an index" do
+    subject do
+      FactoryBot.create_list(:post, 20, title: "The Indexed Block") do |post, index|
```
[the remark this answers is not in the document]
composerinteralia: Safe to ignore hound here as well

lib/factory_bot/strategy_syntax_method_registrar.rb
```diff
@@ -29,7 +29,10 @@ def define_list_strategy_method
           raise ArgumentError, "count missing for #{strategy_name}_list"
         end
 
-        Array.new(amount) { send(strategy_name, name, *traits_and_overrides, &block) }
+        Array.new(amount) do |i|
+          curried_block = block&.arity == 2 ? ->(record) { block.call(record, i) } : block
```
composerinteralia, line 33: code climate is complaining about the complexity of `define_list_strategy_method`, and I usually take that seriously.

What do you think about changing this to:

```suggestion
          block_with_index = StrategySyntaxMethodRegistrar.with_index(block, i)
```

Where `StrategySyntaxMethodRegistrar.with_index(block, i)` would be defined at the top of the class as:

```rb
    def self.with_index(block, index)
      if block&.arity == 2
        ->(instance) { block.call(instance, index) }
      else
        block
      end
    end
```

Note that I expanded the ternary to match the [thoughtbot style guide](https://github.com/thoughtbot/guides/tree/master/style/ruby). I also renamed `record` to `instance`, which is more in line with our naming elsewhere.

emmanuelgautier1994: Sounds great, works great. Done!

composerinteralia:
Thank you!

I squashed this into one commit with the original author listed as a co-author and merged as 0b53d28. (I did this from the command line instead of the GitHub UI so it will look like this was closed instead of merged, but don't worry—it is indeed merged and you are listed as an author.)

composerinteralia:
Also, this is now release in factory_bot 5.2.0.

emmanuelgautier1994:
Thank you!

PR #1390, opened 2020-05-03
Remove code for registering callbacks

composerinteralia:
Before this commit, factory_bot registered default callbacks in a global
registry, and offered a `register_callback` method for registering any
custom callbacks (which would be fired within custom strategies).

The idea here seems to have been that we could check against this global
registry and then raise an error when referring to a callback that was
not registered.

However, nobody ever saw this error because we have always had the line:
`FactoryBot::Internal.register_callback(name)`, which registers
callbacks on the fly whenever they are referred to. (I noticed this when
a [change to default callbacks] accidentally duplicated `after_create`
instead of including `before_create`, and we didn't get any tests
failures.)

We have two options here:

1. Preserve the existing behavior by deleting all the callback
registration code. (That is what this commit does)

2. Change to the code try and capture what I assume was the original
intention. This would be a breaking change, so we would want to
introduce it in a major release. (See #1379)

I prefer preserving the existing behavior because we haven't seen any
issues around this, and making a breaking change for this doesn't seem
worthwhile.

[change to default callbacks]: https://github.com/thoughtbot/factory_bot/commit/f82e40c8c50c538c0fba0362c0f6b7f6f96f9a3c#diff-c6d30be672f880311a7df0820dc4fb21R12-R14

### DIFF
--- a/lib/factory_bot/callback.rb
+++ b/lib/factory_bot/callback.rb
@@ -5,7 +5,6 @@ module FactoryBot
     def initialize(name, block)
       @name  = name.to_sym
       @block = block
-      ensure_valid_callback_name!
     end
 
     def run(instance, evaluator)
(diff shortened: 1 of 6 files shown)


lib/factory_bot/callback.rb
```diff
@@ -27,13 +26,6 @@ def ==(other)
 
     private
 
-    def ensure_valid_callback_name!
-      unless FactoryBot::Internal.callback_names.include?(name)
```
composerinteralia, line 31: Currently this condition will always be `true`

PR #1400, opened 2020-05-31
Remove unused null object code

Insti:
Since `NullObject` inherits from `BasicObject` rather than `Object` it does not need to implement `respond_to?` or `respond_to_missing?` in the way that Rubocop suggests (which assumes that `Object` is the base class).

Rather than writing unused code just to make Rubocop happy, we can tell Rubocop to ignore the `Style/MissingRespondToMissing` cop and delete the useless code.

This improves the test coverage for the project:

Before:
`1140 / 1143 LOC (99.74%)`

After:
`1139 / 1141 LOC (99.82%)`

### DIFF
--- a/lib/factory_bot/null_object.rb
+++ b/lib/factory_bot/null_object.rb
@@ -13,7 +13,7 @@ module FactoryBot
       end
     end
 
-    def respond_to?(method, _include_private = false)
+    def respond_to?(method)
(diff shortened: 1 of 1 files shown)


composerinteralia:
This makes sense to me. `#response_to_missing?` normally makes `#method` work properly, but `BasicObject#method` is not a thing. Thanks.

PR #1403, opened 2020-06-04
Add randomise_sequence_start option

CGA1123:
This change adds the `FactoryBot.randomise_sequence_start` flag, which,
when set to `true` will make default integer sequences start with a
random number rather than always starting at `1`.

We recently had an incorrect spec slip by reviews and cause our test
suite to be flaky as it passed only if that spec file was run before any
other using the specified factory, as the sequences were in sync.

Randomising the sequence ids might help to catch subtle errors like
this.

This is probably the simplest representation of how randomising would
catch subtle errors earlier, the error is in which value is being used in the
expectation `id` vs `publisher_id`.
```ruby
    # frozen_string_literal: true

    require "bundler/inline"

    gemfile(true) do
      source "https://rubygems.org"

      gem "activerecord", "6.0.3"
      gem "sqlite3"
      gem "factory_bot"
      gem "rspec"
    end

    require "active_record"
    require "rspec/autorun"
    require "logger"

    ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")

      create_table :authors, force: true do |t|
        t.timestamps
      end

      create_table :books, force: true do |t|
        t.integer :author_id
        t.integer :publisher_id
        t.timestamps
      end
    end

    class Author < ActiveRecord::Base
      has_many :books

      def publisher_ids
        books.order(created_at: :desc).map(&:publisher_id)
      end
    end

    class Book < ActiveRecord::Base
      belongs_to :author
    end

    FactoryBot.define do
      factory :author do
      end

      factory :book do
        association :author

        sequence :publisher_id
      end
    end

    RSpec.describe Author do
      # Uncomment to simulate another spec file using the factory and updating
      # the sequence!
      #
      # before { FactoryBot.build(:book) }

      let!(:author) { FactoryBot.create(:author) }
      let!(:old_book) { FactoryBot.create(:book, author: author, created_at: 1.month.ago) }
      let!(:older_book) { FactoryBot.create(:book, author: author, created_at: 2.month.ago) }
      let!(:recent_book) { FactoryBot.create(:book, author: author, created_at: 1.day.ago) }

      describe '#publisher_ids' do
        it { expect(author.publisher_ids).to eq([recent_book.id, old_book.id, older_book.id]) }
      end
    end
```

### DIFF
--- a/lib/factory_bot.rb
+++ b/lib/factory_bot.rb
@@ -51,6 +51,9 @@ require "factory_bot/version"
 module FactoryBot
   Deprecation = ActiveSupport::Deprecation.new("7.0", "factory_bot")
 
+  mattr_accessor :randomise_sequence_start, instance_accessor: false
+  self.randomise_sequence_start = false
+
   mattr_accessor :use_parent_strategy, instance_accessor: false
   self.use_parent_strategy = true
 
--- a/lib/factory_bot/sequence.rb
+++ b/lib/factory_bot/sequence.rb
@@ -10,7 +10,7 @@ module FactoryBot
       @proc    = proc
 
       options  = args.extract_options!
-      @value   = args.first || 1
+      @value   = args.first || default_value
       @aliases = options.fetch(:aliases) { [] }
 
       if !@value.respond_to?(:peek)
@@ -39,6 +39,12 @@ module FactoryBot
     end
 
     private
+
+    def default_value
+      offset = FactoryBot.randomise_sequence_start ? Random.rand(1_000) : 0
+
+      offset + 1
+    end
 
     def value
       @value.peek
--- a/spec/acceptance/sequence_spec.rb
+++ b/spec/acceptance/sequence_spec.rb
@@ -69,4 +69,24 @@ describe "sequences" do
     expect(values.first).to eq("somebody1@example.com")
     expect(values.second).to eq("somebody2@example.com")
   end
+
+  context "when the :randomise_sequence_start config option is set to true" do
+    before { allow(Random).to receive(:rand).and_return(123) }
+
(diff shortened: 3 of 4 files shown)


composerinteralia:
I am unable to run the script you provided. I am seeing:

```
An error occurred while loading spec_helper. - Did you mean?
                    rspec ./spec/spec_helper.rb

Failure/Error: require "rails"

LoadError:
  cannot load such file -- rails
# ./lib/factory_bot_rails/railtie.rb:6:in `<top (required)>'
# ./lib/factory_bot_rails.rb:1:in `<top (required)>'
# ./spec/spec_helper.rb:5:in `<top (required)>'
```

Any advice?

composerinteralia:
Thank you for this PR, and for providing a reproduction script. That helped me understand the problem you are trying to solve with this change.

I am leaning toward closing this instead of merging it. The code itself looks great, but I am hesitant to provide a way to introduce randomness into factories. While it may catch the occasional false positive, it can also introduce subtle bugs in the test suite that become really difficult to debug (the `--seed` flag would no longer produce reliable behavior, for example, and that would mean that `rspec bisect` would no longer work).

CGA1123:
Hey @composerinteralia, that sounds reasonable 😄  

I still think that randomness would be nice, relying on specific sequence start numbers feels like a recipe for brittleness!

But, without a nice way to seed that randomness, I think you are right in saying it may cause more headaches than it solves.

I may look into inheriting seeds from RSpec/Minitest and introduce predictable randomness into sequences as a separate gem, I think it is probably not good to add that coupling to factory_bot itself.

Thanks for taking the time to look at this!

PR #1415, opened 2020-06-23
Fix ruby 2.7 kwargs warning

HParker:
Ruby 2.7 deprecated passing kwargs when the method expects a hash or passing a hash when the method expects kwargs. In factory_bot, this creates the warning:

```
/Users/hparker/code/factory_bot/lib/factory_bot/decorator/new_constructor.rb:9: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/Users/hparker/code/factory_bot/spec/acceptance/initialize_with_spec.rb:220: warning: The called method `initialize' is defined here
```

We can fix this warning by updating the syntax. We need to include `**kwargs` in the `method_missing` calls when we are on ruby 2.7 or later.
In decorator.rb, adding `**kwargs` alone doesn't work since adding `**kwargs` can change what arguments remain in the `args`.

In this case we have to class eval the method if we are running ruby 2.7. This way the syntax is valid in previous versions and we can use the `...` operator which allows us to avoid changing the arguments passed on in method missing.

### DIFF
--- a/lib/factory_bot/decorator.rb
+++ b/lib/factory_bot/decorator.rb
@@ -6,16 +6,28 @@ module FactoryBot
       @component = component
     end
 
-    def method_missing(name, *args, &block) # rubocop:disable Style/MethodMissingSuper
-      @component.send(name, *args, &block)
+    if ::Gem::Version.new(::RUBY_VERSION) >= ::Gem::Version.new("2.7")
+      class_eval(<<~RUBY, __FILE__, __LINE__ + 1)
+          def method_missing(...) # rubocop:disable Style/MethodMissingSuper, Style/MissingRespondToMissing
+          @component.send(...)
+        end
+
+        def send(...)
+          __send__(...)
+        end
+      RUBY
+    else
+      def method_missing(name, *args, &block) # rubocop:disable Style/MethodMissingSuper, Style/MissingRespondToMissing
+        @component.send(name, *args, &block)
+      end
+
+      def send(symbol, *args, &block)
+        __send__(symbol, *args, &block)
+      end
     end
 
     def respond_to_missing?(name, include_private = false)
       @component.respond_to?(name, true) || super
-    end
-
-    def send(symbol, *args, &block)
-      __send__(symbol, *args, &block)
     end
 
     def self.const_missing(name)
--- a/lib/factory_bot/decorator/invocation_tracker.rb
+++ b/lib/factory_bot/decorator/invocation_tracker.rb
@@ -6,9 +6,16 @@ module FactoryBot
         @invoked_methods = []
       end
 
-      def method_missing(name, *args, &block) # rubocop:disable Style/MissingRespondToMissing
-        @invoked_methods << name
-        super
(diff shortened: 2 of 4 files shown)


composerinteralia:
Thanks for opening this PR!

Could explain a bit more about what code is causing the deprecation warnings? I ask because I am seeing **more** factory_bot-specific warnings on the 2.7 Travis builds with this change than I did before this change. My expectation would be that for each change made in this PR there would be at least one test that outputs a warning before the change, and no longer outputs the warning after this change.

Here is what I am currently seeing:

* [Before: Travis 2.7 build outputs one warning coming from factory_bot](https://travis-ci.org/github/thoughtbot/factory_bot/jobs/701083954)
* [After: Travis 2.7 build outputs three warnings coming from factory_bot](https://travis-ci.org/github/thoughtbot/factory_bot/jobs/701371499)

(I searched the logs from each build for "/home/travis/build/thoughtbot/factory_bot/lib/")

lib/factory_bot/decorator.rb
```diff
@@ -6,18 +6,30 @@ def initialize(component)
       @component = component
     end
 
-    def method_missing(name, *args, &block) # rubocop:disable Style/MethodMissingSuper
-      @component.send(name, *args, &block)
+    if ::Gem::Version.new(::RUBY_VERSION) >= ::Gem::Version.new("2.7")
+      class_eval(<<~RUBY, __FILE__, __LINE__ + 1)
+          def method_missing(...) # rubocop:disable Style/MethodMissingSuper, Style/MissingRespondToMissing
```
composerinteralia, line 11: We switched over from Rubocop to [standard](https://github.com/testdouble/standard), so feel free to remove these Rubocop directives.

HParker: Standard does seem to still use the `rubocop:disable` comments since removing these (maybe not this one since it is `class_eval`ed) causes the standard test when I run `bundle exec rake` to fail.

HParker:
Hey @composerinteralia thanks for the comments!

I do see,

```
/home/travis/build/thoughtbot/factory_bot/lib/factory_bot/decorator/new_constructor.rb:9: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
/home/travis/build/thoughtbot/factory_bot/spec/acceptance/initialize_with_spec.rb:220: warning: The called method `initialize' is defined here
```

on the 6.0 build in CI, but it does not appear locally for me. It looks like the warning is not happening in the rails master build which makes me think that it is rails 6.0's `delegate` method that is causing the warning in that build.

- rails master build before: https://travis-ci.org/github/thoughtbot/factory_bot/jobs/701083964
- rails master build after: https://travis-ci.org/github/thoughtbot/factory_bot/jobs/701717267

composerinteralia:
Thanks again! I don't like this PR and I am grateful for you opening it, if that makes sense 😄.

HParker:
@composerinteralia agreed! I really wish there was another way to do this. Thanks again for factory_bot! ❤️

PR #1416, opened 2020-06-24
Update rails versions used in test

HParker:
Some of the deprecation warnings in test coming from rails should be resolved by updating the rails versions.

This maintains the test versions, but updates them to the latest release in that version.

Generated by running `bundle exec appraisal update`


HParker:
With this update the rails 6 + ruby 2.7 build only has the one warning shown fixed in: https://github.com/thoughtbot/factory_bot/pull/1415

composerinteralia:
Nice! Thank you!!!

PR #1421, opened 2020-07-09
Add definition names to default trait key errors

### LINKED ISSUE #1222: Include factory name in "Trait not registered" error message
I am not sure how feasible this is, since the global registry doesn't know anything about a specific factory. But it is worth thinking about, and it could help with issues like https://github.com/thoughtbot/factory_bot_rails/issues/249

composerinteralia:
Closes #1222

Before this commit, referencing a trait that didn't exist would raise a
generic `KeyError: Trait not registered: "trait_name"`. This can
sometime make it difficult to know where exactly the error is coming
from. The fact that implicitly declared associations and sequences will
fall back to implicit traits if they can't be found compounds this
problem. If various associations, sequences, or traits share the same
name, the hunt for the error can take some time.

With this commit we include the factory or trait name (i.e. the
definition name) in the error message to help identify where the
problematic trait originated.

Because trait lookup relies on a more generic class that raises a
`KeyError` for missing keys, we rescue that error, then construct a new
error with our custom message using the original error's message and
backtrace.


spec/acceptance/traits_spec.rb
```diff
@@ -293,15 +280,63 @@ def build_user_factory_with_admin_trait(trait_name)
...
+
+        factory :some_other_factory do
+          trait :inaccessible_trait
+        end
+      end
+
+      expect { FactoryBot.build(:user) }.to raise_error(
+        KeyError,
+        'Trait not registered: "inaccessible_trait" ' \
+        '(referenced within "user" definition)'
```
seanpdoyle, line 313: Was this multi-line string formatted by Standard?

Would a HEREDOC be a viable alternative?

lib/factory_bot/definition.rb
```diff
@@ -111,6 +111,15 @@ def callback(*names, &block)
 
     def base_traits
       @base_traits.map { |name| trait_by_name(name) }
+    rescue KeyError => error
+      raise append_definition_name_to_error(error)
+    end
+
+    def append_definition_name_to_error(error)
+      message = "#{error.message} (referenced within \"#{name}\" definition)"
+      error.class.new(message).tap do |new_error|
```
seanpdoyle, line 120: This is a clever and effective way to re-raise the underlying `KeyError`.

Is there a downside to raising an error class that FactoryBot declares, instead of `KeyError` from Ruby's standard library?

das3in: Sean beat me to this. I don't think it's a bad idea necessarily to use a `KeyError` but a FactoryBot error class would be sweet considering this seems like a FactoryBot specific error

seanpdoyle: Do we consider the originally raised `KeyError` part of a factory's public interface? Is re-raising the `KeyError` motivated by backwards compatibility?

composerinteralia: The reason we are using a `KeyError` rather than a custom FactoryBot error class is because `KeyError`s come with `Did you mean?` suggestions (see b86ec13).

But you are reminding me that I should probably add a test around that case. The `Did you mean?` message is meant to come at the end, so I might have to insert the custom bit here into the middle somewhere.

spec/acceptance/traits_spec.rb
composerinteralia: Is this what you were thinking?

```rb
        <<~MSG.strip
          Trait not registered: "inaccessible_trait" (referenced within "user" definition)
        MSG
```

At that point I would prefer:

```rb
        'Trait not registered: "inaccessible_trait" (referenced within "user" definition)'
```

I didn't write it that way because the line was getting a bit long. But I could be convinced that it is better for readability. What do you think?

lib/factory_bot/definition.rb
composerinteralia: OK, I modified this to handle the `Did you mean?` message.

spec/acceptance/traits_spec.rb
composerinteralia: I went with the string all on one line for now, since it is a little bit easier to read.

lib/factory_bot/definition.rb
composerinteralia: As I think about this more, I think we could probably update the `FactoryBot::Registry` to raise a custom factory_bot error instead of a `KeyError`, since by the time we [build the new error](https://github.com/thoughtbot/factory_bot/blob/98790602896bbe93195cface4b448d859dae3721/lib/factory_bot/registry.rb#L42) the `Did you mean?` message is already populated.

das3in: Sure I'll take a stab at that this weekend.

PR #1396, opened 2020-05-15
Document how traits work in factory_bot

entcheva:
Co-authored-by: Daniel Colson <danieljamescolson@gmail.com>
Co-authored-by: Frida Casas <fridacasas.fc@gmail.com>

PR #1428, opened 2020-07-17
Fix inconsistent trait behavior

### LINKED ISSUE #968: Inconsistent and brittle trait inheritance behavior

NoahTheDuke:
NOTE: This is not necessarily how we want to fix #968, but if you ignore the tests that fail due to `to_create` and `instance_with` not working, you'll see that this works otherwise.

To fix the issue of inconsistent "copying" ordering, we need to change how a trait string (an implicit attribute, a trait given to the runner, or a trait given to the factory method) becomes a trait instance. Right now, we copy into each trait instance all of the other trait instances, and we do this for a given factory up the parent chain as they're built (which is how the inconsistency happens).

What we have to do is instead build a [dynamic scope](https://en.wikipedia.org/wiki/Scope_(computer_science)#Dynamic_scoping) system to handle an external trait referencing an internal trait in the build process (as seen in #968).

My changes are as follows:
1) Copy the `attributes` and `aggregate_from_traits_and_self` (and associated methods) into factory and trait, so we can make changes directly to specific places where necessary.
2) Add a "factory" property on traits to store the trait's "parent" factory.
3) Change `trait_for_name` in factory.rb to look for the trait first in its own trait hash (now a hash instead of a set), or look for it in the parent factory's trait hash, or find it in the global trait hash.
4) Change `trait_for_name` in trait.rb to look for the trait in the parent factory, or if global, in the global trait hash.
5) Change `aggregate_from_traits_and_self` in factory.rb to make the factory instance the "parent" of all of the base and additional traits, as they're found. This is the necessary step to make the rest work. Otherwise, traits work on a purely lexical scope level. Once the factory instance is set as the parent, it calls `&:attributes` on them to get their attributes, like normal.

This pretty explicitly messes up `to_create` and `instance_with`, and was built in a javascript environment that relied on inheritance (Factory and Trait as extended Definitions), so I don't know how to solve those. Any suggestions/ideas would be great.

### DIFF
--- a/lib/factory_bot/definition.rb
+++ b/lib/factory_bot/definition.rb
@@ -1,17 +1,17 @@
 module FactoryBot
   # @api private
   class Definition
-    attr_reader :defined_traits, :declarations, :name, :registered_enums
+    attr_reader :traits, :declarations, :name, :registered_enums, :base_traits_, :additional_traits_
 
     def initialize(name, base_traits = [])
       @name = name
       @declarations = DeclarationList.new(name)
       @callbacks = []
-      @defined_traits = Set.new
+      @traits = ActiveSupport::HashWithIndifferentAccess.new
       @registered_enums = []
       @to_create = nil
-      @base_traits = base_traits
-      @additional_traits = []
+      @base_traits_ = base_traits
+      @additional_traits_ = []
       @constructor = nil
       @attributes = nil
       @compiled = false
@@ -33,16 +33,19 @@ module FactoryBot
       if block_given?
         @to_create = block
       else
-        aggregate_from_traits_and_self(:to_create) { @to_create }.last
+        # aggregate_from_traits_and_self(:to_create) { @to_create }.last
+        @to_create
       end
     end
 
     def constructor
-      aggregate_from_traits_and_self(:constructor) { @constructor }.last
+      # aggregate_from_traits_and_self(:constructor) { @constructor }.last
+      @constructor
     end
 
     def callbacks
-      aggregate_from_traits_and_self(:callbacks) { @callbacks }
+      # aggregate_from_traits_and_self(:callbacks) { @callbacks }
+      @callbacks
     end
 
     def compile(klass = nil)
@@ -50,11 +53,6 @@ module FactoryBot
         expand_enum_traits(klass) unless klass.nil?
 
         declarations.attributes
-
-        defined_traits.each do |defined_trait|
-          base_traits.each { |bt| bt.define_trait defined_trait }
-          additional_traits.each { |at| at.define_trait defined_trait }
-        end
 
         @compiled = true
       end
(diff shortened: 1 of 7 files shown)


composerinteralia:
Thanks for exploring this. It might take me a while to look through this in detail, but I am excited to do so. We had started to talk about how traits work in https://github.com/thoughtbot/factory_bot/pull/1396, because we want to change that to solve issues like #968.

composerinteralia:
I haven't forgotten about you, Noah. I am just scared of looking too deeply into how traits work 😆 . Also, I am now on parental leave so it will probably still be a while.

NoahTheDuke:
haha No worries! Take all the time you need, I am patient. I'll see about doing this is in a cleaner way instead of just this half-broken state to make the changes more obvious.

znakaska:
FWIW, this has been annoying me for a really long time. After encountering this issue again yesterday, I discovered this PR and it inspired me to make another attempt at solving this: https://github.com/thoughtbot/factory_bot/compare/master...znakaska:inherited-traits

Specifically, I needed to override part of a parent trait and add some callbacks in a child factory. The [acceptance test](https://github.com/znakaska/factory_bot/blob/inherited-traits/spec/acceptance/trait_inheritance_spec.rb) illustrates it a bit better.

entcheva:
Thanks for your work @NoahTheDuke @znakaska! I'm planning to take a deeper look at both of these approaches this soon and will comment here then.

NoahTheDuke:
If it would be helpful, I can dig up the PR where I fixed this in my javascript adaption of factory_bot, to better help explain the set of changes needed.

entcheva:
@NoahTheDuke That would be great, thanks!

NoahTheDuke:
Here is the PR: [fixture-riveter #1](https://github.com/Batterii/fixture-riveter/pull/1). Please reach out if you need help understanding any of the code changes.

PR #1431, opened 2020-07-29
Introduce transient attributes in the GETTING_STARTED guide.

Shigawire:
This PR provides a very brief introduction to the transient attributes feature. To me (personally) the feature was unclear until I researched it further, so I believe a small introduction would be reasonable to add.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -348,6 +348,7 @@ create(:user, last_name: "Doe").email
 
 Transient Attributes
 --------------------
+Transient attributes behave like virtual attributes without being available on the model itself. This allows for more complex logic inside factories without altering the model for this cause.
 
 ### With other attributes
 
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -348,7 +348,7 @@ create(:user, last_name: "Doe").email
 
 Transient Attributes
 --------------------
-Transient attributes behave like virtual attributes without being available on the model itself. This allows for more complex logic inside factories without altering the model for this cause.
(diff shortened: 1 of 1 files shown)


GETTING_STARTED.md
```diff
@@ -348,6 +348,7 @@ create(:user, last_name: "Doe").email
 
 Transient Attributes
 --------------------
+Transient attributes behave like virtual attributes without being available on the model itself. This allows for more complex logic inside factories without altering the model for this cause.
```
composerinteralia, line 351: I like the idea of adding an intro here, but I am not sure if "virtual attributes" will be clear to everyone reading this. I also would like to avoid using "model" since factory_bot can be used for any type of Ruby object.

What do you think about:

```suggestion
Transient attributes are available within the factory definition, but are not set on the object being built. This allows for more complex logic inside factories.
```

Shigawire: Understood! "model" is probably wrongly worded here. The part "are available within the factory definition, but are not set on the object being built." is intuitively not super clear for me (personally), as it describes the functionality implicitly. Maybe we can find something in between, like "Transient attributes are attributes available within the factory definition only...", making it clear that transient attributes are, in fact, attributes.
However, I do understand this is a superfluous definition of the feature, it would just help me better to understand what's going on.

composerinteralia: So taking your suggestion and combining it with my suggestion, that might look something like:

```suggestion
Transient attributes are attributes only available within the factory definition, and not set on the object being built. This allows for more complex logic inside factories.
```

I am fine with that as well!

composerinteralia:
I went ahead and merged the version that I think captures both of our comments. Thanks for the PR!

PR #1434, opened 2020-08-04
Allow to pass a lambda function to sequence initial value

javierav:
This merge allow us to pass a lambda function as the initial value of sequence. This lambda will be called the first time that is used (lazy load). For example:

```ruby
factory :user do
  sequence(:email, -> { User.count }) { |n| "person#{n}@example.com" }
end
```

### DIFF
--- a/lib/factory_bot/sequence.rb
+++ b/lib/factory_bot/sequence.rb
@@ -55,15 +55,33 @@ module FactoryBot
       end
 
       def peek
-        @value
+        value
       end
 
       def next
-        @value = @value.next
+        @value = value.next
       end
 
       def rewind
-        @value = @first_value
+        @value = first_value
+      end
+
+      private
+
+      def value
+        if @value.respond_to?(:call)
+          @value = @value.call
+        else
+          @value
+        end
+      end
+
+      def first_value
+        if @first_value.respond_to?(:call)
+          @first_value = @first_value.call
+        else
+          @first_value
+        end
       end
     end
   end
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -980,6 +980,14 @@ factory :user do
 end
 ```
 
+You can also pass a lambda function for a lazy evaluation of the initial value.
+
+```ruby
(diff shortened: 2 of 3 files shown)


composerinteralia:
Thanks for the PR. This is an interesting idea. Can you provide some more detail on when this would be useful? For the `-> { User.count }` example you gave, is there a reason you need to start the sequence at that particular value vs. starting either at 1 or, if there are already some user fixtures loaded, at some arbitrary number like 1000?

lib/factory_bot/sequence.rb
```diff
@@ -55,15 +55,33 @@ def initialize(value)
...
+
+      def value
+        if @value.respond_to?(:call)
+          @value = @value.call
+        else
+          @value
+        end
+      end
+
+      def first_value
```
composerinteralia, line 79: Calling the block again when we `rewind` means that rewinding will not bring us back to the value we started at. This sort of goes against why we introduced the feature to rewind sequences - it was for people who were relying on specific ids in their tests for one reason or another.

If we go forward with this feature, I think we will need some additional test coverage around how this works with rewinding sequences.

javierav:
Thanks you @composerinteralia for your comments. I review them ASAP, I not have many time this days....

composerinteralia:
Thanks for the response. ASAP is not necessary at all. This PR isn't going anywhere, so please take your time.

entcheva:
Hi @javierav - are you still interested in working on this, or should we close this PR? If you're still planning on working on this within the next few months, we'll keep it open, but if you no longer have the time to commit we can close it. Thank you!

entcheva:
Hi @javierav - since there hasn't been activity on this PR since August, I'm going to close it. If you find you have time in the future to revisit, please feel free to reopen or open a new PR!

aalvarado:
You can change the existing sequence in the registry itself:

```
FactoryGirl.sequences.find(:email).instance_variable_set(:@value, FactoryGirl::Sequence::EnumeratorAdapter.new(1000))
```

PR #1437, opened 2020-08-17
Use value returned by to_create

TheWudu:
Use value returned by to_create block, also pass this into the following after_create hooks. 

Guess this is a major change, as all tests using to_create have to be adapted if this is merged.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -12,6 +12,11 @@ GEM
     activerecord (6.0.3.1)
       activemodel (= 6.0.3.1)
       activesupport (= 6.0.3.1)
+    activerecord-jdbc-adapter (60.2-java)
+      activerecord (~> 6.0.0)
+    activerecord-jdbcsqlite3-adapter (60.2-java)
+      activerecord-jdbc-adapter (= 60.2)
+      jdbc-sqlite3 (~> 3.8, < 3.30)
     activesupport (6.0.3.1)
       concurrent-ruby (~> 1.0, >= 1.0.2)
       i18n (>= 0.7, < 2)
@@ -54,9 +59,11 @@ GEM
     diff-lcs (1.3)
     docile (1.3.2)
     ffi (1.12.2)
+    ffi (1.12.2-java)
     gherkin (5.1.0)
     i18n (1.8.2)
       concurrent-ruby (~> 1.0)
+    jdbc-sqlite3 (3.28.0)
     minitest (5.14.1)
     multi_json (1.14.1)
     multi_test (0.1.2)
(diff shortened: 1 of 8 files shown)


TheWudu:
Is there a way to retrigger the checks, as they failed because "gem update system" was not possible? ;-)

composerinteralia:
> Guess this is a major change, as all tests using to_create have to be adapted if this is merged.

Yeah, this is not a change I would be comfortable making given the changes to the test suite that were required. This would break `to_create` for everyone.

TheWudu:
Totally understandable. Although it still feels somehow strange to me that the response of to_create (which basically should create the object then), is not considered at all. But i will think of it a bit more.

composerinteralia:
It might be the terminology that is making this seem strange.  `to_create` is involved in persisting the object. It probably could have been called `to_persist` instead, although I am sure there is some history in the name that I am missing. By the time you get into `to_create` the object has already been initialized and its attributes have already been set. At that point I wouldn't expect to create any new objects, but rather to finalize or persist the object that has already been built. The [documentation on `to_create`](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#custom-methods-to-persist-objects) will maybe make this more clear.

ThomasKoppensteiner:
Hey, in my opinion we could introduce a global configuration `use_to_create_result` (or similar) to avoid the sudden breaking change.

1. Step: Add option to switch the result building flow

```ruby
def result(evaluation)
  if USE_TO_CREATE_RESULT # Some global setting, which defaults to `false`. Doesn't need to be a constant.
    modified_result_evalution(evaluation)
  else
    current_result_evalution(evaluation)
  end
end

private

def current_result_evalution(evaluation)
  evaluation.object
  evaluation.notify(:after_build, instance)
  evaluation.notify(:before_create, instance)
  evaluation.create(instance)
  evaluation.notify(:after_create, instance)
end

def modified_result_evalution(evaluation)
  instance = evaluation.object
  evaluation.notify(:after_build, instance)
  evaluation.notify(:before_create, instance)
  instance = evaluation.create(instance)
  evaluation.notify(:after_create, instance)
  instance
end
```

2. Step: Add deprecation warning to default flow and explain how to switch over

```ruby
def result(evaluation)
  if USE_TO_CREATE_RESULT # Some global setting, which defaults to `false`
    modified_result_evalution(evaluation)
  else
    # Add deprecation warning
    current_result_evalution(evaluation)
  end
end
```

3. Step: Switch global setting to true

```ruby
def result(evaluation)
  if USE_TO_CREATE_RESULT # Some global setting, which defaults to `true`
    modified_result_evalution(evaluation)
  else
    # Add deprecation warning
    current_result_evalution(evaluation)
  end
end
```

4. Step: Optional - Remove old flow

```ruby
def result(evaluation)
  modified_result_evalution(evaluation)
end
```

Each step would be a new release.

Issue #683, opened 2014-08-22
How Factory Girl create association with existing object?


crazyjin:
I have two associated models.
```ruby
class Girl < ActiveRecord::Base
   belongs_to :boy
end
class Boy < ActiveRecord::Base
   has_one :girl
end
```
And  I define factory girl for them:
```ruby
factory :girl do
  boy
  name 'saria'
end

factory :boy do
  name do
    puts create boy'
   'leo'
  end
end
```

Then i create  a girl with an exsisting boy:
```ruby
boy = create :boy
girl = create :girl, boy: boy
```

I will get following result:
```ruby
>>>create boy
>>>create boy
```

I am wondering why the block of name of boy runs  twice.
And, if i put code that can only runs once, troubles come.

Version of my factory_girl:
* factory_girl (4.4.0)
* factory_girl_rails (4.4.1)

Kriechi:
I have the same issue or question.
In my case this leads to a huge amount of time spent invoking the factory twice.
Only half of the objects are actually used, because one of the two created objects are discarded anyway.

Is there any way to prevent this from happening?
It would save a lot of time in my test suite.

kofronpi:
+1, same question as this makes difficult for me to test some things and this is not documented

Manifold0:
Ran into a similar problem where we had two associations attempting to create unique parent instances when we wanted them to belong to the same parent. 

```ruby
Class Term < ActiveRecord::Base
  has_many :candidates
  has_many :periods
end

Class Period < ActiveRecord::Base
  belongs_to :term
end

Class Candidate < ActiveRecord::Base
  belongs_to :term
end
```

In order to avoid needing to make multiple factory calls from the spec to reach the same term we ended up defining the factories like this:

```ruby
factory :period do
  term { Term.first || association(:term) }
end

factory :candidate do
  term { Term.first || association(:term) }
end
```

This way the pre-existing Term can be referred to instead of creating two instances.

joshuaclayton:
Calling `Thing.first` introduces non-determinism in the test suite, since records can be created at various layers without the developer even being aware. I'd recommend creating a record and assigning it directly if the intent is to share the value.

scaryguy:
@joshuaclayton is there any chance that you could give some practical example to how to do that?

joshuaclayton:
@scaryguy sure thing!

```ruby
term = create(:term)
period = create(:period, term: term)
candidate = create(:candidate, term: term)
```

[...]

PR #1446, opened 2020-10-08
Update GETTING_STARTED.md

johnpitchko:
Was caught on how to use factories to create associated objects that are linked. Found the answer if #683 so adding documentation to explain how this works.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -40,6 +40,7 @@ Getting Started
   + [Inline definition](#inline-definition)
   + [Specifying the factory](#specifying-the-factory)
   + [Overriding attributes](#overriding-attributes)
+  + [Association overrides] (#association-overrides)
   + [Build strategies](#build-strategies-1)
   + [`has_many` associations](#has_many-associations)
   + [`has_and_belongs_to_many` associations](#has_and_belongs_to_many-associations)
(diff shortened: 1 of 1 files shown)


GETTING_STARTED.md
```diff
@@ -584,6 +584,25 @@ factory :post do
 end
 ```
 
+Attribute overrides can be used to link associated child objects to their parent:
```
composerinteralia, line 587:
```suggestion
Attribute overrides can be used to link associated objects:
```

Could we also add a section title here to make it easier to link to this? Something like `### Association Overrides`. We would also need to manually add that section to the index at the top of this file (sorry, I haven't yet found a great way of doing that automatically).

GETTING_STARTED.md
```diff
@@ -584,6 +584,25 @@ factory :post do
...
+```ruby
+# Define factories with associations
+FactoryBot.define do
+  factory :book do
+    title { 'A Farewell to Arms' }
+   end
+   
+   factory :chapter do
+    name { 'Chapter 1' }
+    book { association :book }
```
composerinteralia, line 597:
```suggestion
    book
```

Although I think I would prefer to lean on the existing `post` and `author` factories rather than introducing new ones for this one section.

GETTING_STARTED.md
```diff
@@ -584,6 +584,25 @@ factory :post do
...
+   end
+end
+
+# Build a book and a chapter, with the chapter linked to the instantiated book (my_book)
```
composerinteralia, line 601: At that point I think we can ditch the comment and change this to:

```rb
author = build(:author)
post = build(:post, author: author)
```

composerinteralia (commented):
Thanks for the PR! I have a few comments, but overall I think it is nice to be explicit that association overrides are possible.

johnpitchko:
Awesome! I implemented your recommendations and pushed a new commit.

GETTING_STARTED.md
```diff
@@ -584,6 +585,28 @@ factory :post do
 end
 ```
 
+### Association overrides
+
+Attribute overrides can be used to link associated objects:
+
+```ruby
```
composerinteralia, line 592: Let's ditch the comments and then leave out the attributes that aren't relevant to this section. I would also like to switch to the implicit style of declaring the `author` association, which works in more cases than the inline style.

In my first set of comments I was thinking we would leave off the factory definition entirely, since we sort of cover it in previous sections, but now seeing it I like that you kept it. But I think we can get it a bit more focused.

With my comments I think we will end up with:

```ruby
FactoryBot.define do
  factory :author

  factory :post do
    author
   end
end

author = build(:author)
post = build(:post, author: author)
```

johnpitchko: Apologies for taking so long to get back to this! I applied your changes and pushed a new commit.

composerinteralia: No worries! Thanks for taking the time to improve the documentation!

PR #1448, opened 2020-10-26
Document behaviour of instance on custom initializations.

Shigawire:
…lize_with'.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -199,17 +199,15 @@ It is also possible to explicitly specify the class:
 
 ```ruby
 # This will use the User class (otherwise Admin would have been guessed)
-factory :admin, class: User
+factory :admin, class: "User"
 ```
 
-If the constant is not available
-(if you are using a Rails engine that waits to load models, for example),
-you can also pass a symbol or string,
-which factory\_bot will constantize later, once you start building objects:
+You can pass a constant as well, if the constant is available (note that this
+can cause test performance problems in large Rails applications, since
+referring to the constant will cause it to be eagerly loaded).
 
 ```ruby
-# It's OK if Doorkeeper::AccessToken isn't loaded yet
-factory :access_token, class: "Doorkeeper::AccessToken"
+factory :access_token, class: User
 ```
 
 ### Hash attributes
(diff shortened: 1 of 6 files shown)


GETTING_STARTED.md
```diff
@@ -909,6 +909,9 @@ end
 Note that this approach works with `build`, `build_stubbed`, and `create`, but
 the associations will return `nil` when using `attributes_for`.
 
+Also, note that `instance` is not available within `initialize_with`, which may produce unexpected behaviour when
```
composerinteralia, line 912: This isn't always true. `instance` would work fine if your custom `initialize_with` only instantiates the object in a custom way but doesn't assign all of the attributes.

How about:

```suggestion
Also, note that if you assign any attributes inside a custom `initialize_with` (e.g. `initialize_with { new(**attributes) }`), those attributes should not refer to `instance`, since it will be `nil`.
```

GETTING_STARTED.md
```diff
@@ -1732,7 +1735,7 @@ end
 
 This will build a hash of all attributes to be passed to `new`. It won't
 include transient attributes, but everything else defined in the factory will be
-passed (associations, evaluated sequences, etc.)
+passed (associations, evaluated sequences, etc.). Also, references `instance` are not available.
```
composerinteralia, line 1738: As above, I don't think this is always true. Let's leave this sentence off.

composerinteralia (commented):
Thank you!

GETTING_STARTED.md
```diff
@@ -909,6 +909,9 @@ end
 Note that this approach works with `build`, `build_stubbed`, and `create`, but
 the associations will return `nil` when using `attributes_for`.
 
+Also, note that if you assign any attributes inside a custom `initialize_with` (e.g. `initialize_with { new(**attributes) }`), those attributes should not refer to `instance`, since it will be `nil`.
+objects are initialized via `initialize_with { new(attributes) }`.
```
composerinteralia, line 913: Oops, it looks like we ended with an extra partial sentence here. Once we get rid of that this should be good to merge. Thanks again!

Shigawire: Damn, sorry. Fixed that!

composerinteralia:
I no longer have merge access, so I will pass this along to @entcheva. Thanks again!

entcheva:
Thank you @Shigawire @composerinteralia!

Issue #199, opened 2011-09-10
Using name-spaced models


ream88:
Hello folks,

I have Rails projects with name-spaced models. At the moment my factories look like the following with a lot of clutter, because of the name-spaced models:

```
# factories/user.rb
FactoryGirl.define do
  factory(:account, :class => MyProject::User::Account) do
    # ...
  end

  factory(:user, :class => MyProject::User::User) do
    # ...
  end

  # and so on ...
end
```

They are working, though it doesn't feel very DRY. Is (or will be) there any way to define the name-space once? Like the following?

```
FactoryGirl.define(:namespace => MyProject::User) do
  factory(:account) do
    # I'm a MyProject::User::Account class
  end
end
```

djnawara:
FYI, given this class:

```ruby
module ModuleA
  class Foo
    attr_accessor :uuid
  end
end
```

Then this works for me:

```ruby
module ModuleA
  FactoryGirl.define do
    factory :module_a_foo, class: Foo do
      sequence(:uuid) { UUIDTools::UUID.random_create }
    end
  end
end
```

So, in theory, you could do something like this:

```ruby
module ModuleA
  FactoryGirl.define(factory_name_prefix: :module_a) do
    factory :foo, class: Foo do
      sequence(:uuid) { UUIDTools::UUID.random_create }
    end
  end
end

FactoryGirl.build(:module_a_foo)
```

Just a random thought.

tf:
Yes, but this does not solve the problem of having to repeat the, for example, gem name all over the gem codebase. Always having to write `FactoryGirl.build(:some_engine_post)` hinders code mobility and makes extracting engines hard when things originally started out as `FactoryGirl.build(:post)`. This is the sort of duplication I'd like to prevent by adding the scope concept.

cec:
@joshuaclayton I believe that having this in FG would benefit not only engine creators, since namespaced models are not uncommon in non-trivial apps.

@tf Thank you for leading this discussion. 
I believe the scoping idea to be the way to go.
When writing specs for a namespaced models, it would be nice to be able to set the default_scope for those specs individually.

````
module NameSpace
  describe MyModel, type: model, factories_scope: NameSpace do
    #....
  end
end
````
Implementing this would be trivial.

mooreniemi:
I wish I had this now, as I am struggling with a large project with many name-spaces.

joshuaclayton:
[...]

PR #1453, opened 2020-11-24
Rough draft of Namespaced Factories

zspencer:
See: https://github.com/thoughtbot/factory_bot/issues/199

I have a number of clients who use FactoryBot and are working through
the difficult challene of decomposing a large monolith into either
domain-driven namespaces; or small gems/engines that are "plugged into"
a larger monolith.

In most cases, we've been able to get away with keeping the factories
bundled with their extracted modules and using distinct names.

However there's also been a bit of copy-pasting of factory code back into the monolith in cases where the names really _do_ need to be shared.

This is an attempt to bring @joshuaclayton's example of namespacing into
the core FactoryBot DSL.

If you'd prefer I packaged this as it's own gem, I'd be happy to do so;
but I'd be stoked to see this in the core FactoryBot package!


spec/acceptance/namespaced_factories_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+        with_namespace(:Movies, require_prefix: false) do
+          factory :category do
+            name { "Emmy Winners" }
+          end
+        end
+      end
+
+      instance = FactoryBot.build(:category)
+
+      expect(instance).to be_kind_of(Category)
```
entcheva, line 68: Can you explain why you think this is the correct behavior? As opposed to clobbering the existing factory, or raising a `DuplicateDefinitionError`?

spec/acceptance/namespaced_factories_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      expect(instance.name).to eql("Classics")
+    end
+  end
+
+  context "when prefixes are not required" do
+    it "uses the namespaced factory without the prefix" do
+      FactoryBot.define do
+        with_namespace(:Movies, require_prefix: false) do
```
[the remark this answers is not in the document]
zspencer: I think I could go either way, with a slight preference to not include a default so that we don't introduce breaking changes by swapping the default later.

spec/acceptance/namespaced_factories_spec.rb
zspencer: Sure, the use case I am currently fiddling with is decomposing a monolith into domain-specific namespaces (and some rails engines); and there's some integration tests that have cross-cutting concerns and also some name collisions.

The integration test suite is currently loading definitions for the following packages:

1. "Core" (the main monolith being decomposed)
2. "ProductA" 
3. "ProductB" 

Each of these have a `User` class; with factory definitions named `:user`; as requiring the developers to use `:product_a_user` within the `ProductA` domain was not particularly desired.

If we clobbered the existing `:user` factory, then existing tests which loaded the `:user` definition from the ProductA module began to fail due to different default values, as well as the reduced surface area of attributes and traits in the ProductA namespace.

Similarly, if we raised the `DuplicateDefinitionError`, our integration tests would stop loading definitions once it attempted to load the `ProductA` `:user` factory.

As a result, discarding the unprefixed duplicate seemed like the safest way to allow folks working within the ProductA/ProductB namespaces the convenience of referencing `create(:user)` while still allowing downstream integration test consumers to  leverage the `:product_a_user` and `:product_b_user` factories.

spec/acceptance/namespaced_factories_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+      FactoryBot.define do
+        factory :category do
+          name { "Classics" }
+        end
+
+        with_namespace(:Movies) do
+          factory :category do
+            name { "Emmy Winners" }
+          end
+        end
```
entcheva, line 18:
```suggestion
        factory :movies_category, class: "Movies::Category" do
          name { "Emmy Winners" }
        end
```

@zspencer Can you expand on what your suggestion gives us that this existing factory_bot syntax does not? 

@composerinteralia and I are working to understand what problem this is solving for your use case that's not covered by existing functionality. Thanks!

zspencer: Sure, given the following application structure:

`core` - A Rails Application that serves as the primary monolith
`product-a` - A Rails Engine that holds the Product A teams code
`product-b` A Rails Engine that holds the Product B teams code

We use FactoryBot across all three codebases; with the `core` code-base having a mix of end-to-end integration tests that want to use the Factories defined in `product-a` and `product-b`

In `core`, `product-a` and `product-b` all of them have `User` classes. In `core` it's simply `User`, but in `product-a` it is `ProductA::User` and in `product-b` it is `ProductB::User`.

When writing tests in the `product-a/spec/` suite, we want to use `FactoryBot.build(:user)` instead of `FactoryBot.build(:product_a_user)`. Same in the `product-b/spec/` suite.

However, in `core` we sometimes want to leverage traits defined in `product-a/spec/factories/product-a/users.rb`. To do so, we have shared the factories with the `core` code-base. _However_ because most of these factories were defined as `factory :user, class: "ProductA::User"` this results in a duplicate factory name definition.

To avoid having to go through and rename all the `FactoryBot.create(:user)` in `product-a/spec` to `FactoryBot.create(:product_a_user)` we sprouted the `with_namespace` method to wrap factories that are shared between `product-a`/`product-b` and `core`. 

Now, when in `product-b/spec/` we can call `FactoryBot.create(:user)` and it will use the factories in `product-b/spec/factories`, and in `core` we can call `FactoryBot.create(:user)` and it will use the top-level User, or `FactoryBot.create(:product_b_user)` and it will use the factory defined in `product-b/spec/factories/users.rb`

This is really only a useful feature in cases where a group has started to decompose a large monolith; so it may not be worth including in `FactoryBot` core.

elliotcm:
@entcheva I'm an unrelated third party, but is there any progress on this?  I looks like this PR is waiting on ThoughtBot's response.

zspencer:
@elliotcm - I've been using this in production for about a year and a half now. It's not perfect; but I'd be down to publish it as a standalone gem sometime this week if you'd like.

elliotcm:
> @elliotcm - I've been using this in production for about a year and a half now. It's not perfect; but I'd be down to publish it as a standalone gem sometime this week if you'd like.

@zspencer That'd be great!  Thanks.

zspencer:
@elliotcm - https://github.com/zinc-collective/factory_bot_namespaced_factories Here ya go! I just pulled it out of the production application I've been using it in since 2020. We mostly relied on our relatively comprehensive test suite to actually test the gem; but if you want to add better docs / specs / etc. I'd be happy to take patches!

gap777:
@zspencer AGPL? Wouldn't that license prevent me from using this in a closed-source business application?

zspencer:
@gap777 - My understanding is that AGPL is safe to use in proprietary applications; so long as you're not distributing it as part of your production deployment. That said, I believe Sidekiq uses the LGPL so maybe I am mixed up.

PR #1461, opened 2021-01-16
Fix default value for attribute named `new`

skryukov:
This is a fix to make it possible to define an attribute named `new` (which is a bad name for an attribute by the way =)).

```
FactoryBot.define do
  factory :post do
    add_attribute(:title) { "Title" }
    add_attribute(:sequence) { "Sequence" }
    add_attribute(:new) { true }
  end
end

FactoryBot.attributes_for(:post)
# => {:title=>"Title", :sequence=>"Sequence", :new=>true}

FactoryBot.build(:post)
# => #<Post id: nil, title: "Title", sequence: "Sequence", new: nil>
```

### DIFF
--- a/lib/factory_bot/attribute_assigner.rb
+++ b/lib/factory_bot/attribute_assigner.rb
@@ -37,8 +37,9 @@ module FactoryBot
     end
 
     def decorated_evaluator
-      Decorator::InvocationTracker.new(
-        Decorator::NewConstructor.new(@evaluator, @build_class)
+      Decorator::NewConstructor.new(
+        Decorator::InvocationTracker.new(@evaluator),
+        @build_class
       )
     end
 
--- a/spec/acceptance/add_attribute_spec.rb
+++ b/spec/acceptance/add_attribute_spec.rb
@@ -1,0 +1,37 @@
+describe "#add_attribute" do
+  it "assigns attributes for reserved words on .build" do
+    define_model("Post", title: :string, sequence: :string, new: :boolean)
+
+    FactoryBot.define do
+      factory :post do
+        add_attribute(:title) { "Title" }
+        add_attribute(:sequence) { "Sequence" }
+        add_attribute(:new) { true }
+      end
+    end
+
+    post = FactoryBot.build(:post)
+
+    expect(post.title).to eq "Title"
+    expect(post.sequence).to eq "Sequence"
+    expect(post.new).to eq true
+  end
+
+  it "assigns attributes for reserved words on .attributes_for" do
+    define_model("Post", title: :string, sequence: :string, new: :boolean)
+
+    FactoryBot.define do
+      factory :post do
+        add_attribute(:title) { "Title" }
+        add_attribute(:sequence) { "Sequence" }
+        add_attribute(:new) { true }
+      end
+    end
+
+    post = FactoryBot.attributes_for(:post)
+
+    expect(post[:title]).to eq "Title"
+    expect(post[:sequence]).to eq "Sequence"
+    expect(post[:new]).to eq true
+  end
+end


spec/acceptance/add_attribute_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+        add_attribute(:new) { true }
+      end
+    end
+  end
+
+  context "returning attributes for a factory" do
+    subject { FactoryBot.attributes_for(:post) }
+
+    it "assigns default values" do
+      expect(subject[:title]).to eq "Title"
```
gabrielbaldao, line 18: What do you think about use `have attribute` instead have 3 expects
https://relishapp.com/rspec/rspec-expectations/docs/built-in-matchers/have-attributes-matcher

skryukov: @gabrielbaldao, thanks for the review!

`#have_attributes` is looking good to me, but I did not find any usage of this method in the code base, so I think I'll leave it as it is for the sake of consistency. =)

lib/factory_bot/attribute_assigner.rb
```diff
@@ -43,7 +43,7 @@ def decorated_evaluator
     end
 
     def methods_invoked_on_evaluator
-      method_tracking_evaluator.__invoked_methods__
+      method_tracking_evaluator.__invoked_methods__ - [:new]
```
entcheva, line 46: What are the implications of removing a method from `method_tracking_evaluator.__invoked_methods__`?

To me, this starts to establish a less than ideal pattern of special cases for individual methods. 

Since this is likely a rather uncommon scenario, I wonder if you could work around it using transient attributes and callbacks?

```ruby
    FactoryBot.define do
      factory :post do
        add_attribute(:title) { "Title" }
        add_attribute(:sequence) { "Sequence" }
        transient do 
          new_attribute { true }
        end

        after(:build)  { |post| post.new = new_attribute }
      end
    end
```

skryukov: Hi, @entcheva, thanks for the review!

You are right: it was an ugly fix =) But I think I came up with a better idea.

lib/factory_bot/attribute_assigner.rb
```diff
@@ -37,8 +37,9 @@ def method_tracking_evaluator
     end
 
     def decorated_evaluator
-      Decorator::InvocationTracker.new(
-        Decorator::NewConstructor.new(@evaluator, @build_class)
+      Decorator::NewConstructor.new(
```
composerinteralia, line 40: Ha, clever! Yeah, I'm good with this change.

I want to say there might still be an edge case if you have `add_attribute(:new)` and then try to use the `new` attribute inside `initialize_with`, but I think that's an acceptable edge case.

spec/acceptance/add_attribute_spec.rb
```diff
@@ -0,0 +1,33 @@
+describe "add_attribute" do
```
composerinteralia, line 1: A few thoughts about this test:

- We've been gradually moving away from using `let`, `before`, and `subject` for the reasons mentioned in this [blog post](https://thoughtbot.com/blog/lets-not). I'd prefer moving the setup into the test itself.
- I'd like to see some mention of assigning "[reserved words](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#method-name--reserved-word-attributes)" in the test description, since this is the main reason for using `#add_attribute`. 
-. I'm not sure we need separate tests for the `attributes_for` and `build` strategies here, since the strategies themselves are tested well enough elsewhere.

Putting that all together, I think we'd get something like:

```rb
describe "#add_attribute" do
  it "assigns attributes for reserved words" do
    define_model("Post", title: :string, sequence: :string, new: :boolean)
    
    FactoryBot.define do
      factory :post do
        add_attribute(:title) { "Title" }
        add_attribute(:sequence) { "Sequence" }
        add_attribute(:new) { true }
      end
    end

    post = FactoryBot.build(:post)

    expect(post.title).to eq "Title"   
    expect(post.sequence).to eq "Sequence"
    expect(post.new).to eq true
  end
end
```

composerinteralia (commented):
Nice work digging into the guts of factory_bot and finding such a clean solution. A few thoughts about the test, but otherwise looks good to me.

spec/acceptance/add_attribute_spec.rb
skryukov: Hi, @composerinteralia! Thanks for the review, hope this will do =)

PR #1464, opened 2021-02-05
Support definition file paths with extensions

elia:
In the process of adding reload support to @solidusio which has the need to control the loading sequence of its factories in relation to those provided by applications and extension, I wished I had the ability to add paths coming from a glob (e.g. coming from `Dir["…"]`).


composerinteralia:
> adding reload support to @solidusio

Oh, that's exciting. You might be interested in https://github.com/thoughtbot/factory_bot_rails/issues/341#issuecomment-538614110

navcdr:
Please unsubscribe me.

elia:
@composerinteralia thanks for pointing that out!
If you (or anyone from this project) are interested you can follow all the required gimmicks in this PR: https://github.com/solidusio/solidus/pull/3907.

navcdr:
Unsubscribe

composerinteralia:
There is nothing we can do about your notification preferences. If you don’t want to get factory_bot notifications then you will need to stop “watching” the repository.

spec/factory_bot/find_definitions_spec.rb
```diff
@@ -109,4 +109,36 @@ def self.in_directory_with_files(*files)
       end
     end
   end
+
+  describe "definition_file_paths" do
+    in_directory_with_files "spec/my_factories/factory_one.rb", "spec/my_factories/factory_two.rb"
+
+    before { allow(FactoryBot).to receive(:load) }
+
+    it "supports files without extension" do
```
composerinteralia, line 118: We've already testing supporting files without extension and directories, so I don't think we need additional coverage there. I think the last test is enough.

lib/factory_bot/find_definitions.rb
```diff
@@ -5,21 +5,27 @@ class << self
     # "factories", "test/factories" and "spec/factories". Only the first
     # existing file will be loaded.
     attr_accessor :definition_file_paths
-  end
 
-  self.definition_file_paths = %w[factories test/factories spec/factories]
+    def find_definitions
```
composerinteralia, line 9: This is a larger diff than I was expecting for this change. Would adding `load(path) if File.file?(path)` to the existing code be enough to get the test passing?

composerinteralia (commented):
Unsure whether you still need this for Solidus, but either way it seems reasonable to accept paths with extensions. Thanks for working on it!

spec/factory_bot/find_definitions_spec.rb
elia: I think all the existing coverage is implicit about the contents of `definition_file_paths`, this was meant to pin the supported cases.

I can surely remove the two extra examples, is that your preference?

lib/factory_bot/find_definitions.rb
elia: You got it right, this was just a refactoring to make it look more linear, I'll drop the commit from this PR 👍

elia: Oh, now I remember, that was also in response to CodeClimate flagging that method for increased complexity, I still have the commit locally, let me know if you want me to add it back.

spec/factory_bot/find_definitions_spec.rb
composerinteralia: Yeah, I see what you mean. It looks like the only place we actually test setting `definition_file_paths` is in a cucumber test.

Let's keep your additional test, but instead of stubbing `definition_file_paths` set it with something like:

```rb
it "" do
  old_paths = FactoryBot.definition_file_paths
  FactoryBot.definition_file_paths = ["whatever/paths/you/want/for/that/test"]
  
  # test stuff
ensure
  FactoryBot.definition_file_paths = old_paths
end
```

That will point us in the direction of eventually getting rid of the cucumber specs.

lib/factory_bot/find_definitions.rb
composerinteralia: What do you think about extracting a helper method instead. Something like:

```ruby
def load_file_or_directory(path)
  load("#{path}.rb") if File.exist?("#{path}.rb")
  
  load path if File.file? path

  if File.directory? path
    Dir[File.join(path, "**", "*.rb")].sort.each do |file|
      load file
    end
  end
end
```

elia: Good call! That was enough for CodeClimate ✅ 

One thing I just realized was in the refactoring is that there was a protection against loading a file twice (say you have `foo` and `foo.rb` in the paths). Honestly I don't remember if it was something we encountered on some solidus extension or it was just a side-effect of the refactoring.

You think would be worth adding a spec for that and fixing it in a separate PR?

PR #1465, opened 2021-02-05
Replace Travis with GitHub Actions

elia:


composerinteralia:
I'm not seeing any CI build happening on this branch yet. Looks like somebody had a similar idea over in https://github.com/thoughtbot/factory_bot/pull/1469

elia:
@composerinteralia the results are available here https://github.com/nebulab/factory_bot/actions which is the repo this PR is coming from

aledustet:
@elia I went ahead and rebased master onto your branch to require the jruby tests to pass(they are fixed 🎉 ) I tried to update the fork but unfortunately I don't have permissions. I created a [branch](https://github.com/thoughtbot/factory_bot/tree/nebulab-elia/gh-actions) where I did that. We could use that one and merge it or we can update the fork.

elia:
Thanks @aledustet, I added your commit and rebased 👍
_Sorry about not allowing edits, not sure why is that, I find that feature very useful, maybe I clicked inadvertently 🤷‍♂️_

composerinteralia:
Sweet! Thanks for working on this @elia and thanks for merging it @aledustet. It seems like the allow_failure part isn't quite working as expected—all the required parts are passing, but it's still showing up as a failed build. But we can follow up and fix that later. It's definitely better than the failing build we had before this was merged!

composerinteralia:
FWIW, https://github.com/thoughtbot/factory_bot/pull/1469 was using continue-on-error instead of allow_failure. It looks nice an concise, if it does what we need. If we end up using that, let's be sure to add them as a co-author on that commit.

PR #1466, opened 2021-02-07
Fix stub strategy for non-ActiveRecord models

splattael:
The following example failed previously with:

```
NoMethodError: undefined method `id' for #<Foo:0x...>
```

```ruby
class Foo
end

FactoryBot.define do
  factory :foo do
    skip_create # non ActiveRecord object
  end
end

FactoryBot.build_stubbed(:foo) # => NoMethodError above
```

It's a follow-up of https://github.com/thoughtbot/factory_bot/pull/1316.

This allows GitLab to use non-`ActiveRecord` models with FactoryBot :tada: See https://gitlab.com/gitlab-org/gitlab/-/merge_requests/52923

### DIFF
--- a/lib/factory_bot/strategy/stub.rb
+++ b/lib/factory_bot/strategy/stub.rb
@@ -75,7 +75,7 @@ module FactoryBot
       end
 
       def has_settable_id?(result_instance)
-        !result_instance.class.respond_to?(:primary_key) ||
+        result_instance.class.respond_to?(:primary_key) &&
           result_instance.class.primary_key
       end
 
--- a/spec/acceptance/stub_spec.rb
+++ b/spec/acceptance/stub_spec.rb
@@ -99,3 +99,21 @@ describe "a stubbed instance with no primary key" do
     end
   end
 end
+
+describe "a stubbed instance for non active record models" do
+  it "builds a stubbed instance" do
+    define_class("NonArModel")
+
+    FactoryBot.define do
+      factory :non_ar_model do
+        skip_create
+      end
+    end
+
+    model = FactoryBot.build_stubbed(:non_ar_model)
+
+    expect(model).to be_truthy
+    expect(model).to be_persisted
+    expect(model).not_to be_new_record
+  end
+end


splattael:
:wave: @composerinteralia 

Do you mind reviewing this small fix? :pray:

aledustet:
Hi @splattael thank you for your suggestion. 
We are having trouble fully understanding the scenario where this feature might be used. Your addition talks about a non active record class, but the tests mention a couple of active record behaviors, like checking for the class to be persisted. 
 
While your example and solution are very specific. It also feels like if we change the behavior of this strategy to work with a different ORM than active record, it might open the door for factory bot to expand support into other ORM's.

I wonder if writing a [custom strategy] could solve the problem that prompted this pull request?

[custom strategy]: https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#custom-strategies

splattael:
@aledustet Thanks for the review :bow: 

> We are having trouble fully understanding the scenario where this feature might be used. Your addition talks about a non active record class, but the tests mention a couple of active record behaviors, like checking for the class to be persisted.

Yes, according to https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#custom-construction one can use "Build non-ActiveRecord objects" with FactoryBot which is expected to support all common strategies (create, build, and **build_stubbed**). This works like a charm for all strategies but `build_stubbed` and I believe this a bug hence this PR.

> While your example and solution are very specific. It also feels like if we change the behavior of this strategy to work with a different ORM than active record, it might open the door for factory bot to expand support into other ORM's.

I am not sure I understand the concern. :thinking:  Do you mind giving an example? :pray: 

Note that changing this single condition did not cause any spec failures. Are we missing a spec then? :shrug:

> I wonder if writing a custom strategy could solve the problem that prompted this pull request?

Not sure since FactoryBot's support for "non-ActiveRecord objects" works for `create` and `build` but not for `build_stubbed` I believe.

Now, the question is "Why would one use `build_stubbed` instead of `build` for 'non-ActiveRecord objects'"?

Although it's obvious when the factory is used directly (via `foo = build(:foo)`) it's sometimes not obvious if the non-ActiveRecord factory is used in an association (implicit, explicit or inline definition).

Please consider this example:

```ruby
class Foo < ApplicationRecord
  attr_accessor :bar
end

class Bar
end

FactoryBot.define do
  factory :foo, class: 'Foo' do
    bar
  end

  factory :bar, class: 'Bar'
end

p FactoryBot.build(:foo).bar # => #<Bar:0x000055a931ac0528>
p FactoryBot.build_stubbed(:foo).bar # => NoMethodError: undefined method `id' for #<Bar:0x000055a931ac6dd8>
```

I believe this a bug. :bug: 

Moreover, looking at the condition in question

```ruby
       def has_settable_id?(result_instance)
         !result_instance.class.respond_to?(:primary_key) ||
           result_instance.class.primary_key
        end
```

I am not sure I understand why it's assumed that a non-ActiveRecord object (lacking a class method `primary_key`) should have a "settable id". This seems like logical error to me :man_shrugging: 

Since I am not very familiar with FactoryBot's codebase it's also likely I am missing a point here, of course :sweat_smile: 

If it's the case we should update the documentation (non-ActiveRecord object must have a "settable id") and add a missing spec as mentioned above :nerd_face:

composerinteralia:
Thanks for your detailed response. That example helps a lot.

Based on that example it seems like you are calling `build_stubbed` on an ActiveRecord model, but the model has an association that is NOT an ActiveRecord model, and we [continue using `build_stubbed` as the strategy](https://github.com/thoughtbot/factory_bot/blob/18bbc5b5822e32aa8259a1f67fb6296fcf04a2b0/lib/factory_bot/strategy/stub.rb#L32) for any associations. That definitely seems like a bug—we don't want to use `build_stubbed` for non-ActiveRecord associations. I don't have a great solution at the moment.

I don't think this PR is the way to fix it though—there is a lot of ActiveRecord-specific stuff in `build_stubbed`, and it's not really meant for other kinds of objects. It defines [all these methods](https://github.com/thoughtbot/factory_bot/blob/18bbc5b5822e32aa8259a1f67fb6296fcf04a2b0/lib/factory_bot/strategy/stub.rb#L6-L25) on the object, for example—you don't want a non-ActiveRecord object to all of a sudden have a bunch of extra methods in test. There's probably room to improve the documentation on this.

composerinteralia:
As a workaround, you can use a regular attribute for bar instead of an association. That would allow you to specific `build` explicitly:

```rb
  factory :foo, class: 'Foo' do
    bar { build(:bar) }
  end
```

splattael:
@composerinteralia Thanks a ton for clarifying the caveats of `build_stubbed` and _why_ it should not be used for non-ActiveRecord objects :bow: :heart: 

Now it all makes sense to me and I realize that although the suggested fix works it could potentially break integrations with other ORMs and should be avoided :sweat_smile: 

I will close this PR and go with [workaround](https://github.com/thoughtbot/factory_bot/pull/1466#issuecomment-791962096) :muscle: Thanks again for suggesting it!

Now I wonder if we should:
* Document the caveat about using `build_stubbed` with non-ActiveRecord and potential workaround
* Document that `build_stubbed` uses `primary_key` for ActiveRecord objects or the hardcoded `id` in case of other ORMs
* Add a spec (use `build_stubbed` with other ORM) to prevent future accidental breakage of `has_setter_id?`

Also, shout out to @aledustet :heartbeat: for the initial review and making me add a concrete example which I should have included initially.

composerinteralia:
I think the answer to whether we should add more documentation is generally yes. 😁

And to clarify a bit more, `build_stubbed` isn't meant to work with other ORM's either—only Active Record. The `has_setter_id?` stuff was added in https://github.com/thoughtbot/factory_bot/commit/5a3247dd8c77e8ae98c4d1586726e382b143406c#diff-6a3e752baa14b9dc9ad10420765645dd115085bae6e17a1a046cd90d6e91acef to addresses some differences in different versions of Active Record.

PR #1469, opened 2021-02-10
Use Github Actions for CI

mlarraz:
* Remove lockfiles and adds them to `.gitignore`. Not sure why these were committed but they were causing installation issues.
* `standard` fails on the newest version, so pin to what was in the lockfiles for now.
* `did_you_mean` doesn't seem to by enabled by default in Ruby < 2.7, so don't test it there

Sample: https://github.com/mlarraz/factory_bot/actions/runs/555863648

image

There are still some failures in ActiveRecord `master`, Ruby `head`, and jruby, but I've marked them as allowed for now. My original branch had significantly more changes addressing these and other issues, but I'm trying to keep this as small as possible for the sake of review.


composerinteralia:
Cool, thanks for working on this!

- I'm good with ignoring and removing the lockfiles. We already do that on factory_bot_rails. The downside is that the tests get a bit flakier, but the upside is that we catch breaking changes with other gems sooner.
- I've fixed the standard violations in https://github.com/thoughtbot/factory_bot/commit/630aca4b870f0c937a63250f96afa783f25e78a5, so let's avoid pinning to an older version.
- I'm not loving the did_you_mean changes. Those same tests are all passing on Travis, so I'd like to figure out what's different here.

mlarraz:
Sounds good.

I've updated everything except the did_you_mean changes. I'm also not clear on why this is happening, but it seems like the behavior is different on Ruby 2.5 and 2.6 in Github Actions than in Travis.

composerinteralia:
Pardon my ignorance here, but how do I see the results of the GitHub Actions CI build? The "Checks" tab here doesn't have anything in it. Is there something I need to do to get this thing running?

mlarraz:
> Pardon my ignorance here, but how do I see the results of the GitHub Actions CI build? The "Checks" tab here doesn't have anything in it. Is there something I need to do to get this thing running?

The issue is Github Actions won't let PRs from forks add new checks. The check will only show up in the base repo once the PR is merged. You can see the latest run at my fork: https://github.com/mlarraz/factory_bot/actions/runs/585297899

I *think* I could work around this by adding the workflow with only trivial steps in one PR, then sending a second PR with the actual rake task step after the first one gets merged.

composerinteralia:
Thank you so much for your work on this PR! I really appreciate it. We ended up merging https://github.com/thoughtbot/factory_bot/pull/1465, both because it was opened first and because the tests were passing.

PR #1470, opened 2021-02-11
Remove Null Strategy

NoahTheDuke:
I couldn't determine what the purpose of the Null Strategy is, given that it's only used in the `AttributesFor#association` and simply returns `nil`. I tried substituting that method call for `nil`, and everything still works. Seems like it might be worth excising.


aledustet:
Thanks for the addition, this seems like a reasonable change, however, this would require a major release since it is removing an existing feature. Even though probably no one is using the strategy, and it's meant to be private, we can't be sure it won't result in a breaking change. When it's time for a new release we'll take a look at this again.

PR #1471, opened 2021-02-13
Fix delegation

eregon:
* (*args, **kwargs) delegation is incorrect in 2.7, only ruby2_keywords works on 2.7.
* See https://eregon.me/blog/2021/02/13/correct-delegation-in-ruby-2-27-3.html or https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/

I noticed this while investigating https://github.com/oracle/truffleruby/issues/2258.

### DIFF
--- a/lib/factory_bot/decorator/invocation_tracker.rb
+++ b/lib/factory_bot/decorator/invocation_tracker.rb
@@ -6,22 +6,11 @@ module FactoryBot
         @invoked_methods = []
       end
 
-      if ::Gem::Version.new(::RUBY_VERSION) >= ::Gem::Version.new("3.0")
-        def method_missing(name, *args, **kwargs, &block) # rubocop:disable Style/MissingRespondToMissing
-          @invoked_methods << name
-          super
-        end
-      elsif ::Gem::Version.new(::RUBY_VERSION) >= ::Gem::Version.new("2.7")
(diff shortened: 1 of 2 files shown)


composerinteralia:
Will this nightmare ever end? 😭🤣

Thank you! This looks much better, and it looks like we probably did have some incorrect behavior in the evaluator.

lib/factory_bot/decorator/invocation_tracker.rb
```diff
@@ -6,22 +6,11 @@ def initialize(component)
...
-      elsif ::Gem::Version.new(::RUBY_VERSION) >= ::Gem::Version.new("2.7")
-        ruby2_keywords def method_missing(name, *args, &block) # rubocop:disable Style/MissingRespondToMissing
-          @invoked_methods << name
-          super
-        end
-      else
-        def method_missing(name, *args, &block) # rubocop:disable Style/MissingRespondToMissing
-          @invoked_methods << name
-          super
-        end
```
eregon, line 23: The version above was correct, but it seemed a lot of duplication, so I used the simpler pattern here too.

eregon:
Yeah, delegation in 2.7 is quite tricky.

This PR lets the gem work on TruffleRuby and fixes delegation in general, so it would be nice to merge it and if possible have a release with this fix :)

PR #1474, opened 2021-02-20
Test against latest versions of Ruby and Rails

composerinteralia:
This adds Ruby 3 and Rails 6.1 to the test matrix. It also bumps all the
dev dependencies listed in the Gemfiles to the latest versions.

We may end up switching over to GitHub actions, but this should get us
covered in the meantime.


composerinteralia:
@mvz I recall you had added jruby back to our test matrix at one point. I've just moved the jruby tests back to allowed failures, since they are failing and I don't know how to fix them, but I'd be most grateful if you could help get them passing again.

mvz:
@composerinteralia I think it's due to an issue with I18n and JRuby. I made #1475 to fix it. I was getting a slightly different error with a slightly different version of JRuby locally, so we'll see what the results on Travis are.

mvz:
Right, so that that still fails on JRuby 9.2.8.0. For some reason I can't even get bundler to run locally on 9.2.8.0, so I've bumped the version used on Travis CI to 9.2.14.0 instead.

PR #1475, opened 2021-02-21
Fix JRuby CI build

mvz:


composerinteralia:
Thank you! Not sure what the situation is with those did_you_mean failures. At one point or another they were passing—not sure what's different now, but the PR trying to switch from travis to GitHub actions (https://github.com/thoughtbot/factory_bot/pull/1469) is having the same problem.

composerinteralia:
I merged this, thinking I could address the did_you_mean failures with https://github.com/thoughtbot/factory_bot/pull/1482, but sadly that didn't help. I may skip that test entirely for JRuby.

PR #1483, opened 2021-04-01
Acknowledge there are more than two genders

Aliagender:


GETTING_STARTED.md
```diff
@@ -1154,6 +1154,12 @@ factory :user do
     gender { "Female" }
     login { "#{name} (F)" }
   end
+  
+  trait :other do
+    name { "It Doe" }
```
composerinteralia, line 1159: We'd need to use a different name here. "It" might well be somebody's name, but in this particular context it comes across as describing a person with the English word "it". That's not OK.

composerinteralia (commented):
Thanks for bringing this up. I recall us having similar problems in some of our specs, and I've wanted to make those more inclusive.

I think adding an additional genders is an improvement, but I also don't think gender is a particularly good example for trait usage. I'd like to change this to a more typical example from real software, like :active and :inactive.

composerinteralia:
I've changed this to use an active/inactive status instead in 92114db048b70096183af622173506535b150e65, and listed you as a co-author.

---

Based on past experience, I'm also going to preemptively remind folks who come across this issue that we have a [Code of Conduct](https://github.com/thoughtbot/factory_bot/blob/master/CODE_OF_CONDUCT.md) we expect you to follow.

PR #1487, opened 2021-04-16
Improve Github Actions labels

mlarraz:
Before and after:

imageimage


composerinteralia:
Love it. Thanks!

If you know how to get the allowed failures to not show up as unsuccessful checks, I'd be grateful for that change too. No pressure though.

PR #1493, opened 2021-05-08
Add TruffleRuby in CI

eregon:
All jobs pass on TruffleRuby: https://github.com/eregon/factory_bot/actions/runs/823666810

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,7 @@ jobs:
       matrix:
         ruby:
           - jruby-9.2.16.0
+          - truffleruby
           - ruby-head
           - "3.0"
           - "2.7"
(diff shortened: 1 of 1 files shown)


composerinteralia:
Nice! Our allowed failures stuff wasn't working quite as expected, so I changed some things around in https://github.com/thoughtbot/factory_bot/commit/ffcd343b9ab2b5acc3d26834af8976610cdb6b03 and merged those changes into your branch.

The one catch is that I don't think any of the current maintainers use truffleruby, so if anything goes awry with these builds we'll likely reach out to you for help (we've got somebody helping us with the jruby builds as well, and I think I've only had to bother them them once or twice).

eregon:
> The one catch is that I don't think any of the current maintainers use truffleruby, so if anything goes awry with these builds we'll likely reach out to you for help

Yes, that's fine, you can ping me or file an issue to https://github.com/oracle/truffleruby/issues (can be as simple as "it fails on TruffleRuby but works on CRuby").
We also track CI stauses via https://github.com/eregon/truffleruby-gem-tracker, so we should notice within a week if it starts failing for some reason.

PR #1497, opened 2021-06-02
Update GETTING_STARTED.md with mention Array#cycle

aishek:
Hey, comrades! Many thanks for the great tool!

I've created this PR just to improve the Getting Started guide a little. I found that often the pattern for sequences is to use an array to cycle through its values. So I believe it will save some time for developers if the proper way to use arrays will be mentioned in docs. But I'm not sure if I mentioned the best one 😀

`sequence :priority, %i[low medium high urgent].cycle`

What do you think?

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -56,6 +56,7 @@ Getting Started
   + [Aliases](#aliases-1)
   + [Rewinding](#rewinding)
   + [Uniqueness](#uniqueness)
+  + [Arrays](#arrays)
 * [Traits](#traits)
   + [Defining traits](#defining-traits)
   + [As implicit attributes](#as-implicit-attributes-1)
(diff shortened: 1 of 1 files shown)


aledustet (commented):
Thanks so much for the contribution. Mentioning a specific array method(cycle) is a subset of what the value on a sequence can be. A more general statement would be to mention that the value for the sequence could be *any* Enumerable instance, as long as it responds to `next` as mentioned briefly in the [aliases section](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#aliases-1) of the sequence documentation.
We could probably improve the documentation a bit by moving this mention into a more obvious place, the [without block section](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#without-a-block) might be a good fit. As part of the documentation improvement we could add an example like the one you included, I just don't think it should get its separate section.

aishek:
Good point! I've made the changes.

PR #1504, opened 2021-06-30
fix attribute name mismatch in the associations example

iiwo:
Follow-up to https://github.com/thoughtbot/factory_bot/pull/1446
looks like `name` should be used here (in line with this and other examples), not `author_last_name`, this is how it was suggested in [this PR review comment](https://github.com/thoughtbot/factory_bot/pull/1446#discussion_r516157987) but got mixed up in the end somehow


composerinteralia:
Nice catch. Thanks!

PR #1507, opened 2021-07-29
Update factory definitions

tiffany-hsu:
Updated syntax to FactoryBot


composerinteralia:
Thanks so much for the PR.

It looks like you've opened this up against the 4.9.0 branch, rather than the main branch. Version 4.9.0 was before we updated to the new name and syntax, so the documentation is correct for that version. We don't really use the 4.9.0 branch for much any more—we keep it around mostly so people can easily refer to the old upgrade documentation.

PR #1508, opened 2021-08-05
Add spaces to enum_attribute in the Enum traits column of GETTING_STARTED.md

tnandate:
In other places, hash literals were having spaces around them, so I added spaces.


composerinteralia:
Thanks for the PR! We're using [standard](https://github.com/testdouble/standard) on this project, which prefers no spaces inside hash literal braces. I think the examples should match that style, although we don't have an automated way to ensure that right now. I'd accept a PR to fix any hash literals with spaces inside to remove the spaces.

Issue #1452, opened 2020-11-19
enum traits do not use prefix


tjbarker:
### Description

When defining enums with a prefix or suffix the trait that is automatically generated does not take into account the prefix or suffix.

### Reproduction Steps

```
class Foo
  enum bar: [:baz, :bing], _prefix: :bang 
end

FactoryBot.define do
  factory :foo do
    ...
  end
end

# in a test
create(:foo, :bang_baz)
```

### Expected behavior

I would expect the trait to correspond to the predicate or bang methods defined with the enum.
in the above examples an instance of Foo would have methods `bang_baz?` and `bang_baz!`.

### Actual behavior

The traits are define simple from the enum values. So in the above example the traits would be used:
```
create(:foo, :baz)
create(:foo, :bing)
```

### System configuration
**factory_bot version**:  6.1.0
**rails version**:  5.2.4.4
**ruby version**: 2.5.5

rhannequin:
Hello, it looks like it is not due to FactoryBot's implementation but ActiveRecord's.
FactoryBot retrieves enum values from its pluralized name, [as defined in ActiveRecord](https://github.com/rails/rails/blob/master/activerecord/lib/active_record/enum.rb#L162).

ActiveRecord doesn't itself include suffix and prefix in the list of enum's values:

```rb
class Book < ApplicationRecord
  enum language: %i[english japanese], _prefix: :in
end

Book.languages # => {"english"=>0, "japanese"=>1}
```

So a solution would be to change ActiveRecord's enum definition but forcing to use prefix and suffix in the list of enum's values. But I don't think this is the way for this feature: `_prefix` and `_suffix` were added to avoid conflict between a model's column and an enum's value (https://github.com/rails/rails/issues/17511).

What do you think?

agvald:
Hi, I recently ran into this problem and decided to take a look at it.
I've taken a look at ActiveRecord and I don't know how I would go about creating a public api to inspect how the enums objects are constructed, but what i do now is  meta programming...

I wonder if a PR with a somewhat dirty solution like i attached here would be accepted? Of course it needs some cleaning, and better error handling, and a couple of specs. 
Please let me know if this is interesting, I'll gladly take a look at it :)

In lib/factory_bot/enum.rb we can redefine the `enum_values` to be like this:
```ruby
def enum_values(klass)
  return @values if @values
  # Fetching AcitveRecord::Enums

  enum = klass.send(@attribute_name.to_s.pluralize)
[...]

PR #1513, opened 2021-09-20
[1452] update traits_for_enum to accept _prefix and _suffix

tjbarker:
in response to [Issue](https://github.com/thoughtbot/factory_bot/issues/1452).

goal is to allow _prefix and _suffix for enum trait definitions to stay in line with ruby implementations
also will not define the standard enum (without prefix/suffix) will only define the ones with prefix/suffix

### DIFF
--- a/lib/factory_bot/definition.rb
+++ b/lib/factory_bot/definition.rb
@@ -173,7 +173,12 @@ module FactoryBot
     end
 
     def automatically_register_defined_enums(klass)
-      klass.defined_enums.each_key { |name| register_enum(Enum.new(name)) }
+      klass.defined_enums.each_key do |name|
+        next if registered_enums.any? do |registered_enum|
+          registered_enum.attribute_name.to_s == name
+        end
+        register_enum(Enum.new(name))
+      end
     end
 
     def automatically_register_defined_enums?(klass)
--- a/lib/factory_bot/definition.rb
+++ b/lib/factory_bot/definition.rb
@@ -8,7 +8,7 @@ module FactoryBot
       @declarations = DeclarationList.new(name)
       @callbacks = []
       @defined_traits = Set.new
-      @registered_enums = []
+      @registered_enums = {}
       @to_create = nil
       @base_traits = base_traits
       @additional_traits = []
@@ -86,7 +86,7 @@ module FactoryBot
     end
 
     def register_enum(enum)
-      @registered_enums << enum
+      @registered_enums[enum.attribute_name.to_sym] = enum
     end
 
     def define_constructor(&block)
@@ -164,7 +164,7 @@ module FactoryBot
         automatically_register_defined_enums(klass)
       end
 
-      registered_enums.each do |enum|
+      registered_enums.each do |_name, enum|
         traits = enum.build_traits(klass)
         traits.each { |trait| define_trait(trait) }
       end
@@ -174,10 +174,7 @@ module FactoryBot
 
     def automatically_register_defined_enums(klass)
       klass.defined_enums.each_key do |name|
-        next if registered_enums.any? do |registered_enum|
-          registered_enum.attribute_name.to_s == name
-        end
-        register_enum(Enum.new(name))
+        register_enum(Enum.new(name)) unless registered_enums.key? name.to_sym
       end
     end
 
--- a/lib/factory_bot/definition_proxy.rb
+++ b/lib/factory_bot/definition_proxy.rb
@@ -274,11 +274,18 @@ module FactoryBot
     #     those traits. When this argument is not provided, factory_bot will
     #     attempt to get the values by calling the pluralized `attribute_name`
     #     class method.
-    def traits_for_enum(attribute_name, values = nil, _prefix: false, _suffix: false, **values_as_hash)
-      enum_prefix = _prefix || (values.delete(:_prefix) if values.respond_to?(:delete))
-      enum_suffix = _suffix || (values.delete(:_suffix) if values.respond_to?(:delete))
-      values ||= values_as_hash.present? ? values_as_hash : nil
-      @definition.register_enum(Enum.new(attribute_name, values, prefix: enum_prefix, suffix: enum_suffix))
+    def traits_for_enum(attribute_name, values = nil, options = {})
+      if options.empty? && values.is_a?(Hash)
+        prefix = values.delete(:_prefix) { false }
+        suffix = values&.delete(:_suffix) { false }
+        values = nil if values.empty?
+        enum = Enum.new(attribute_name, values, prefix: prefix, suffix: suffix)
+      else
+        prefix = options.fetch(:_prefix, false)
+        suffix = options.fetch(:_suffix, false)
+        enum = Enum.new(attribute_name, values, prefix: prefix, suffix: suffix)
+      end
+      @definition.register_enum enum
     end
 
     def initialize_with(&block)
--- a/lib/factory_bot/definition_proxy.rb
+++ b/lib/factory_bot/definition_proxy.rb
@@ -221,6 +221,50 @@ module FactoryBot
     #     end
     #   end
     #
+    # Prefix and Suffixes can be used for trait definition.
+    # both can be set to true (prefix/suffix will be set to enum name), or to a custom string
+    #
+    # Example:
+    #   factory :task do
+    #     traits_for_enum :status, [:started, :finished], _prefix: true
+    #   end
+    #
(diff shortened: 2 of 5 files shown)


jasonkarns:
I came just to report this issue but it already exists, yay!

Glad to see work addressing it. Though as a user, I wouldn't be excited the current implementation requires duplicating the enum configuration in both the model and the factory. I would have expected that the factory "inherits" the enum configuration from the model itself. Otherwise, the prefix/suffix config could drift from what is defined in the model.

tjbarker:
Hey jasonkarns.

not wanting to replicate enum config in factories and models makes sense. This implementation was chosen because the flip side of explicitly setting these values would require hooking into active record which currently does not store or make available enough information to build this out.

my reasoning for thinking it wouldn’t be _that_ bad to use this is:
- the factory only needs to define the enum and prefix/suffix. If you are inheriting the enum list from the model then it will continue to inherit. Change in model enums are usually due to adding values which would automatically be added to the factory’s trait list, it’s not often that the enum attribute name or prefix/suffix change
- this will only be needed for enums with a prefix/suffix. Most enums don’t have these so will continue to be automatically added as traits to factories

lib/factory_bot/definition_proxy.rb
```diff
@@ -230,8 +274,11 @@ def trait(name, &block)
     #     those traits. When this argument is not provided, factory_bot will
     #     attempt to get the values by calling the pluralized `attribute_name`
     #     class method.
-    def traits_for_enum(attribute_name, values = nil)
-      @definition.register_enum(Enum.new(attribute_name, values))
+    def traits_for_enum(attribute_name, values = nil, _prefix: false, _suffix: false, **values_as_hash)
```
composerinteralia, line 277: Supporting both hash and keyword arguments is making this a bit complicated. I don't think we need to match the Rails `enum` method exactly here, since the use cases are a bit different (for example, I think passing the values to `traits_for_enum` as keyword arguments would be fairly uncommon, and it's also not something we've tested).

I'd rather accept a single values argument (like a hash or array), and keep the prefix and suffix keyword arguments separate. That also allows us to get rid of the ugly underscores.

So that would look like:

```rb
    def traits_for_enum(attribute_name, values = nil, prefix: false, suffix: false)
      @definition.register_enum(Enum.new(attribute_name, values, prefix: prefix, suffix: suffix))
    end
```

All of our tests still pass with this change (assuming we update the _prefix and _suffix keyword arguments to remove the underscores).

spec/acceptance/enum_traits_spec.rb
```diff
@@ -113,6 +113,211 @@ def each(&block)
         expect(task.status).to eq(trait_name)
       end
     end
+
+    context 'when prefix is used' do
+      context 'when values are a hash' do
```
composerinteralia, line 118: Do prefix and suffixes interact differently with a hash vs. list, or is there something specific we're trying to cover by testing against both? If not, I think we could choose one or the other to test with, since we've already got coverage of the various enumerable types.

lib/factory_bot/enum.rb
```diff
@@ -12,14 +23,16 @@ def build_traits(klass)
       end
     end
 
+    attr_reader :attribute_name
```
composerinteralia, line 26: Are we using this anywhere?

lib/factory_bot/definition.rb
```diff
@@ -173,7 +173,12 @@ def expand_enum_traits(klass)
     end
 
     def automatically_register_defined_enums(klass)
-      klass.defined_enums.each_key { |name| register_enum(Enum.new(name)) }
+      klass.defined_enums.each_key do |name|
+        next if registered_enums.any? do |registered_enum|
```
composerinteralia, line 177: This might be slightly easier, and have better performance when there are many enums, if we change registered enums to a hash of name -> Enum. Then checking whether it was registered would be `registered_enums.has_key?(name)`

composerinteralia (commented):
Excellent work! A few thoughts, but I think we should be able to support this.

lib/factory_bot/definition_proxy.rb
```diff
@@ -221,6 +221,50 @@ def trait(name, &block)
...
+    #   end
+    #
+    # Example:
+    #   factory :task do
+    #     traits_for_enum :status, [:started, :finished], _suffix: true
+    #   end
+    #
+    # Equivalent to:
+    #   factory :task do
+    #     traits_for_enum :status, [:started, :finished], _suffix: 'status'
```
aledustet, line 255: Might be worth pointing out we would need to update the docs here if we end up using the signature without the hash arguments.

lib/factory_bot/definition.rb
aledustet: I agree with your comment, this nested loop could loop through all enums every time.

aledustet (commented):
Nice addition! I think the change would be less "disruptive" if the new options become extra arguments there and we don't support the values as options signature but I'm excited to bring this in. Thank you!

lib/factory_bot/enum.rb
tjbarker: This was being used in `lib/factory_bot/definition.rb` to help with not redefining traits in the automatically_register_defined_enums method, it is not used in the `Definition#register_enum` method as the key in the `registered_enum` hash

lib/factory_bot/definition_proxy.rb
tjbarker: from what i can see, calling
```
statuses = {queued: 0, started: 1, finished: 2}
traits_for_enum(:status, statuses)
```
will fail with this definition
```
traits_for_enum(attribute_name, values = nil, prefix: false, suffix: false)
```
the error I'm getting is
```
ArgumentError: unknown keywords: :queued, :started, :finished
```
My understanding here is ruby is converting the hash to keyword on the way in due to the prefix/suffix keywords.

I can continue to investigate this further to get the current expectation to work (i'm assuming my understanding of ruby's keywords is the issue here, not ruby itself)

The other option I can see is having another explicit "options" hash as a third argument.
```
def traits_for_enum(attribute_name, values = nil, options = {})
  @definition.register_enum(Enum.new(attribute_name, values, prefix: options.fetch(:prefix, false), suffix: options.fetch(:suffix, false)))
end
```
which will be called as:
```
traits_for_enum :status, {queued: 0, started: 1, finished: 2}, {prefix: true}
```

any preferences which way to go?

tjbarker: the issue here is not necessarily that key names are needed, but more that a third argument is needed for setting options, but the traits_for_enum allows for values to be nil (in which case the second arg passed becomes options, not values).
if traits_for_enum is invoked with values being passed in, a third arg with options works fine, but if values is not passed in, the "options" now are the second arg, meaning values (was meant to default to nil) and options (was meant to be passed in) are both jeopardised.

i can think of 2 options here,
1: the code update I've made, which looks good from an api usage point of view, but means traits_for_enum becomes messy, as it has to try to determine on the fly what the intention of the second arg is.

2: if "options" are being used, values is a key name too (eg: `traits_for_enum(:foo, values: [:a, :b], prefix: true, suffix: true)`, which means the consumer of the api has to change their code (only when setting options, which are a new concept so all current usage continues to work) but will need to migrate over when wanting to set prefix/suffix.

spec/acceptance/enum_traits_spec.rb
tjbarker: With the current form of the code, no they don't behave differently, however with hash being passed in one of the tests (enum_for_trait_spec.rb:60) was failing as the keynames of the hash didn't match with the prefix, suffix options, so i chose to leave the 2 different tests in (for now at least) to ensure i pushed a solution that worked for all implementation types of the api

spec/acceptance/enum_traits_spec.rb
```diff
@@ -57,7 +57,7 @@
 
       FactoryBot.define do
         factory :task do
-          traits_for_enum :status, statuses
+          traits_for_enum :status, **statuses
```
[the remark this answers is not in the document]
tjbarker: updated 👍

lib/factory_bot/definition.rb
tjbarker: updated 👍

lib/factory_bot/enum.rb
tjbarker: this is now also being used in the registered_enums hash as the key to look up to ensure enums are not registered twice

PR #1514, opened 2021-09-23
Build stubbed generate uuids primary keys

### LINKED ISSUE #1498: `build_stubbed` does not fill `id` for `uuid` column type
### Description

When using `uuid` as the primary key type, they `build_stubbed` method does not automatically fill the `id` column. I would assume that this is filled independent of the column type.

### Reproduction Steps

<details>
<summary>Reproduction script</summary>

```ruby
#!/usr/bin/env ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "factory_bot", "~> 6.0"
  gem "activerecord"
  gem "pg"
end

require "active_record"
require "factory_bot"
require "minitest/autorun"
require "logger"

ActiveRecord::Base.establish_connection(adapter: "postgresql", database: "factory_bot_test")

  enable_extension 'pgcrypto'

  create_table :int_posts, force: true do |t|
    t.string :body
  end

  create_table :uuid_posts, id: :uuid, force: true do |t|
    t.string :body
  end
end

class IntPost < ActiveRecord::Base
end

class UuidPost < ActiveRecord::Base
end

FactoryBot.define do
  factory :int_post do
    body { "Post body" }
  end

  factory :uuid_post do
    body { "Post body" }
  end
end

  def test_working_with_int
    post = FactoryBot.build_stubbed(:int_post)
    refute_nil post.id
  end

  def test_failing_with_uuid
    post = FactoryBot.build_stubbed(:uuid_post)
    refute_nil post.id
  end
end

# Run the tests with `ruby <filename>`
```

</details>

### Expected behavior

When you use `build_stubbed`, the primary key is always set. Also if you use the `uuid` type.

### Actual behavior

The `id` of the generated instance is `nil`.

### System configuration
**factory_bot version**:  `6.2.0`
**rails version**:  `6.1.3.2`
**ruby version**: `3.0.1`

alexandreruban:
Closes #1498

When the primary key of a table has a type of `uuid` and when using the `build_stubbed` strategy, the `id` was not set properly.

This PR checks if the primary key of the `result_instance` is an `uuid`. If so, we use `SecureRandom.uuid` instead of an `Integer` to generate the `id` of the model.

This is my first contribution to FactoryBot so please tell me if I did anything wrong 😄

### DIFF
--- a/lib/factory_bot/strategy/stub.rb
+++ b/lib/factory_bot/strategy/stub.rb
@@ -43,13 +43,17 @@ module FactoryBot
 
       private
 
-      def next_id
-        @@next_id += 1
+      def next_id(result_instance)
+        if uuid_primary_key?(result_instance)
+          SecureRandom.uuid
+        else
+          @@next_id += 1
+        end
       end
 
       def stub_database_interaction_on_result(result_instance)
         if has_settable_id?(result_instance)
-          result_instance.id ||= next_id
+          result_instance.id ||= next_id(result_instance)
         end
 
         result_instance.instance_eval do
@@ -77,6 +81,11 @@ module FactoryBot
       def has_settable_id?(result_instance)
         !result_instance.class.respond_to?(:primary_key) ||
           result_instance.class.primary_key
+      end
+
+      def uuid_primary_key?(result_instance)
+        result_instance.respond_to?(:column_for_attribute) &&
+          result_instance.column_for_attribute(result_instance.class.primary_key).sql_type == "uuid"
       end
 
       def clear_changes_information(result_instance)
--- a/spec/acceptance/stub_spec.rb
+++ b/spec/acceptance/stub_spec.rb
@@ -59,43 +59,81 @@ describe "a stubbed instance overriding strategy" do
   end
 end
 
-describe "a stubbed instance with no primary key" do
-  it "builds a stubbed instance" do
-    using_model_without_pk do
-      FactoryBot.define do
-        factory :model_without_pk
-      end
+describe "overridden primary keys conventions" do
+  describe "a stubbed instance with a uuid primary key" do
+    it "builds a stubbed instance" do
+      using_model("ModelWithUuid", primary_key: :uuid) do
+        FactoryBot.define do
+          factory :model_with_uuid
+        end
 
-      model = FactoryBot.build_stubbed(:model_without_pk)
-      expect(model).to be_truthy
+        model = FactoryBot.build_stubbed(:model_with_uuid)
+        expect(model).to be_truthy
+      end
+    end
+
+    it "behaves like a persisted record" do
+      using_model("ModelWithUuid", primary_key: :uuid) do
+        FactoryBot.define do
+          factory :model_with_uuid
+        end
+
+        model = FactoryBot.build_stubbed(:model_with_uuid)
+        expect(model).to be_persisted
+        expect(model).not_to be_new_record
+      end
+    end
+
+    it "has a uuid primary key" do
+      using_model("ModelWithUuid", primary_key: :uuid) do
+        FactoryBot.define do
+          factory :model_with_uuid
+        end
+
+        model = FactoryBot.build_stubbed(:model_with_uuid)
+        expect(model.id).to be_a(String)
+      end
     end
   end
 
-  it "behaves like a persisted record" do
-    using_model_without_pk do
-      FactoryBot.define do
-        factory :model_without_pk
-      end
+  describe "a stubbed instance with no primary key" do
+    it "builds a stubbed instance" do
+      using_model("ModelWithoutPk", primary_key: false) do
+        FactoryBot.define do
+          factory :model_without_pk
+        end
 
-      model = FactoryBot.build_stubbed(:model_without_pk)
-      expect(model).to be_persisted
-      expect(model).not_to be_new_record
+        model = FactoryBot.build_stubbed(:model_without_pk)
+        expect(model).to be_truthy
+      end
+    end
+
+    it "behaves like a persisted record" do
+      using_model("ModelWithoutPk", primary_key: false) do
+        FactoryBot.define do
+          factory :model_without_pk
+        end
+
+        model = FactoryBot.build_stubbed(:model_without_pk)
+        expect(model).to be_persisted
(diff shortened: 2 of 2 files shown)


lib/factory_bot/strategy/stub.rb
```diff
@@ -43,13 +43,17 @@ def result(evaluation)
 
       private
 
-      def next_id
-        @@next_id += 1
+      def next_id(result_instance)
+        if result_instance.respond_to?(:column_for_attribute) && result_instance.column_for_attribute("id").sql_type == "uuid"
```
composerinteralia, line 47: I don't think `sql_type` is part of the public rails api, but I'm not sure of another way to get this.

We should us `result_instance.class.primary_key` instead of `"id"` for classes with a custom primary key.

Let's also extract this conditional into a private helper method like `uuid_primary_key?`

spec/acceptance/stub_spec.rb
```diff
@@ -59,6 +59,57 @@
...
+      FactoryBot.define do
+        factory :model_with_uuid
+      end
+
+      model = FactoryBot.build_stubbed(:model_with_uuid)
+      expect(model.id).to be_a(String)
+    end
+  end
+
+  def using_model_with_uuid
```
composerinteralia, line 97: It might be nice to update the existing helper to something more this, rather than adding a second very similar helper:

```rb
  def using_model(name, primary_key:)
    define_class(name, ActiveRecord::Base)

    connection = ActiveRecord::Base.connection
    begin
      clear_generated_table(name.tableize)
      connection.create_table(name.tableize, id: primary_key) do |t|
        t.column :updated_at, :datetime
      end

      yield
    ensure
      clear_generated_table(name.tableize)
    end
  end
```

Then we could call  `using_model("ModelWithoutPk", primary_key: false)` in the existing tests, and `using_model("ModelWithUuid", primary_key: :uuid)` in these new tests.

We'd probably have to wrap this whole file in an extra describe block to be able to share the helper method.

composerinteralia (commented):
Nice! Overall this makes sense to me.

alexandreruban:
@composerinteralia thank you very much for your review! I updated the PR according to your comments.

> I don't think sql_type is part of the public rails api, but I'm not sure of another way to get this.

Yes you are right, [ActiveRecord::ConnectionAdapters::Column](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/column.rb), [ActiveRecord::ConnectionAdapters::Postgresql::Column](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/postgresql/column.rb), and [ActiveRecord::ConnectionAdapters::MySQL::Column](https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/mysql/column.rb) that implement `sql_type` are not documented, should we still use them? I don't know of another way to get this information either.

composerinteralia:
Using private API involves some amount of risk that the private method will change unexpectedly, but we have taken that risk elsewhere in factory_bot (`defined_enums`, for example). `sql_type` seems fairly stable, so I think I'm OK with it. I'll leave this open for a bit to see if any other maintainers have different opinions. 

It's sort of awkward that `column_for_attribute` is public and returns a connection adapter column, but then none of the methods on the column are public. What good is the method if you can't do anything with the return value?

lib/factory_bot/strategy/stub.rb
```diff
@@ -79,6 +83,11 @@ def has_settable_id?(result_instance)
           result_instance.class.primary_key
       end
 
+      def uuid_primary_key?(result_instance)
+        result_instance.respond_to?(:column_for_attribute) &&
+          result_instance.column_for_attribute(result_instance.class.primary_key).sql_type == "uuid"
```
aledustet, line 88: How do we feel about adding some sort of guard here to protect us from loosing access to a private api here. Could we add a comment noting it? This feels like something future us would appreciate 😄 I could also add that the `column_for_attribute` could return nil(probably not for the primary key, but the API documents this return value), how do we feel about safe navigating after the method call to respect the spec?

aledustet (commented):
I was reading through the `column_for_attribute` after the issue was opened and thought that might be the way to go. Thanks for the addition, I have one comment non-blocking if you feel confident we should move forward I can give it the go ahead!

lib/factory_bot/strategy/stub.rb
alexandreruban: > How do we feel about adding some sort of guard here to protect us from loosing access to a private api here. Could we add a comment noting it? This feels like something future us would appreciate 😄

When I created the PR, I didn't notice that `column_for_attribute` was private API. Maybe this PR is a bit risky then, but I'm not sure there is an easy workaround. As composerinteralia said: "It's sort of awkward that column_for_attribute is public and returns a connection adapter column, but then none of the methods on the column are public. What good is the method if you can't do anything with the return value?". 

Maybe it's worth asking the Rails team? What kind of guard did you have in mind here?

> I could also add that the column_for_attribute could return nil(probably not for the primary key, but the API documents this return value), how do we feel about safe navigating after the method call to respect the spec?

I'm not sure `column_for_attribute` can return nil here. According to [the method definition](https://github.com/rails/rails/blob/581aac754c6cbe9f3f4de396bc8ef3728e56f4f4/activerecord/lib/active_record/model_schema.rb#L460), in case the column does not exist, it would return a `NullColumn` that would still respond to `sql_type` and return nil. I'm not an ActiveRecord expert, so maybe I'm missing something here 😄.

composerinteralia: [`column_for_attribute`](https://api.rubyonrails.org/classes/ActiveRecord/ModelSchema/ClassMethods.html#method-i-column_for_attribute) is publicly documented. It's only the `sql_type` method that isn't.

> in case the column does not exist, it would return a NullColumn that would still respond to sql_type and return nil

Yup, that sounds right. `sql_type` could return `nil`, in which case `nil == :uuid` would return false and we'd assume it's not a uuid. But I don't think we'd ever get that, since the stub factory checks that the model has a primary key before getting into this code.

mike-burns:
Mentioned elsewhere, but thank you @alexandreruban for kicking off this fix.

PR #1516, opened 2021-10-02
[DOC] Clarify association build strategy behavior

pboling:
It's not just factory_bot 5, but 6, and presumably 7, etc until the end of all time and the atomic scaffolding of the universe breaks down.


composerinteralia:
LOL. Let's reconsider this change once the universe begins to break down.

Issue #1503, opened 2021-06-29
Improve user feedback if defining static association with a block


grekko:
### Problem this feature will solve

Given:

```rb
FactoryBot.define do
  user(factory: :author) { role: role }
end
````

[`DefinitionProxy#method_missing`](https://github.com/thoughtbot/factory_bot/blob/907139808d728dbe8da59856afe068b7c0699b2a/lib/factory_bot/definition_proxy.rb#L91-L104) will declare the [association](https://github.com/thoughtbot/factory_bot/blob/master/lib/factory_bot/definition_proxy.rb#L97) and ignore any given `block`-argument.

I ran into this situation and wondered.

### Desired solution

`DefinitionProxy#method_missing` could warn users like so:

```rb
def method_missing(name, *args, &block) # rubocop:disable Style/MissingRespondToMissing
  association_options = args.first

  if association_options.nil?
    __declare_attribute__(name, block)
  elsif __valid_association_options?(association_options)
    Kernel.puts 'warn: Dynamically provided attributes are ignored if `factory`-key is set.' if block
    association(name, association_options)
  else
    raise NoMethodError.new(<<~MSG)
      undefined method '#{name}' in '#{@definition.name}' factory
      Did you mean? '#{name} { #{association_options.inspect} }'
    MSG
  end
end
````

**Comment by composerinteralia** [comment]
Nice idea. What if we forward the block along:

```diff
elsif __valid_association_options?(association_options)
-  association(name, association_options)
+  association(name, association_options, &block)
```

So we can get the error already defined in the association method:

https://github.com/thoughtbot/factory_bot/blob/907139808d728dbe8da59856afe068b7c0699b2a/lib/factory_bot/definition_proxy.rb#L152-L156

That's technically a breaking change, so we may need to wait until factory_bot 7.

grekko:
> So we can get the error already defined in the association method:

That look pretty clean to me 🤔 if you are interested, I could create a PR.

composerinteralia:
Yeah, that would be great. Thanks!

jeduardo824:
Hey @grekko , can I work on this issue, or are you already doing it? Thanks!

sharvy:
@composerinteralia I see grekko closed the PR. Was this feature implemented already? 
As I want to give it a try, is this still a good first issue? :)

tisba:
[...]

PR #1518, opened 2021-10-11
DefinitionProxy#method_missing: Forward given block to #association

grekko:
Addresses #1503 

## Why

Providing an options `factory`-key with an assocation definition effectively ignores a given block. Users might be confused by this behaviour, so we might want to raise an error.

## Considerations

As @composerinteralia [pointed](https://github.com/thoughtbot/factory_bot/issues/1503#issuecomment-895677669) out, this is a breaking change and might need to wait for FactoryBot 7.0.

I am not sure about the process here though. Happy to learn how this is organised.

### DIFF
--- a/lib/factory_bot/definition_proxy.rb
+++ b/lib/factory_bot/definition_proxy.rb
@@ -94,7 +94,7 @@ module FactoryBot
       if association_options.nil?
         __declare_attribute__(name, block)
       elsif __valid_association_options?(association_options)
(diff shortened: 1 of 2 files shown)


tisba:
Has this been part of a release, @aledustet?

aledustet:
> Has this been part of a release, @aledustet?

Not that I know of @tisba could it be that the fork repo was deleted, the branch closed?

PR #1519, opened 2021-10-13
Cache gem dependencies in CI

aried3r:
Around 20 seconds is spent on installing gems, using gem caching it should make the tests even faster.


composerinteralia:
It looks like this caused a bunch of test failures. Is this something you are still interested in (I know it's been a long time), or shall I close this PR for now?

mike-burns:
This seems to be incompatible with Appraisal. I'm not currently feeling the pain, so I'm going to close this. Thank you for trying!

PR #1523, opened 2021-12-16
Document 6.0.0 may break `FactoryBot.lint`

choznerol:
In a sufficiently large codebase, the automatic definition of traits for Active Record enum  (#1380) may result in more combinations being checked by `FactoryBot.lint`. This breaking change may not be obvious because the first impression of the error messages may be "these factories were passing before the upgrade but now failing ?", while the truth is that they don't exist before the upgrade. It took me a while to realize this so I hope adding this note could help people be aware of this situation when upgrading.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -28,7 +28,9 @@
   * Added: automatic definition of traits for Active Record enum attributes, enabled by default
     (Note that this required changing where factory_bot constantizes the build
      class, which may affect applications that were using abstract factories for
-     inheritance. See issue #1409.)
+     inheritance. See issue #1409.) (This may break `FactoryBot.lint` because
+     there may be previously non-existing factory+trait combinations being
+     defined and checked)
   * Added: `traits_for_enum` method to define traits for non-Active Record enums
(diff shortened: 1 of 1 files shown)


Azfletch (approved):

NEWS.md
```diff
@@ -21,7 +21,7 @@
   * Added: automatic definition of traits for Active Record enum attributes, enabled by default
     (Note that this required changing where factory_bot constantizes the build
      class, which may affect applications that were using abstract factories for
-     inheritance. See issue #1409.)
+     inheritance. See issue #1409.) (This may break `FactoryBot.lint` because there may be previously non-existing factory+trait combinations being defined and checked)
```
calebhearth, line 24: This should be wrapped to 80 chars to match the rest of this file.

calebhearth (changes requested):

NEWS.md
choznerol: I have amended and push 710dd7adaa520b1c057a93d31b380baea59e2bc6 instead

mike-burns:
Thanks!

PR #1526, opened 2022-01-14
Add Ruby 3.1 and Rails 7 to CI

petergoldstein:
This PR adds the Ruby 3.1 and Rails 7 elements to the CI matrix.  To support those changes a number of other changes were required:

- Adding a basic Rails 7 Gemfile and updating `Appraisals`
- Quoting 3.0 in the `.github/workflows/build.yml` file so that this entry pulls Ruby 3.0.x and not Ruby 3.1
- Adding the "--disable-error_highlight" value for RUBYOPT in the build environment to disable Ruby 3.1 error highlighting
- Bumping the version of `standard` to pick up a Ruby 3.1 compatible Rubocop version
- Disabling the `standard` run for `bundle exec rake` for Rails version >= 3.1.0 because of a conflicting evaluation for `Style/MethodMissingSuper` for Ruby 3.1 and earlier Rubies


.github/workflows/build.yml
```diff
@@ -24,24 +24,30 @@ jobs:
...
+            rails: "6.1"
         exclude:
           - ruby: 2.5
             rails: main
           - ruby: 2.6
             rails: main
           - ruby: jruby-9.2.16.0
             rails: main
-          - ruby: 3.0
+          - ruby: "3.0"
```
composerinteralia, line 39: I'm not sure I understand why we need quotes here but don't for Ruby 2.5 and 2.6

.github/workflows/build.yml
```diff
@@ -24,24 +24,30 @@ jobs:
           - "6.0"
           - "6.1"
           - main
+        include:
```
composerinteralia, line 27: I'd prefer to keep the existing pattern of adding the new Rails and Ruby versions to the matrix section above, then excluding the combinations we don't want. That has generally made it slightly easier to add new versions of things.

.github/workflows/build.yml
```diff
@@ -24,24 +24,30 @@ jobs:
...
             rails: "5.1"
-          - ruby: 3.0
+          - ruby: "3.0"
             rails: "5.2"
 
     runs-on: 'ubuntu-latest'
 
     env:
       BUNDLE_GEMFILE: gemfiles/${{ matrix.rails }}.gemfile
+      RUBYOPT: "--disable-error_highlight"
```
composerinteralia, line 50: I'm curious what error you were seeing that led you to add this. Is everyone running workflows with Ruby 3.1 going to need this?

Appraisals
```diff
@@ -28,6 +28,12 @@ appraise "6.1" do
   gem "sqlite3", platforms: [:ruby]
 end
 
+appraise "6.1" do
```
composerinteralia, line 31:
```suggestion
appraise "7.0" do
```

spec/acceptance/build_stubbed_spec.rb
```diff
@@ -167,7 +167,7 @@
   it "allows overriding created_at for objects with created_at" do
     created_at = 3.days.ago
     stubbed = build_stubbed(:thing_with_timestamp, created_at: created_at)
-    expect(stubbed.created_at).to eq created_at
+    expect(stubbed.created_at).to be_within(1.second).of created_at
```
composerinteralia, line 170: Has something changed in Rails that was causing this test to fail? Are we getting back a different precision after the database round trip, perhaps?

Rakefile
```diff
@@ -9,7 +9,10 @@ require "standard/rake"
 Bundler::GemHelper.install_tasks(name: "factory_bot")
 
 desc "Default: run the specs, features, and standard ."
-task default: %w[spec:unit spec:acceptance features standard]
+
+default_tasks = %w[spec:unit spec:acceptance features]
+default_tasks << "standard" if RUBY_VERSION < "3.1.0"
```
composerinteralia, line 14: My concern with this is that as we drop support for older Ruby versions we'll eventually stop running standard unless we remember to change this.

I think we can avoid this if we add a `.standard.yml` with:

```rb
ruby_version: "2.5"
```

and then get rid of the references to the outdated cop by running:

```rb 
bundle exec rake standard:fix
```

We'd probably still want to update the `.standard.yml` as we drop versions of Ruby, but even if we forget to do that at least we're guaranteed to have some version of standard running in CI.

composerinteralia (commented):
Thanks! It may be slightly easier to add Rails and Ruby separately, since there are quite a few changes here.

.github/workflows/build.yml
composerinteralia: After running the workflow I see we're missing a bunch of combinations, so all the more reason to stick to the existing matrix approach.

.github/workflows/build.yml
berkos: 👋  As @petergoldstein mentioned in the PR description it is using 3.1 otherwise.
I believe it is because of this https://github.com/actions/runner/issues/849
It is also mentioned here https://github.com/ruby/setup-ruby#matrix-of-ruby-versions

spec/acceptance/build_stubbed_spec.rb
petergoldstein: @composerinteralia Yes, this is a precision issue.  We get a difference in the milliseconds as I recall from running the specs.  During the round trip it appears to be rounded to 0.

.github/workflows/build.yml
petergoldstein: You can find the discussion here - https://github.com/rspec/rspec-expectations/issues/1338 .

And here's what it looked like for `factory_bot`:

Screen Shot 2022-01-14 at 10 44 47 AM

petergoldstein: As you can see in `rspec-expectations` I submitted a PR to resolve this issue in matchers that was just merged today.

Rakefile
petergoldstein: Ok, I've done this.

Appraisals
petergoldstein: This is fixed.

.github/workflows/build.yml
petergoldstein: @berkos Yep, that's it.

composerinteralia: 😭 Thank you for the explanation. I saw that in the PR description, but didn't understand why.

.github/workflows/build.yml
composerinteralia: 👍🏻 Thank you for explaining, and for fixing the underlying problem!

petergoldstein:
@composerinteralia Ok, this is all green now.  Let me know if you still want to split it into two different PRs.

PR #1530, opened 2022-02-04
Fix strategy tracking through associations

joshuaclayton:
What?
=====

Within FactoryBot::Evaluator, the build strategy was reported either as a
class OR symbol.

A side-effect of this is misreporting within ActiveSupport::Notifications
hooks, since the instrumentation payload may have strategies listed as e.g.
`:create` or
`FactoryBot::Strategy::Create`, even though they semantically represent the
same information.

This introduces a new instance method for all strategies, `to_sym`, to be
called rather than `class`.


lib/factory_bot/strategy/attributes_for.rb
```diff
@@ -8,6 +8,10 @@ def association(runner)
       def result(evaluation)
         evaluation.hash
       end
+
+      def to_sym
+        :attributes_for
+      end
```
aledustet, line 14: How do we feel about adding a constant to each strategy, like:
``` ruby
module Strategy
  class AttributesFor
  # ...
     SIMBOLYZED = :attributes_for
  # ...
  end
end

module Strategy
    included do
      def to_sym
        self.class::SIMBOLYZED
      end
    end
  end
```

aledustet (approved):
Nice consolidation! I have one proposal about the execution, non-blocking. I understand why this might be problematic and I like the approach.

qrush:
+1 to merge this 🙏🏼

PR #1532, opened 2022-03-08
Attempt to cache bundler

joshuaclayton:


joshuaclayton:
This flag does not appear to work correctly based on our current configuration (perhaps related to appraisal + multiple versions of Ruby/Rails)?

Issue #1536, opened 2022-04-01
NoMethodError: undefined method 'to_sym' for CustomStrategy


rocket-turtle:
### Description

After upgrading from Version 6.2.0 to 6.2.1 our custom strategy was broken.
-> NoMethodError: undefined method 'to_sym' for CustomStrategy

This commit broke our code https://github.com/thoughtbot/factory_bot/commit/1b81d5dc258e5f1ad25c5109cb4bf3f6be83deb4

The change is not mentioned in the NEWS.rb nor in the GETTING_STARTED.md

https://www.rubydoc.info/gems/factory_bot/file/GETTING_STARTED.md#custom-strategies

### Solution

If someone comes here with the same Problem it is super easy to fix. Just add the `to_sym` to your strategy. :)

-------

### Question

Could someone explain to me or point me in the direction where I can find more information why the custom strategies should be defined like this and not like the one underneath. There is no problem i'm just curious.

```
class JsonStrategy
  def initialize
    @strategy = FactoryBot.strategy_by_name(:create).new
  end

  delegate :association, to: :@strategy

  def result(evaluation)
    @strategy.result(evaluation).to_json
  end

  def to_sym
    :json
  end
end
```

```
class JsonStrategy < FactoryBot::Strategy::Create
  def result(evaluation)
    super.to_json
  end

  def to_sym
    :json
  end
end
```

EdmundKorley:
Great catch! I've noted this breaking change in https://github.com/thoughtbot/factory_bot/commit/fc9bb59e434aaa5b6b96eb41015dcb093c4bfc33 for now. It seems like the now required `to_sym` on a custom strategy didn't get documented but we would really love an PR / update to the [GETTING_STARTED] guide.

For your second question, there isn't really a particular reason rather favoring "composition over inheritance". When you subclass from a framework's internal class, then users custom code can be tightly coupled to such undocumented changes in the framework code like what happened here.

[GETTING_STARTED]: https://github.com/thoughtbot/factory_bot/blob/main/GETTING_STARTED.md#custom-strategies

bkDJ:
This also breaks a custom strategy that inherits from an existing one, though the error will not mention `to_sym` since the wrong strategy will be used instead (a project I work on had `ReadOnly < Create` to add records to what is normally a read-only database).

Adding `to_sym` fixed it for us as well. 👍

chris-at-work:
[...]

PR #1537, opened 2022-04-04
Add 'to_sym' to CustomStrategy in GETTING_STARTED

rocket-turtle:
See https://github.com/thoughtbot/factory_bot/issues/1536


composerinteralia (approved):
Thanks! It might be nice to have some default implementation so this isn't necessary for every custom strategy, but in the meantime this documentation change matches what we've currently got.

PR #1541, opened 2022-06-03
Update getting_started guide, use_parent_strategy

DC00:
### Description

From my understanding, based on the [acceptance spec here](https://github.com/thoughtbot/factory_bot/blob/main/spec/acceptance/build_spec.rb#L33-L39), `use_parent_strategy` should be `true` if we _do not_ want to save associations.

Should the corresponding example in the [build strategies section](https://github.com/thoughtbot/factory_bot/blob/main/GETTING_STARTED.md#build-strategies-1) of the Getting Started guide be changed to `FactoryBot.use_parent_strategy = true` ?

### System configuration
**factory_bot version**:  
**rails version**:  
**ruby version**:

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -649,7 +649,7 @@ post.author.new_record? # => false
 To not save the associated object, specify `strategy: :build` in the factory:
 
 ```ruby
-FactoryBot.use_parent_strategy = false
+FactoryBot.use_parent_strategy = true
 
 factory :post do
   # ...


GETTING_STARTED.md
```diff
@@ -649,7 +649,7 @@ post.author.new_record? # => false
 To not save the associated object, specify `strategy: :build` in the factory:
 
 ```ruby
-FactoryBot.use_parent_strategy = false
+FactoryBot.use_parent_strategy = true
```
composerinteralia, line 652: Yeah, I think you are correct that something is off here. I think we could actually get rid of this line entirely, since `strategy: :build` takes precedence regardless of how `use_parent_strategy` is set.

PR #1544, opened 2022-07-13
Ensure factory options are a Hash

leoarnold:
I accidentally typed

```ruby
factory :thing, Thing do
  # ...
end
```

and was confused by the error message.

This commit ensures that the options are a Hash,
and gives an informative error message otherwise.

### DIFF
--- a/lib/factory_bot/factory.rb
+++ b/lib/factory_bot/factory.rb
@@ -7,8 +7,10 @@ module FactoryBot
     attr_reader :name, :definition
 
     def initialize(name, options = {})
-      assert_valid_options(options)
       @name = name.respond_to?(:to_sym) ? name.to_sym : name.to_s.underscore.to_sym
+
+      assert_valid_options(options)
+
       @parent = options[:parent]
       @aliases = options[:aliases] || []
       @class_name = options[:class]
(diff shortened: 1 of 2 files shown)


Azfletch (approved):

lib/factory_bot/factory.rb
```diff
@@ -140,6 +142,8 @@ def compiled_constructor
     private
 
     def assert_valid_options(options)
+      raise ArgumentError, "Expected a Hash of options for factory '#{name.inspect}', got '#{options.inspect}'" unless options.is_a?(Hash)
```
calebhearth, line 145: Unless things have changed, thoughtbot prefer 80 char lines and to avoid conditional surprises

```suggestion
      if !options.is_a?(Hash)
        raise ArgumentError,
          "Expected a Hash of options for factory '#{name.inspect}', got '#{options.inspect}'"
      end
```

spec/factory_bot/factory_spec.rb
```diff
@@ -185,6 +185,12 @@ def self.to_s
   end
 end
 
+describe FactoryBot::Factory, "when options are not a Hash" do
+  it "raises an ArgumentError" do
+    expect { FactoryBot::Factory.new(:author, Class.new) }.to raise_error(ArgumentError, /^Expected a Hash of options for factory ':author', got '#<Class:/)
```
calebhearth, line 190:
```suggestion
    expect do
      FactoryBot::Factory.new(:author, Class.new)
    end.to raise_error(
      ArgumentError,
      /^Expected a Hash of options for factory ':author', got '#<Class:/
    )
```

calebhearth (changes requested):

PR #1546, opened 2022-08-07
rename references master -> main

mockdeep:
This renames all references to `master` to `main`. The `master` branch
is no longer up to date, so these links go to outdated docs. There was
one in `.gitignore` that I removed entirely, as it looks like the
`.lock` files in the `gemfiles/` directory are intentionally checked in.


composerinteralia:
Thanks!

PR #1561, opened 2023-01-02
Pass key and receiver to custom key errors

composerinteralia:
did_you_mean tests started failing on Ruby 3.2 because of https://bugs.ruby-lang.org/issues/18564.

This gets did_you_mean working again for Ruby 3.2 by adding the original KeyError's key and receiver over to the custom KeyError. That way the did_you_mean can get added to the custom KeyError via the new detailed_message.

Once this is merged we can add Ruby 3.2 to the test matrix

### DIFF
--- a/lib/factory_bot/definition.rb
+++ b/lib/factory_bot/definition.rb
@@ -115,15 +115,26 @@ module FactoryBot
       raise error_with_definition_name(error)
     end
 
-    def error_with_definition_name(error)
-      message = error.message
-      message.insert(
-        message.index("\nDid you mean?") || message.length,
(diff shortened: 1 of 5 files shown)


composerinteralia:
truffleruby against Rails 7.0 is [failing without these changes](https://github.com/thoughtbot/factory_bot/actions/runs/3824964346/jobs/6507565826) as well—something related to simplecov, I think. We can fix that in a separate PR.

composerinteralia:
Going to merge for now so I can get a passing Ruby 3.2 build, but happy to address any review comments later if anybody else looks at this.

PR #1563, opened 2023-01-03
Don't run simplecov on alternate Ruby

composerinteralia:
The truffleruby + Rails 7.0 build was failing because of some kind of simplecov incompatibility. I don't think we need to run simplecov on the truffleruby and jruby builds (we might not need it on any build—it prints some output but doesn't affect the build status at all—but I figured I'd leave it alone for now in case somebody is looking at it).

I needed to add a fileutils require to one test that was presumably getting implicitly required before via simplecov.

### DIFF
--- a/features/support/env.rb
+++ b/features/support/env.rb
@@ -1,6 +1,6 @@
 PROJECT_ROOT = File.expand_path(File.join(File.dirname(__FILE__), "..", ".."))
 
-require "simplecov"
+require "simplecov" if RUBY_ENGINE == "ruby"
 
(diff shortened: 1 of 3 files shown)


composerinteralia:
cc @eregon in case there's anything interesting to you here. Here's an [example build failure](https://github.com/thoughtbot/factory_bot/actions/runs/3824964346/jobs/6507565826).

eregon:
It's https://github.com/simplecov-ruby/simplecov/issues/1044.
Indeed, there is no need to run coverage on other Rubies than CRuby.

PR #1571, opened 2023-04-18
Major docs overhaul: refs, guide, cookbook

mike-burns:
Split the docs into reference, guide, and cookbook.

The guide is mostly as it was before, with a lot of cleanup. Some sections were combined and simplified.

Some guides were moved into a new cookbook section. This section describes techniques that are helpful in specific situations of using factory_bot but not, in itself, part of factory_bot. For example, using factory_bot to simulate a has_many relation.

Other guides were moved into the wiki. The criteria for this was: does it have to do with factory_bot, or with something external? Will we change the docs alongside a change in this repo's code, or due to another repo's code change?

A new reference section is added. It lists terse facts about methods you are expected to use. The reason RubyDoc isn't the right solution here is because these methods are defined all over the codebase. Due to `instance_eval` and mixins, factory_bot is more of a DSL than a set of Ruby classes. Thus, the reference section.


docs/src/ref/add_attribute.md
```diff
@@ -0,0 +1,27 @@
...
+
+The `add_attribute` method takes two arguments: a name (Symbol or String) and a
+block. This block is called each time this object is constructed. The block is
+not called when the attribute is overriden by a build strategy.
+
+Assignment is done by calling the Ruby attribute setter. For example, given
+
+```ruby
+FactoryBot.define do
+  factory :user
```
seanpdoyle, line 14: Is this code sample missing a `do`?

docs/src/ref/build-and-create.md
```diff
@@ -0,0 +1,39 @@
...
+## initialize_with
+
+The `initialize_with` method takes a block and returns an instance of the
+factory's class. It has access to the `attributes` method, which is a hash of
+all the fields and values for the object.
+
+The default definition is:
+
+```ruby
+initialize_with { new }
```
seanpdoyle, line 19: I believe that an `attributes` method is available within this block. Is the default block value `new(attributes)`?

docs/src/callbacks/summary.md
```diff
@@ -1 +1,19 @@
...
+* before(:create) - called before a factory is saved  (via `FactoryBot.create`)
+* after(:create)  - called after a factory is saved   (via `FactoryBot.create`)
+* after(:stub)    - called after a factory is stubbed (via `FactoryBot.build_stubbed`)
+
+Examples:
+
+```ruby
+# Define a factory that calls the generate_hashed_password method after it is built
+factory :user do
+  after(:build) { |user, evaluator| generate_hashed_password(user) }
```
seanpdoyle, line 15: I'm not sure if this is the point in time you're interested in feedback about verbiage, but I've found the usage of `evaluator` to be vague and somewhat confusing.

Anecdotally, we've been naming this second block argument as `factory`, and that has felt more self-documenting.

docs/src/ref/build-and-create.md
mike-burns: FB new-s up a blank object and then assigns to it using setters. This is perfect for AR, and bad for most other objects.

docs/src/ref/build-and-create.md
```diff
@@ -0,0 +1,39 @@
+# skip_create, to_create, and initialize_with
+
+The `skip_create`, `to_create`, and `initialize_with` methods control how
+factory\_bot interacts with the [build strategies](build-strategies.html).
+
+These methods can be called within a `factory` definition block, to scope their
+effects to just that factory; or within `FactoryBot.define`, to effect global
```
sej3506, line 7:
```suggestion
effects to just that factory; or within `FactoryBot.define`, to affect global
```

docs/src/ref/lint.md
```diff
@@ -0,0 +1,16 @@
...
+
+The `FactoryBot.lint` method tries each factory and raises
+`FactoryBot::InvalidFactoryError` on failure.
+
+It can take the following optional arguments:
+
+- A splat of factory names. This will restrict the linting to just the ones listed. The default is all.
+- `:strategy` - the [build strategy] to use. The default is `:create`.
+- `:traits` - whether to try building each trait, too. The default is `false`.
+- `:verbose` - whether to show a stack trace on error. The default is `flase`.
```
sej3506, line 11:
```suggestion
- `:verbose` - whether to show a stack trace on error. The default is `false`.
```

docs/src/ref/method_missing.md
```diff
@@ -0,0 +1,67 @@
...
+an association, sequence, trait, or attribute to the factory:
+
+1. If the method missing is passed a block, it always defines an attribute.
+   This allows you to set the value for the attribute.
+
+1. If the method missing is passed a hash as a argument with the key
+   `:factory`, then it always defines an association. This allows you to
+   override the factory used for the association.
+
+1. If there in another factory of the same name, then it defines an
```
sej3506, line 17:
```suggestion
1. If there is another factory of the same name, then it defines an
```

docs/src/ref/modify.md
```diff
@@ -0,0 +1,13 @@
+# FactoryBot.modify
+
+The `FactoryBot.modify` class method defines a block with an _overriding_
+`factory` method available. That is the only method you can call within the
+block.
+
+The `factory` method within this block takes a mandatory factory name, and a
+block. All other arguments are ignored. The factory name must already be
+defined. The block is a normal [factory definition block](factory.html). Take
+notes that [hooks](hooks.html) cannot be cleared and continue to compound.
```
sej3506, line 10:
```suggestion
note that [hooks](hooks.html) cannot be cleared and continue to compound.
```

docs/src/ref/sequence.md
```diff
@@ -0,0 +1,51 @@
...
+A factory\_bot set up supports two levels of sequences: global and factory-specific.
+
+## Global sequences
+
+With a [`Factory.define`] block, use the `sequence` method to define global
+sequences that can be shared with other factories.
+
+[`Factory.define`]: define.html
+
+The `sequence` method takes a name, optional arguments, and a block. The name is expected to be a symbol.
```
sej3506, line 12: Nitpick: I've noticed sometimes "symbol" is capitalized and sometimes not.

docs/src/ref/traits_for_enum.md
```diff
@@ -0,0 +1,57 @@
...
+
+The `traits_for_enum` method takes a required attribute name and an optional
+set of values. The values can be any Enumerable, such as Array or Hash. By
+default, the values are `nil`.
+
+If the values are an Array, this method defines a trait for each element in the
+array. The trait's name is the array element, and it sets the attribute to the
+same array element.
+
+If the value values are a Hash, this method defines traits based on the keys,
```
sej3506, line 20:
```suggestion
If the values are a Hash, this method defines traits based on the keys,
```

docs/src/transient-attributes/summary.md
```diff
@@ -3,3 +3,16 @@
 Transient attributes are attributes only available within the factory
 definition, and not set on the object being built. This allows for more complex
 logic inside factories.
+
+Thse are defined using within a `transient` block:
```
sej3506, line 7:
```suggestion
These are defined within a `transient` block:
```

docs/src/callbacks/summary.md
```diff
@@ -1 +1,19 @@
...
+
+* after(:build)   - called after a factory is built   (via `FactoryBot.build`, `FactoryBot.create`)
+* before(:create) - called before a factory is saved  (via `FactoryBot.create`)
+* after(:create)  - called after a factory is saved   (via `FactoryBot.create`)
+* after(:stub)    - called after a factory is stubbed (via `FactoryBot.build_stubbed`)
+
+Examples:
+
+```ruby
+# Define a factory that calls the generate_hashed_password method after it is built
```
stefannibrasil, line 13: Does the following suggestion read better for you as well?

```suggestion
# Define a factory that calls the generate_hashed_password method after the user factory is built
```

docs/src/transient-attributes/summary.md
```diff
@@ -3,3 +3,16 @@
...
+
+```ruby
+factory :user do
+  name { "Zero Cool" }
+  dob { age&.years.ago }
```
stefannibrasil, line 12: Fun fact: once I had to fill a form in English and I had no idea what `dob` meant. I didn't fill in and the person explained it meant "date of birth" 😆

docs/src/summary.md
```diff
@@ -0,0 +1,34 @@
...
+Its documentation is split as such:
+
+- the [guide](setup/summary.html) is a great place to start for first-time users.
+- the [cookbook](cookbook/has_many-associations.html) is the go-to place for creative solutions to common situations.
+- the [wiki](https://github.com/thoughtbot/factory_bot/wiki) details considerations for integrating with other software.
+- the [reference](ref/build-strategies.html) is terse facts for those who use this project often.
+
+## License
+
+factory_bot is Copyright © 2008-2022 Joe Ferris and thoughtbot. It is free
```
stefannibrasil, line 17: Do we need to keep the final year here? Otherwise, we would need to update things every year 🤔 Not sure how these things work, though.

stefannibrasil (commented):
Nice! Thanks for working on this 🎉 

I've left a couple of comments.

One thing that feels a bit weird to me is using “guide”, “reference”, etc. They give me the impression there’s only one guide and one reference. Have you considered using them in the plural form?

docs/src/callbacks/summary.md
mike-burns: I'm quite open to verbiage!

My hesitation about `factory` is that we also have a method with the same name. How bad would it be to call it `definition` instead?

```ruby
factory :user do
  after(:build) { |user, definition| ... }
end
```

mike-burns: After a heated discussion in #dev , I've recognized that the best name for this parameter is `context`. Going to make that change. Holler if you hate it!

sej3506 (approved):
I appreciated the Hackers references 😎

mike-burns:
Looks like GitHub closed the PR before it merged it. Wrong order, GitHub!

Anyway this is merged now.

PR #1574, opened 2023-05-12
Link to the book

mike-burns:
Thank you for your years of service, `GETTING_STARTED.md`. You're in a better place now.

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -1,3 +1,19 @@
+**Deprecated**
+
+See our extensive reference, guides, and cookbook in [the factory_bot book][].
+
+For information on integrations with third party libraries, such as RSpec or
+Rails, see [the factory_bot wiki][].
+
+ We also have [a detailed introductory video][], available for free on Upcase.
+
+[a detailed introductory video]: https://upcase.com/videos/factory-bot?utm_source=github&utm_medium=open-source&utm_campaign=factory-girl
+[the factory_bot book]: https://thoughtbot.github.io/factory_bot
+[the factory_bot wiki]: https://github.com/thoughtbot/factory_bot/wiki
+
+This document is deprecated and preserved for historical use. It may disappear
+at any time.
+
 Getting Started
 ===============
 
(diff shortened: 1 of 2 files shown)


sej3506 (approved):
LGTM 👍🏻

jasonkarns:
I'm a little worried that the docbook is going to make finding things harder.

As a realistic example from a few seconds ago, we searched the getting-started file for invocations of `association` to see how to provide a trait. Cmd-F `association` found all examples in the page.

The same approach in the docbook is broken because of the book's multipage format. 😞

mike-burns:
@jasonkarns that's my biggest concern, too.

Does the book search handle that? https://thoughtbot.github.io/factory_bot/?search=association

jasonkarns:
@mike-burns perhaps, but TBQH, i'm not going to know (or look for) bespoke search functionality in each doc site of the multiple and various tools we use. Cmd+F is ubiquitous.

I don't really have good suggestions :/ overriding Cmd+F to use custom site search is fraught with peril. but ignoring it and requiring one use site-specific search is also a frustrating UX.

And in most cases with FB in particular, i don't want a list of search results. i want to bounce _very_ quickly from result to result. jumping back and forth between a result and the search results listing would be... not good (even presuming one used the site-search).

I guess I'm just being salty that my cheese was moved. :D

jasonkarns:
So now i'm just carrying on, but i did see the new docbook site has a print view, which is one giant page.

If i'm making nitpicky requests, it would be fantastic if that single-page view were linked somewhere. As it stands, hitting the print icon and dismissing the browser print dialog isn't too terrible. i'll probably just bookmark that single-page url. But making that page a "first class" option seems like a decent compromise?

mike-burns:
Good call on the print view, I'll add some more links to it so people can jump straight there.

If [mdBook ever adds single-page output](https://github.com/rust-lang/mdBook/issues/988), I'll generate that from CI so we can have that option.

---

Also, thanks for understanding. I feel the same frustrations you do regarding Ruby docs (Rails, factory_bot, and other special cases, plus rdoc for each library ...) and I hope we as a community can find a better way forward.

PR #1576, opened 2023-05-25
fix(Stub): align #destroyed with #new_record/#persisted

bf4:
I was surprised to find that #destroyed? returns nil
instead of false when I build_stubbed.  I don't see
a reason for it in the commit history nor in Rails,
so I thought it was fair game for this PR.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,10 @@
 # News
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


mike-burns:
Thanks for aligning it with ActiveRecord!

PR #1578, opened 2023-06-07
Warn about viewing docs on old `master` branch

thoughtbot-summer:
Since [the deprecation of `GETTING_STARTED.md`](https://github.com/thoughtbot/factory_bot/pull/1574), I've gotten better at noticing when people online link to it on the old `master` branch. That version does not have the deprecation notice linking to the new docs, nor does it have any changes made in the past couple years.

This PR targets the `master` branch and adds a warning to the top of each Markdown file with a link to it on the `main` branch. It doesn't directly link to the news docs, which is deliberate so that we don't potentially have to update the `master` branch again in the future (if we move the docs again, for example).

If this PR is merged, I plan to make a similar change in [factory_bot_rails](https://github.com/thoughtbot/factory_bot_rails).


mike-burns (approved):
Good call, thanks for adding this to all the relevant places.

(Also I just noticed that we haven't renamed to `main`. Adding that to my to-do list.)

PR #1583, opened 2023-06-24
Support uuids in build_stubbed

### LINKED ISSUE #1498: `build_stubbed` does not fill `id` for `uuid` column type

jaynetics:
This is based on #1514 and fixes #1498.

The only things i did, compared to #1514, are the following:

- rebased on main
- added one more guard clause, to ensure that the worst-case result if the relevant rails API is ever removed is the same lack of UUID support that we have currently

The relevant rails API hasn't changed since PR #1514 was created 2 years ago. It would be nice to finally get support for UUIDs. 😊

### DIFF
--- a/lib/factory_bot/strategy/stub.rb
+++ b/lib/factory_bot/strategy/stub.rb
@@ -47,13 +47,17 @@ module FactoryBot
 
       private
 
-      def next_id
-        @@next_id += 1
+      def next_id(result_instance)
+        if uuid_primary_key?(result_instance)
+          SecureRandom.uuid
+        else
+          @@next_id += 1
+        end
       end
 
       def stub_database_interaction_on_result(result_instance)
         if has_settable_id?(result_instance)
-          result_instance.id ||= next_id
+          result_instance.id ||= next_id(result_instance)
         end
 
         result_instance.instance_eval do
@@ -81,6 +85,13 @@ module FactoryBot
       def has_settable_id?(result_instance)
         !result_instance.class.respond_to?(:primary_key) ||
           result_instance.class.primary_key
+      end
+
+      def uuid_primary_key?(result_instance)
+        result_instance.respond_to?(:column_for_attribute) &&
+          (column = result_instance.column_for_attribute(result_instance.class.primary_key)) &&
+          column.respond_to?(:sql_type) &&
+          column.sql_type == "uuid"
       end
 
       def clear_changes_information(result_instance)
--- a/spec/acceptance/stub_spec.rb
+++ b/spec/acceptance/stub_spec.rb
@@ -59,43 +59,81 @@ describe "a stubbed instance overriding strategy" do
   end
 end
 
-describe "a stubbed instance with no primary key" do
-  it "builds a stubbed instance" do
-    using_model_without_pk do
-      FactoryBot.define do
-        factory :model_without_pk
-      end
+describe "overridden primary keys conventions" do
+  describe "a stubbed instance with a uuid primary key" do
+    it "builds a stubbed instance" do
+      using_model("ModelWithUuid", primary_key: :uuid) do
+        FactoryBot.define do
+          factory :model_with_uuid
+        end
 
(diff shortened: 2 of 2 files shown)


mike-burns:
Right, and we're going to need this now that [we've outlawed setting the `id` explicitly](https://github.com/thoughtbot/factory_bot_rails/pull/419).

The thing I've been staring at all afternoon, though, is: this is all very ActiveRecord-specific, which means it should go in `factory_bot_rails` instead. But I think the original sin was adding this stub strategy to the non-Rails repo in the first place.

OK, I think the plan is: merge this PR, then work on moving the stub strategy into f_b_rails. That will require a lot of finesse and delicacy because I don't want to break any existing code in the process.

Thanks for listening as I talk this through.

Hey @jaynetics @alexandreruban and @StefSchenkelaars , thank you all for the work you put into this!

mike-burns (approved):

mike-burns:
(Current status: figuring out why CI isn't running. _sigh_)

MatheusRich:
@mike-burns might be related https://github.com/thoughtbot/factory_bot/pull/1593

mike-burns:
Thanks for being on top of things, @MatheusRich! But if that's the case, why aren't those CI checks running?

mike-burns:
I pushed this branch to the main repo and watched CI pass: https://github.com/thoughtbot/factory_bot/actions/runs/6592716225

Going to merge bypassing CI for this PR.

jaynetics:
@mike-burns thanks for merging!

> this is all very ActiveRecord-specific, which means it should go in factory_bot_rails instead. But I think the original sin was adding this stub strategy to the non-Rails repo in the first place.
> OK, I think the plan is: merge this PR, then work on moving the stub strategy into f_b_rails.

if you want to support `build_stubbed` for ORMs other than `activerecord` in the future, a consistent solution might be easier to achieve if its all in the same base repo.

either way, `f_b_rails` doesn't seem quite the right place for it because some people use `activerecord` without rails (e.g. on `sinatra`), and some people use rails without `activerecord` (e.g. using `sequel` instead). currently, `f_b_rails` crashes if rails is not present (despite having only `activerecord`, not `rails`, in its gemspec dependencies), so moving features to there might affect a few people.

mike-burns:
Yeah fair - perhaps the actual original problem was splitting out a Rails-specific gem to begin with, with all the vagueness that "Rails" implies.

PR #1590, opened 2023-08-28
Chore - Add link to the changelog 

brkn:
[Ruby gems](https://rubygems.org/gems/factory_bot) doesn't show the link to the changelog.
I thought this was especially important since the filename is unconventional, it would speed up the workflow of this gem's users.


mike-burns (approved):

mike-burns:
Thanks!

PR #1592, opened 2023-09-01
Document integration with 2.7 and 3.0 syntactic sugars

seanpdoyle:
Ruby 2.7 introduced [numbered block parameters][], and 3.0 introduced [Hash literal value omission][], so document how they integrate with Factory Bot.

[numbered block parameters]: https://ruby-doc.org/core-2.7.1/Proc.html#class-Proc-label-Numbered+parameters
[Hash literal value omission]: https://docs.ruby-lang.org/en/3.1/syntax/literals_rdoc.html#label-Hash+Literals

### DIFF
--- a/spec/acceptance/attributes_for_spec.rb
+++ b/spec/acceptance/attributes_for_spec.rb
@@ -1,3 +1,7 @@
+if RUBY_VERSION >= "3.0" && RUBY_ENGINE != "truffleruby"
+  require_relative "./attributes_for_destructuring"
+end
+
 describe "a generated attributes hash" do
   include FactoryBot::Syntax::Methods
 
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -285,6 +285,9 @@ user = create(:user)
 # Returns a hash of attributes that can be used to build a User instance
 attrs = attributes_for(:user)
 
+# Integrates with Ruby 3.0's support for pattern matching assignment
+attributes_for(:user) => {email:, name:, **attrs}
+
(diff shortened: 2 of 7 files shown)


spec/acceptance/attributes_for_spec.rb
```diff
@@ -1,3 +1,7 @@
+if RUBY_VERSION >= "3.0"
```
seanpdoyle, line 1: Without this conditional, Ruby 2.7 will fail in CI.

docs/src/ref/build-strategies.md
```diff
@@ -58,6 +58,18 @@ The `FactoryBot.attributes_for` method constructs a Hash with the attributes
 and their values, using `initialize_with`. The `attributes_for_pair` and
 `attributes_for_list` methods work similarly as `build_pair` and `build_list`.
 
+```ruby
```
mike-burns, line 61: This might fit better into `docs/src/using-factories/build-strategies.md`, since that documents at a higher level than the files under `ref/` do.

mike-burns (approved):
Love it, thanks for these docs!

spec/acceptance/attributes_for_spec.rb
```diff
@@ -1,3 +1,7 @@
+if RUBY_VERSION >= "3.0" && RUBY_ENGINE != "truffleruby"
```
seanpdoyle, line 1: @mike-burns CI forced me to skip Truffle Ruby. Since this is only introducing test coverage, one that implementation adopts the latest language features support will kick-in without additional implementation changes on our end.

PR #1599, opened 2023-11-17
Pass block to `expect` as needed

mike-burns:
RSpec's `expect` has two forms:

- scalar: `expect(result).to be_truthy`
- block: `expect { thunk }.not_to raise_error(RuntimeError)`

So what happens when you mix them?

```ruby
thunk = -> { 1 / 0 }
expect(thunk).not_to raise_error(RuntimeError)
```

If you do that, RSpec complains:

> The implicit block expectation syntax is deprecated, you should pass a
> block rather than an argument to `expect` to use the provided block
> expectation matcher or the matcher must implement
> `supports_value_expectations?`

The easy solution is to pass the Proc as a block, using `&`:

```
thunk = -> { 1 / 0 }
expect(&thunk).not_to raise_error(RuntimeError)
```

### DIFF
--- a/spec/acceptance/defining_methods_inside_a_factory_spec.rb
+++ b/spec/acceptance/defining_methods_inside_a_factory_spec.rb
@@ -12,7 +12,7 @@ describe "defining methods inside FactoryBot" do
       end
     end
 
-    expect(bad_factory_definition).to raise_error(
+    expect(&bad_factory_definition).to raise_error(
       FactoryBot::MethodDefinitionError,
       /Defining methods in blocks \(trait or factory\) is not supported \(generate_name\)/
     )
--- a/spec/acceptance/modify_factories_spec.rb
+++ b/spec/acceptance/modify_factories_spec.rb
@@ -179,6 +179,6 @@ describe "modifying factories" do
       end
     end
 
-    expect(modify_unknown_factory).to raise_error(KeyError)
+    expect(&modify_unknown_factory).to raise_error(KeyError)
   end
 end
--- a/spec/factory_bot/definition_proxy_spec.rb
+++ b/spec/factory_bot/definition_proxy_spec.rb
@@ -68,7 +68,7 @@ describe FactoryBot::DefinitionProxy, "#method_missing" do
     proxy = FactoryBot::DefinitionProxy.new(definition)
     invalid_call = -> { proxy.static_attributes_are_gone "true" }
 
-    expect(invalid_call).to raise_error(
+    expect(&invalid_call).to raise_error(
       NoMethodError,
       "undefined method 'static_attributes_are_gone' in 'broken' factory\n" \
       "Did you mean? 'static_attributes_are_gone { \"true\" }'\n"
--- a/spec/support/matchers/raise_did_you_mean_error.rb
+++ b/spec/support/matchers/raise_did_you_mean_error.rb
@@ -13,6 +13,6 @@ RSpec::Matchers.define :raise_did_you_mean_error do
       raise_error(KeyError, /Did you mean\?/)
     end
 
-    expect(actual).to matcher
+    expect(&actual).to matcher
   end
 end


MatheusRich (approved):

thoughtbot-summer (approved):

seanpdoyle:
For my own understanding, what does `&` do in this scenario?

For example, reading `&:to_s` aloud, I'd describe it as something like "calling Symbol to_proc on to_s".

With that in mind, I explored transforming the `&` coercion syntax into an explicit method call (specifically, a call to `#to_proc`):

```diff
diff --git a/spec/acceptance/defining_methods_inside_a_factory_spec.rb b/spec/acceptance/defining_methods_inside_a_factory_spec.rb
index f7e19bf..01fe7af 100644
--- a/spec/acceptance/defining_methods_inside_a_factory_spec.rb
+++ b/spec/acceptance/defining_methods_inside_a_factory_spec.rb
@@ -12,7 +12,7 @@ describe "defining methods inside FactoryBot" do
       end
     end
 
-    expect(&bad_factory_definition).to raise_error(
+    expect(bad_factory_definition.to_proc).to raise_error(
       FactoryBot::MethodDefinitionError,
       /Defining methods in blocks \(trait or factory\) is not supported \(generate_name\)/
     )
diff --git a/spec/acceptance/modify_factories_spec.rb b/spec/acceptance/modify_factories_spec.rb
index 737097b..2652f71 100644
--- a/spec/acceptance/modify_factories_spec.rb
+++ b/spec/acceptance/modify_factories_spec.rb
@@ -179,6 +179,6 @@ describe "modifying factories" do
       end
     end
 
-    expect(&modify_unknown_factory).to raise_error(KeyError)
+    expect(modify_unknown_factory.to_proc).to raise_error(KeyError)
   end
 end
diff --git a/spec/factory_bot/definition_proxy_spec.rb b/spec/factory_bot/definition_proxy_spec.rb
index 0a291a6..b8d6502 100644
--- a/spec/factory_bot/definition_proxy_spec.rb
+++ b/spec/factory_bot/definition_proxy_spec.rb
@@ -68,7 +68,7 @@ describe FactoryBot::DefinitionProxy, "#method_missing" do
     proxy = FactoryBot::DefinitionProxy.new(definition)
     invalid_call = -> { proxy.static_attributes_are_gone "true" }
 
-    expect(&invalid_call).to raise_error(
+    expect(invalid_call.to_proc).to raise_error(
       NoMethodError,
       "undefined method 'static_attributes_are_gone' in 'broken' factory\n" \
       "Did you mean? 'static_attributes_are_gone { \"true\" }'\n"
diff --git a/spec/support/matchers/raise_did_you_mean_error.rb b/spec/support/matchers/raise_did_you_mean_error.rb
index bc9a85c..26d936f 100644
--- a/spec/support/matchers/raise_did_you_mean_error.rb
+++ b/spec/support/matchers/raise_did_you_mean_error.rb
@@ -13,6 +13,6 @@ RSpec::Matchers.define :raise_did_you_mean_error do
       raise_error(KeyError, /Did you mean\?/)
     end
 
-    expect(&actual).to matcher
+    expect(actual.to_proc).to matcher
   end
 end
```

Unfortunately, that didn't address the underlying warnings. What additional steps does `&` take to communicate the argument as a block?

mike-burns:
Good question, and that sent me into the C. Here's [the relevant part](https://github.com/ruby/ruby/blob/master/vm_insnhelper.c#L515-L534):

```c
VALUE
rb_vm_bh_to_procval(const rb_execution_context_t *ec, VALUE block_handler)
{
    if (block_handler == VM_BLOCK_HANDLER_NONE) {
        return Qnil;
    }
    else {
        switch (vm_block_handler_type(block_handler)) {
          case block_handler_type_iseq:
          case block_handler_type_ifunc:
            return rb_vm_make_proc(ec, VM_BH_TO_CAPT_BLOCK(block_handler), rb_cProc);
          case block_handler_type_symbol:
            return rb_sym_to_proc(VM_BH_TO_SYMBOL(block_handler));
          case block_handler_type_proc:
            return VM_BH_TO_PROC(block_handler);
          default:
            VM_UNREACHABLE(rb_vm_bh_to_procval);
        }
    }
}
```

(For those following along at home, I traced through from `parse.y` (find for `'&'`), to `node.h`, `compile.c`, `insns.def`, until I got to this function.)

So we can see that there are quite a few special cases for the `&` syntax:

- if it's a low-level `RubyVM::InstructionSequence` or ifunc, do some magic;
- if it's `&:`, call `&to_proc`;
- if it's a Proc, pass the C block handler through.

Since we're passing a Proc, we're passing the block handler through. This means no conversion, it just passes it.

So `&` simply marks the argument as a block argument in the parser and AST.

```
irb(main):020:0> def foo(a = nil)
irb(main):021:1>   if a
irb(main):022:2>     puts "got arg"
irb(main):023:2>   end
irb(main):024:1>   if block_given?
irb(main):025:2>     puts "got block"
irb(main):026:2>   end
irb(main):027:1> end
=> :foo
irb(main):028:0> arg = -> { 1 / 0 }
=> #<Proc:0x00007f922782ef90 (irb):28 (lambda)>
irb(main):029:0> foo(arg)
got arg
=> nil
irb(main):030:0> foo(&arg)
got block
=> nil
```

PR #1600, opened 2023-11-20
Assign klass to defined_traits to pass it's class to ActiveSuport::Notifications

makicamel:
#1587 published `factory_bot.compile_factory notification`.
That maked to be able tracking the attributes and traits, e.g. https://github.com/thoughtbot/factory_bot_rails/pull/419 .
But when using trait the payload has no class, so assign definition's class to defined_traits.

Fix https://github.com/thoughtbot/factory_bot_rails/issues/431

PR #1601, opened 2023-11-21
Assign klass to global traits when find by name

makicamel:
We fixed the error that `defined_traits` had no class in #1600 .
But we can declare traits also in top level and mixin them as `base_traits` and `additional_traits`, like:

```ruby
FactoryBot.define do
  # top level trait
  trait :email do
    email { "#{name}@example.com" }
  end

  factory :school, traits: [:email] do
    name { "Charlie" }
  end

  factory :company do
    name { "Charlie" }
  end
end

# base_traits
FactoryBot.build(:school)
# additional_traits
FactoryBot.build(:company, :email)
```

The traits declared at the top level have their class determined when they are used, so assign the definition klass when the traits are used to pass their class to ActiveSuport::Notifications.
Re Fix https://github.com/thoughtbot/factory_bot_rails/issues/431

### DIFF
--- a/lib/factory_bot/definition.rb
+++ b/lib/factory_bot/definition.rb
@@ -52,6 +52,7 @@ module FactoryBot
 
         declarations.attributes
 
+        self.klass ||= klass
         defined_traits.each do |defined_trait|
           defined_trait.klass ||= klass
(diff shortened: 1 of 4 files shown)


mike-burns:
Thank you for tracking this down! Released as 6.4.2.

PR #1603, opened 2023-12-08
Support models without ID setters in build_stubbed

olivier-thatch:
Some "models" may not have an ID setter at all. This PR ensures that `build_stubbed` only tries to set the ID if the model does have an ID setter (i.e. does have an `#id=` method).

### DIFF
--- a/lib/factory_bot/strategy/stub.rb
+++ b/lib/factory_bot/strategy/stub.rb
@@ -83,8 +83,9 @@ module FactoryBot
       end
 
       def has_settable_id?(result_instance)
-        !result_instance.class.respond_to?(:primary_key) ||
-          result_instance.class.primary_key
+        result_instance.respond_to?(:id=) &&
+          (!result_instance.class.respond_to?(:primary_key) ||
+          result_instance.class.primary_key)
       end
 
       def uuid_primary_key?(result_instance)
--- a/spec/acceptance/stub_spec.rb
+++ b/spec/acceptance/stub_spec.rb
@@ -121,6 +121,17 @@ describe "overridden primary keys conventions" do
     end
   end
 
+  describe "a stubbed instance with no id setter" do
+    it "builds a stubbed instance" do
+      FactoryBot.define do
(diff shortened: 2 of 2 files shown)


mike-burns:
Is the fundamental problem that we're trying to set `id` specifically, and instead we should set whatever the primary key is?

```ruby
pk = result_instance.primary_key
if result_instance.send(pk).blank?
  result_instance.send("#{pk}=", next_id(result_instance))
end
```

olivier-thatch:
That's a good point! I can update the PR to do that instead.

olivier-thatch:
Having thought about it a little more, I think these are 2 related but not quite overlapping issues:
- the instance being built may or may not have an ID setter method (e.g. when building a `Hash`)
- if it exists, the ID setter method is not necessarily named `#id=`

This PR fixes the former. I think the latter should be addressed separately. @mike-burns wdyt? I'm happy to submit a followup PR for the latter issue once this is merged. Or if you prefer I can update this PR to address both issues at once.

mike-burns:
That's a useful breakdown, thanks!

Is there a reason to call `build_stubbed` on a Hash?

olivier-thatch:
> Is there a reason to call build_stubbed on a Hash?

Not directly, but we ran into this bug when calling `build_stubbed` on a regular ActiveRecord model which itself calls a factory on a `Hash` for one of its properties.

mike-burns:
Great answer.

I won't be able to cut a release until Friday, so up to you whether you want this PR to just be checking for `#id=`, or checking for the primary key setting.

olivier-thatch:
After doing a bit more research, for Active Record models, the setter is always named `#id=` even if a custom column name is used for the primary key: https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/PrimaryKey.html#method-i-id-3D. So for Active Record models at least, no additional work is needed!

mike-burns:
Thanks for doing that research! I've merged this so we can try to solve your problem faster.

PR #1607, opened 2023-12-25
Fix Ruby 3.3 deprecation warning

olegantonyan:
Fix this warning in Ruby 3.3
```
warning: observer was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add observer to your Gemfile or gemspec
```

PR #1610, opened 2023-12-26
Unconditionally declare observer as a dependency

Earlopain:
I believe #1607 wasn't sufficient. After upgrading the factory_bot gem my lockfile still doesn't include `observer` as a dependency and the warning is still triggered.

This can also be confirmed by looking at its rubygems page https://rubygems.org/gems/factory_bot

I'm not certain on the reason for this but I do know that condition dependencies like this aren't great. I believe it may have something to do with the ruby version being used during release.

Either way, doing this should fix it. I don't think there should be any issues with declaring it like this. Many gems did it the same, see https://github.com/rails/rails/pull/48907 for example.

### DIFF
--- a/factory_bot.gemspec
+++ b/factory_bot.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
   }
 
   s.add_dependency("activesupport", ">= 5.0.0")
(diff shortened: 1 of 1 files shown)


mike-burns:
Closing in favor of #1611. Thanks for exploring that!

Earlopain:
Great, much appreciated

PR #1625, opened 2024-02-19
Add standard settings

ydah:
The following settings have been added.
* Set parallel execution to true
* Change format to progress

### DIFF
--- a/.standard.yml
+++ b/.standard.yml
@@ -1,1 +1,3 @@
 ruby_version: "3.0"
+parallel: true
+format: progress


mike-burns:
## Documentation

- `parallel` passes down to Rubocop, and it [does](https://docs.rubocop.org/rubocop/usage/basic_usage.html#command-line-flags) this:
    > Use available CPUs to execute inspection in parallel.
- `progress` also passes down to Rubocop, and [prints](https://docs.rubocop.org/rubocop/formatters.html#progress-formatter-default) a `.` for each clean file.

Both of these are the Rubocop defaults. Standard overrides the defaults:

- `parallel` interacts poorly with `stdin` mode. As such, it is [forcibly overridden](https://github.com/standardrb/standard/commit/facc8ede91eb6248028dcc966cb94dcd5ba4fb11) in `stdin` mode, and the default is not to use it.
- `progress` doesn't mention standard, which seems to be why the `Standard::Formatter` [was introduced](https://github.com/standardrb/standard/commit/cf1ccb2b3d982900cd0b7ee08cb00c621c953766#diff-aeb2669c278690a295d637dd1738bcb603670fca460f3815af587aec8df07f91).

## Analysis

I don't even use Standard, but it runs as part of CI. Sure, I'll merge these changes, and if someone complains I'll back them out.

PR #1629, opened 2024-03-21
Update has_many-associations.md

mzagaja:
I struggled to figure out how to make the has_many association work with the build strategy in a simple way. I asked ChatGPT and it came up with this strategy, which I tested and worked on an app last night. I also looked in my past code and realized years ago I figured this out previously in other apps I have written. I imagine its a common pattern.

It would be handy to be able to look up this code snippet for me in the repetitive but not common case I need to do this. I find it easier to reason about than the other examples in the documentation, it works well, and can save the need to persist to the database. I also would prefer to be able to look this up in official documentation and not an AI assistant.

### DIFF
--- a/docs/src/cookbook/has_many-associations.md
+++ b/docs/src/cookbook/has_many-associations.md
@@ -68,9 +68,9 @@ create(:user_with_posts).posts.length # 5
 create(:user_with_posts, posts_count: 15).posts.length # 15
 ```
 
-Or, for a solution that works with `build`, `build_stubbed`, and `create`
-(although it doesn't work well with `attributes_for`), you can use inline
-associations:
+The following is a simple example that works without having to save to the
+database. It works with `build`, `build_stubbed`, and `create` (although it
+doesn't work well with `attributes_for`), you can use inline associations:
 
 ```ruby
 FactoryBot.define do
@@ -83,6 +83,12 @@ FactoryBot.define do
     name { "Taylor Kim" }
 
     factory :user_with_posts do
+      posts { [association(:post)] }
+    end
+
+    # or
+
+    trait :with_posts do
       posts { [association(:post)] }
     end
   end


docs/src/cookbook/has_many-associations.md
```diff
@@ -68,7 +68,20 @@ create(:user_with_posts).posts.length # 5
...
+    trait :with_completed_survey_visit do
+      after(:build) do |home, evaluator|
+        survey_visit = build(:survey_visit, home: home) #belongs_to association
+        # Below line is required for the association to work from home.survey_visits without saving to the database
+        # Otherwise it only works in one direction: survey_visit.home
+        home.survey_visits << survey_visit # has_many association
+      end
+    end
+  end
+```
```
purinkle, line 82: What do you think of the following? I tested the definition on my machine and it works without saving to the database.

```suggestion

```ruby
    trait :with_completed_survey_visit do
       survey_visits { [association(:survey_visit)] }
    end
  end
```
```

purinkle (commented):
@mzagaja, thank you for contributing. I want to understand the intent of the documentation change. You can already do the same by [using a nested factory][]. Do you intend to document the use of a trait? I'm focusing on merging documentation changes this week, so I'll be around 😊

[using a nested factory]: https://github.com/thoughtbot/factory_bot/blob/5f9e369c980f11ef3bfbce989739b3b1b3332da3/docs/src/cookbook/has_many-associations.md?plain=1#L85-L87

docs/src/cookbook/has_many-associations.md
mzagaja: @purinkle That looks good to me. My approach/example is a bit verbose.

mzagaja:
@purinkle the use of the nested factory, while present, was not entirely obvious as the solution to me when I was skimming/searching the docs for this, so I think the majority of the value of my contribution comes from the short description I added on 71, but I do also like having it as a trait mostly because it more closely follows the mental model of how I use/build factories.After you pointed it out, I did like your more concise approach/example to building out the association versus what I've been doing.

purinkle:
@mzagaja, I condensed the two examples into one. What do you think?

mzagaja:
@purinkle I like it 😄

PR #1632, opened 2024-03-31
Add `require 'factory_bot_rails'` in GETTING-STARTED instructions

salmanasiddiqui:
Following readme results in the code below

rails_helper.rb:
```ruby
# This file is copied to spec/ when you run 'rails generate rspec:install'
require 'spec_helper'
ENV['RAILS_ENV'] ||= 'test'
require_relative '../config/environment'
# Prevent database truncation if the environment is production
abort('The Rails environment is running in production mode!') if Rails.env.production?
require 'rspec/rails'
# Add additional requires below this line. Rails is not loaded until this point!
require 'support/factory_bot'

...
```

support/factory_bot.rb:
```ruby
RSpec.configure do |config|
  config.include FactoryBot::Syntax::Methods
end
```

This results in an error:

```ruby
uninitialized constant `FactoryBot`
```

because factory_bot gem has not bene loaded yet. So I required it in factory_bot.rb to fix the issue.

support/factory_bot.rb:
```ruby
require 'factory_bot_rails'

RSpec.configure do |config|
  config.include FactoryBot::Syntax::Methods
end
```

This PR updates the instructions in Getting-Started accordingly

### DIFF
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -124,7 +124,7 @@ If you're using Rails, add the following configuration to
 `rails_helper.rb`:
 
 ```ruby
-require 'factory_bot'
+require 'factory_bot_rails'
 
 RSpec.configure do |config|
   config.include FactoryBot::Syntax::Methods
@@ -134,6 +134,8 @@ end
 If you're *not* using Rails:
 
 ```ruby
+require 'factory_bot'
+
 RSpec.configure do |config|
   config.include FactoryBot::Syntax::Methods
 
--- a/GETTING_STARTED.md
+++ b/GETTING_STARTED.md
@@ -124,6 +124,8 @@ If you're using Rails, add the following configuration to
 `rails_helper.rb`:
 
 ```ruby
+require 'factory_bot'
+
 RSpec.configure do |config|
   config.include FactoryBot::Syntax::Methods
 end


sarahraqueld:
Hi @salmanasiddiqui, I was not able to reproduce your issue. Following the README works well for me. Can you add more details, like what version of Factory Bot are you using? Did you add the gem to the Gemfile correctly and ran `bundle`?

OughtPuts:
Hey @salmanasiddiqui - thanks for this PR :) would you mind please adding exactly what you ran before the error was thrown to make it easier for us to try to reproduce? Thanks in advance!

stratigos:
I dont think this solution is appropriate as it forces `factory_bot` to no longer be framework agnostic, and instead forces integration with Rails. I do understand that nearly everyone using the gem is using it in a Rails app, but it appears to be designed specifically to be used in any Ruby application, Rails or not.

The current README file does point to the wiki at the top, and the first section of the wiki is integrations with 3rd parties (Rails, Rspec...). [The RSpec guide document](https://github.com/thoughtbot/factory_bot/wiki/Integration%3A-RSpec#find_definitions) does state that `factory_bot_rails` is necessary for this integration to make use of the DSL in the way the examples show (though technically, not required).

I also think the current README and GETTING_STARTED guides are clear enough in specifying that _if_ one is integrating with Rails/RSpec, and one wants to use the DSL as specified in examples, then `factory_bot_rails` needs included. It does also state that a developer can opt to do this shim with plain old Ruby on their own. 

I think `factory_bot` is healthiest when presenting a default case agnostic of Rails, and continues to express that there are gems to assist with Rails/RSpec integration.

PR #1630, opened 2024-03-25
Update README again

stefannibrasil:
There's a rule for when a branch is merged for factory_bot. I tried removing `[ci-skip]` from the original workflow to trigger all CI actions. It didn't work, still got the same error:

```
remote: error: GH006: Protected branch update failed for refs/heads/main.
remote: error: Required status check "Run standard" is expected.
To https://github.com/thoughtbot/factory_bot.git
 ! [remote rejected] main -> main (protected branch hook declined)
```

Gonna try going through the PR workflow to see if the rule gets happy.

PR #1635, opened 2024-04-12
Only run this workflow if the README has been updated

stefannibrasil:
In the templates repo, the update-templates workflow has been updated to accommodate repositories that have main branch rules (like factory_bot). To generate the README dynamically, you’ll need to merge a PR with the README changes. That will trigger the GitHub workflow

The PR won’t be generated if the footer content does not have any differences from the template. The PR will only be opened if there’s an update to the footer snippet.

In this PR, we are limiting this workflow to be triggered only when README changes have been merged to main (where the snippet lives).

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ software, and may be redistributed under the terms specified in the
 [LICENSE]: https://github.com/thoughtbot/factory_bot/blob/main/LICENSE
 
 <!-- START /templates/footer.md -->
+
 <!-- END /templates/footer.md -->
 
 [ci-image]: https://github.com/thoughtbot/factory_bot/actions/workflows/build.yml/badge.svg?branch=main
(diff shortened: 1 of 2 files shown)


stefannibrasil:
Thanks for your patience, @mike-burns this is the follow up of https://github.com/thoughtbot/factory_bot/pull/1630

In summary, some repos have branch protection rules that conflicted with the workflow pushing changes directly to the main branch. In the templates repo, the workflow has been updated to open a PR with the README updates. So it will follow the regular flow of PR -> merge. Thank you!

README.md
```diff
@@ -82,6 +82,7 @@ software, and may be redistributed under the terms specified in the
 [LICENSE]: https://github.com/thoughtbot/factory_bot/blob/main/LICENSE
 
 <!-- START /templates/footer.md -->
+
```
stefannibrasil, line 85: With this change, the workflow will generate an automatic PR with the generated content.

mike-burns (approved):
Thank you for working around the commit limitations! Let's try it out.

Issue #1647, opened 2024-05-10
Ad hoc sequences in create_list


mike-burns:
This comes up all the time. Here's an example from a real test:

```ruby
prompts = [
  create(:prompt, name: "Prompt 1"),
  create(:prompt, name: "Prompt 2"),
  create(:prompt, name: "Prompt 3"),
]
```

I'd love to be able to write something else. Here's an example, maybe:

```ruby
prompts = FactoryBot.create_list(
            :prompt,
            3,
            name: FactoryBot.build_sequence { |n| "Prompt #{n}" },
          )
```

Note that `create_list` already takes a block:

> This block takes the produced object and the zero-based index as arguments, and returns an updated object.

But we want to get in there before it produces the object.

PR #1639, opened 2024-04-28
Add support for before build callback

mohammednasser-32:
Fixes https://github.com/thoughtbot/factory_bot/issues/1633

add support for `before(:build)` callback as requested in the issue

Tested the changes by adding the local gem to a project
image
image

I am honestly not sure how to add unit tests for the change, any help would be appreciated

PR #1650, opened 2024-05-12
Add hoc sequences in create_list

mohammednasser-32:
Fixes https://github.com/thoughtbot/factory_bot/issues/1647

Allow creation of sequence attributes in `create_list`

This PR is more of a proof of concept, if I get the green light I can add documentation and more tests if needed.

It is working and all tests pass, however I am not confident about the code structure..in this approach `FactoryBot.build_sequence` is just a syntactical sugar for a `Proc`, however it will work fine if we pass a `Proc` too. Is it better to have a dedicated class for this instead?


mohammednasser-32:
Yes right didn't consider that, thanks for raising it!
[here](https://github.com/thoughtbot/factory_bot/pull/1650/commits/213bd29839d2826bc7dbc98d622c769bf8bf5b1b) is another approach where I created `FactoryBot::AttributeSequence` class which encapsulates the proc so we can differentiate between a normal proc and an attribute sequence one

mohammednasser-32:
Hello @mike-burns! I can see you are no longer maintaining this repository, Can you please tell me who should I follow up regarding this (and the other [PR](https://github.com/thoughtbot/factory_bot/pull/1639))
(I am really sorry for being obtrusive, I am just keen to contribute here :sweat_smile: )

mike-burns:
You were quick to notice!

We're still collecting maintainers, but these people will be in charge: https://github.com/thoughtbot/factory_bot/pull/1651/files

May 23, 2024 08:18:44 Mohammed Nasser ***@***.***>:

smaboshe:
> Hello @mike-burns! I can see you are no longer maintaining this repository, Can you please tell me who should I follow up regarding this (and the other [PR](https://github.com/thoughtbot/factory_bot/pull/1639)) (I am really sorry for being obtrusive, I am just keen to contribute here 😅 )

We're steadily getting up to speed 😅. Thanks in advance for your patience, @mohammednasser-32.

spec/acceptance/create_list_spec.rb
```diff
@@ -38,6 +38,16 @@
     end
   end
 
+  context "with sequencial attributes" do
+    subject { FactoryBot.create_list(:post, 20, title: FactoryBot.build_sequence{ |n| "title_#{n}" }) }
+
+    it "create sequencial attribute values" do
```
smaboshe, line 44:
```suggestion
    it "creates sequential attribute values" do
```

spec/acceptance/create_list_spec.rb
```diff
@@ -38,6 +38,16 @@
     end
   end
 
+  context "with sequencial attributes" do
```
smaboshe, line 41:
```suggestion
  context "with sequential attributes" do
```

lib/factory_bot/attribute_sequence.rb
```diff
@@ -0,0 +1,24 @@
...
+  # @api private
+  class AttributeSequence
+    attr_reader :expression
+
+    def initialize(&block)
+      @expression = block
+    end
+
+    def evaluate(index)
+      expression.call(index)
```
smaboshe, line 11: I am curious to know what is the result if `expression` is `nil`?

smaboshe:
What do you think about adding an additional test for the default behaviour. Something like:

```
context "with an unspecified sequential attribute" do 
    subject { FactoryBot.create_list(:post, 20, title: FactoryBot.build_sequence ) }

    it "creates sequential attribute values" do
      subject.each_with_index do |record, i|
        expect(record.title).to eq ????
      end
    end
  end
```
Credit: @gkosmo 

Open to a discussion about what the default behaviour should ideally be 🙂.

mohammednasser-32:
Hey @smaboshe , thanks for the review :pray: 
yeah I missed this case, I would say we have one of 3 options
- raise an error
- return the index
- return the default value
They all seem valid so I will apply whatever you believe fits the best :smile:

mohammednasser-32:
@smaboshe @gkosmo looking forward to hear your opinions on this so I can apply it :pray:

PR #1653, opened 2024-05-22
Fix build_stubbed timestamps being different

kimesf:
### Description

Ensures build_stubbed generates a record with equal created_at and updated_at.

### Current behavior

Calling `.build_stubbed` on factories with entities that have created_at and updated_at assigns a slightly different time for each one of then because the code calls `Time.current` twice.

### Expected behavior

I think the expected behavior should be for them to be exact the same.

### Concerns

There are two concerns with the current solution:

1. We would be calling `Time.current` even if created_at and updated_at don't exist on the entity;
2. ~~It might be a problem that we're assigning the same object to both timestamps, should we `dup` updated_at?~~ (edit: dates are immutable, never mind :grimacing: :face_with_peeking_eye:).

I preferred to ask if these are valid concerns before investigating how to solve them.

### DIFF
--- a/lib/factory_bot/strategy/stub.rb
+++ b/lib/factory_bot/strategy/stub.rb
@@ -102,12 +102,14 @@ module FactoryBot
       end
 
       def set_timestamps(result_instance)
+        timestamp = Time.current
+
         if missing_created_at?(result_instance)
-          result_instance.created_at = Time.current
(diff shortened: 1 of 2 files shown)


neilvcarvalho (approved):
Looks good. Thank you for your contribution!

PR #1674, opened 2024-07-19
Allow the initial value of a sequence to be lazy loaded

ajesler:
## Allow the initial value of a sequence to be evaluated on first use

In order to be able to use factory_bot to generate seeds for a database where the database might not have the schema set up yet or might have data already populated, it is useful to be able to defer assigning the initial value to a sequence until it is needed. 
Currently, factory_bot needs the initial value of the sequence to be defined when the factory file is loaded.

This lazy loading behaviour has been very handy for use when we need to create objects that have unique sequential id in formats specific to third party systems. We can use these factories in our development and CI environments without having to think about which values have already been used.

I found https://github.com/thoughtbot/factory_bot/pull/1434 and thought it was worth submitting this PR.

This PR adds a `lazy` option to the sequence method, where a Proc can be passed that will be called the first time as sequence is used, to set the initial value. 
After the initial `#next` call, these sequences behave exactly the same as any other factory_bot sequence.
Rewinding a lazy sequence uses the originally calculated value, it does not re-call the lazy Proc object.

Example usage:

```ruby
FactoryBot.define do
  sequence(:email, lazy: -> {  User.count + 1 }) { "user#{_1}@example.com" }
end

generate(:email) # user42@example.com if you already have 41 users
```

The way we initially solved this was to pass in a `LazyEnumeratorAdapter` instance (as defined in the PR) as the initial value, eg 
`sequence(:name, LazyEnumeratorAdaper.new {  User.count }) { "User#{_1}" }`. 

This can be done without any changes to factory_bot, but I thought this functionality might be useful to others. 

I'm happy to remove the `lazy` option and instead do detection on whether the initial arg is a Proc object, but I like that the `lazy` arg name sets expectations about when this will be evaluated. Although it does result in the need for a whole extra class.

### DIFF
--- a/docs/src/ref/sequence.md
+++ b/docs/src/ref/sequence.md
@@ -13,7 +13,7 @@ The `sequence` method takes a name, optional arguments, and a block. The name
 is expected to be a Symbol.
 
 The supported arguments are a number representing the starting value (default:
-`1`), and `:aliases` (default `[]`). The starting value must respond to `#next`.
+`1`), `:aliases` (default `[]`), and `lazy` (default:`nil`). The starting value must respond to `#next`.
 
 The block takes a value as an argument, and returns a result.
 
--- a/docs/src/sequences/initial-value.md
+++ b/docs/src/sequences/initial-value.md
@@ -8,3 +8,19 @@ factory :user do
   sequence(:email, 1000) { |n| "person#{n}@example.com" }
 end
 ```
+
+The initial value can also be lazily set by passing a Proc object via the `lazy` option. This Proc will be called the first time the `sequence.next` is called.
+
+```ruby
+factory :user do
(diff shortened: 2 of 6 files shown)


ajesler:
Hi @neilvcarvalho,

No worries, thanks for following up.

I had some time tonight so took a crack at updating the PR to work with the `main` branch.
Instead of messing with this PR, I decided to open a new one, https://github.com/thoughtbot/factory_bot/pull/1761. 

It looks like the rewinding sequence specs have been refactored and included as part of `spec/factory_bot/sequence_spec.rb`, so I dropped the changes that were in `spec/acceptance/sequence_resetting_spec.rb`. 

Use/close either PR as you wish.

vburzynski (commented):
~~Overall looks good. I added a few suggestions 🙂~~
wrong browser tab 🙃

PR #1622, opened 2024-01-30
Bump minimum required ruby version to 3.0

Earlopain:
See #1612
Closes #1614

I would suggest against yanking the affected version. Doing so will result in installs for those that already upgraded not working anymore so you get another disruption for free. It's been a month since that release and it would have been better to do this as early as possible instead.

Issue #1614, opened 2024-01-01
Factory Bot 6.4.5: Syntax error with Ruby < 2.7, but the ruby gem requirement is still `>= 2.5.0`


tagliala:
### Description

Factory Bot 6.4.5 contains syntax which is not compatible with Ruby < 2.7, but the ruby gem requirement is still `>= 2.5.0`

Probably introduced in: #1612

I understand that the intent is to drop Ruby < 3.0, but the requirement in the Gemfile stays the same.

I would suggest to yank 6.4.5 and release 6.4.6 with the minimum ruby version at 3.0

Ref: 
- https://github.com/thoughtbot/factory_bot/blob/main/factory_bot.gemspec#L18
- https://github.com/diowa/ruby2-rails5-bootstrap-heroku/actions/runs/7375030629/job/20066130892#step:6:7

### Reproduction Steps

Run the following script:

```rb
# frozen_string_literal: true

require 'bundler/inline'

begin
  gemfile(true) do
    source 'https://rubygems.org'

    gem 'factory_bot', '6.4.5'
  end

  puts "\nRuby version: #{RUBY_VERSION}"
rescue Gem::LoadError => e
  puts "\nMissing Dependency:\n#{e.backtrace.first} #{e.message}"
rescue LoadError => e
  puts "\nError:\n#{e.backtrace.first} #{e.message}"
  puts DATA.read
  exit 1
end

require 'factory_bot'
```

### Expected behavior

No raise

### Actual behavior

```
~/.rvm/gems/ruby-2.5.9/gems/factory_bot-6.4.5/lib/factory_bot.rb:25:in `require': ~/.rvm/gems/ruby-2.5.9/gems/factory_bot-6.4.5/lib/factory_bot/evaluator.rb:38: syntax error, unexpected ... (SyntaxError)
...method_missing(method_name, ...)
...                            ^~~
~/.rvm/gems/ruby-2.5.9/gems/factory_bot-6.4.5/lib/factory_bot/evaluator.rb:40: syntax error, unexpected ..., expecting ')'
...@instance.send(method_name, ...)
...                            ^~~
~/.rvm/gems/ruby-2.5.9/gems/factory_bot-6.4.5/lib/factory_bot/evaluator.rb:42: syntax error, unexpected ..., expecting ')'
...unner.new.send(method_name, ...)
...                            ^~~
[...]

PR #1687, opened 2024-08-13
Bump the minimum required activesupport version to 6.1

Earlopain:
They were dropped from CI in #1612. Also see #1614 and #1622.

Noticed while adding 7.2 to CI.


neilvcarvalho (approved):
Oops. Thanks for catching that. I think there's some work to do on factory_bot_rails, too - it's on my list for this Friday.

PR #1688, opened 2024-08-16
Fix the factory definition in traits documentation

ddieulivol:
## What's in this MR?

The https://thoughtbot.github.io/factory_bot/traits/using.html page has a small typo.

The second code block uses the `active` trait, but it's not defined in the Factory.

This MR adds the missing trait.


sarahraqueld (approved):
Good catch, thank you!

sarahraqueld:
The check is failing due to a known issue, so I'll go ahead and merge this.

Issue #1694, opened 2024-09-08
Make strategy used easier to detect in callbacks


CodeMeister:
### Problem this feature will solve

Like many, I use the same factory with multiple strategies, based on the individual test. 

Critically, I want the same code executed on completion, regardless of the strategy used.

With :create also calling :build I need to determine in the :after_build callback if I should execute the code or not.

The strategy can be found with `context.instance_variable_get(:@build_strategy).to_sym`, but that required a degree of familiarity with the code that most users won't have.

### Desired solution

A helper method for the context would simplify the whole process:

```ruby
factory :user do
  after(:build)   { |user, context| run_this_code() if context.strategy.build?}
  after(:stub)    { |user, context| run_this_code() }
  after(:create)  { |user, context| run_this_code() }
end
```

## Additional context

I'll do a pull request for this feature.

Issue #1649, opened 2024-05-10
Expose which keys were overridden by the user


mike-burns:
The `context`, as passed to the various hooks, does not tell us whether a key was overridden by the user. This means we can't know whether a value is the default, or intentionally set.

Let's expose this on the `context`. It could be as simple as a list of keys set by the user.

Zooip:
The list of keys is already exposed in the evaluator using the private undocumented method `__override_names__` : https://github.com/thoughtbot/factory_bot/blob/main/lib/factory_bot/evaluator.rb#L50-L52
It would make sense to me to make this method part of the public stable API.

I personnaly use it to "sync" associations' associations like so :
```ruby
transient do
  school do
    if __override_names__.include?(:exam)
      exam.school
    elsif __override_names__.include?(:student)
      student.school
    else
      association :school
    end
  end
end

exam { association :exam, school: school }
student { association :student, school: school }
```

So you technically already can use this in the callbacks but it is undocumented and not part of the public API
```ruby
after(:build) do |instance, evaluator|
  evaluator.__override_names_
end
```

CodeMeister:
I've submitted a pull request that will make it easier to check which keys were overridden by the user, and which by the factory: #1695.  let me know if this works for you.

PR #1695, opened 2024-09-08
Context helpers for callbacks:

CodeMeister:
Fixes: #1694
Fixes: #1649
 
Context is provided as a block parameter in callbacks.  This pull requests adds some helper methods to `context`, making it easier to detect the strategy used and identify which attributes were defined by the user / factory.

Methods included are:
- :stategy
- :defined_attributes
- :user_defined_attributes
- :factory_defined_attributes

All can be queried: 
```ruby
context.strategy         #=> 'build'
context.strategy.build?  #=> true

context.user_defined_attributes           #=> [:age, :name]
context.user_defined_attributes.age?      #=> true
context.user_defined_attributes.height?   #=> false
```

Although the `Evaluator` class is marked `@api private`, being included as a callback parameter means users are already accessing it. Making it easier to use can only be a good thing.

Tests and documentation added.


CodeMeister:
Hi @sarahraqueld & @smaboshe,  any chance of reviewing the pull request?

I have another request ready to go, but it uses some of the same code from this one, so need this one approved (or rejected) first.

😄

unused:
The PR is great work @CodeMeister very nice extension to the feature set. However, I wonder what use case(s) does justify the exposure of that information? I can't think of one and never encountered problems in that way (afair). Do you think there is a risk of encouraging developers to come up with adding unnecessary complexity to their factories exposing such interface? If there's a couple of real world examples to follow I think it would be much easier to decide.

CodeMeister:
Thanks @unused.

I don't think this encourages developers to create more complex factories.  Most developers know what they need their factories to do, then hunt for the means to make it work.

If a developer needs to know the build strategy used, they will check the docs first, then google it and get `context.instance_variable_get(:@build_strategy).to_sym` in the results.

I think simplifying access to a factory's internal data & settings can only be a good thing; provided it doesn't change, or allow the user to change, the underlying information. Maintenance considerations should be minimal as these methods will only ever change if the underlying factory structure changes.

As for real-world examples.

(1)

I often improve the performance of an object's unit tests, by _stubbing_ its associations when it was being _built_, and acting normally otherwise.

I appreciate mocking and stubbing would be even faster, but then you don't catch any changes to the associated objects.


(2)

When running a complex system test, I have to pre-populate the application with '000s of records.  When an unexpected value appears in the generated data, the easiest way to track it down is to create a log file of every factory called, with the object's id, specific attributes and strategy used.  Skipping a log entry on `after(:build)` for objects being _created_ reduces this list significantly.

(perhaps I should add an `after(:all)` callback that only runs once on completion)

😃

unused:
Thank you a lot for adding that many details. Looking at the two examples I don't think there is a general case where the context build is needed and this rather solving edge cases or optimizations. However, I don't see any disadvantages when adding it. When it comes to the suggested changes, the documentation you added is definitely helpful :rocket: My suggestion would be to not add inquiry, remove attributes defined methods, as well as replace output-check in tests (e.g. just assigning values to an array and check contents). What do you think about that?

CodeMeister:
Interesting thought. 

As this request is just syntactic sugar, perhaps it represents a bigger decision that ThoughtBot needs to make. Rather than focussing on the helpers in this specific request, it should be a more general strategic decision on whether developers should, or should not, have easy access to a factory's internal data & settings.

As you said earlier, easy access will allow developers to create more complex factories and require more code/maintenance. Restricting access will mean less code/maintenance and have developers write their own helpers methods (when needed), with possible unforseen consequences for them.  

If the decision is to encourage easy access, then `context.strategy` is more friendly than `context.instance_variable_get(:@build_strategy).to_sym`, and `context.strategy.build?` is more friendly than `context.strategy == :build`

If not, then contributors will know to avoid this type of syntactic sugar.

As to the tests: we are creatures of habit.  My first thought was to parse the captured output, but you are quite correct that adding to an array would be just as effective.

@smaboshe & @sarahraqueld, does ThoughtBot have a policy or opinion on this type of issue? If not, how about setting up workshop like the 'OpenSource day' where a we could discus the issue and reach a consensus?

🤔

CodeMeister:
@unused, on a separate note, I've submitted a pull-request (#1709) for the \<attribute\> & \<attribute\>_id conflict we look at during the Open Source day.  I'd value your thoughts on it if you have the time, or inclination, to review it. 🫶

CodeMeister:
Updated the tests to remove the use of `:std_out` by recording the results within the object itself.

Great idea @unused

DoodlingDev:
Hello @CodeMeister! Thanks for your efforts here and sorry it's taken us so long to reply.

I find myself generally inclined to agree that exposing more context and possible complexity will inevitably cause developers to create more complicated factories. If it's there, they will use it. The fact that one can google how to sneakily get at a private api doesn't mean that it should be treated as public api, but if one does do that, they'll know they're off the path proposed by the gem and "there be dragons".

There could be an argument made for the idea that having to know the internals of the factory you've built is a smell. What in the actual running of the application would get something to the state that the test is trying to reach? Is there a better heuristic for whether or not to perform those additional actions? Can those actions be part of a trait, or done with another first-party mechanism?

Adding associations to a built (but not persisted) object would also not reflect the real world either. I think I understand what you're getting at with adding the associations and wanting to keep the speed benefits of `build` over `create`, but I think it's breaking the implicit semantic differences between the two strategies and it's another situation where it's not part of the framework because it is not recommended behavior. The abstraction is leaky from the get go, and we'd likely advise another approach.

DoodlingDev:
Hello @CodeMeister! Thanks for your efforts here and sorry it's taken us so long to reply.

I find myself generally inclined to agree that exposing more context and possible complexity will inevitably cause developers to create more complicated factories. If it's there, they will use it. The fact that one can google how to sneakily get at a private api doesn't mean that it should be treated as public api, but if one does do that, they'll know they're off the path proposed by the gem and "there be dragons".

There could be an argument made for the idea that having to know the internals of the factory you've built is a smell. What in the actual running of the application would get something to the state that the factory needs to know its build strategy? Is there a better heuristic for whether or not to perform those additional actions? Can those actions be part of a trait, or done with another first-party mechanism? A custom strategy or callback?

CodeMeister:
Hi @DoodlingDev, thanks for the feedback!

I agree with both yourself and @unused, that exposing this information could lead to more complex factories and we're starting to get into edge case territory.

That being said, because one strategy can directly implement another (:create triggers :build) there is no way to know which is the last strategy implemented.

Perhaps an alternative solution, that would still address my needs, would be to add an `after(:all)` callback for tear-down actions.

What do you think of closing this PR, and I'll do a new one to implement `before(:all)` and `after(:all)` callbacks.

DoodlingDev:
That sounds like a good plan, based on a little poking around the internet, it seems like a set of `:all` callbacks could be useful to manny people..as well as more directly address the `:create` triggering `:build` double execution issue.

CodeMeister:
@DoodlingDev, I've done a PR for the `before(:all)` and `after(:all)` callbacks. You can find it here: #1740

Issue #1680, opened 2024-07-20
Do not overwrite the dafault value


anko20094:
### Description

```ruby
# db/structure.sql
CREATE TABLE public.realtors (
.......
duplicate_id character varying,
duplicate boolean DEFAULT false NOT NULL,
.......
```

```ruby
# spec/factories/realtors.rb

FactoryBot.define do
  factory :realtor do
.......
duplicate_id { pry; Faker::Alphanumeric.alpha(number: 10) }
.......
```

```ruby
# in some spec
realtor1 = create(:realtor, id: 6, duplicate: true, duplicate_id: 'dup1')
[1] pry> realtor1.duplicate_id
=> nil
```

So, it does not set any value from factory neither from wrote manually

### Reproduction Steps
Create the realtors table with the duplicate_id column.
Define a factory for realtor with a duplicate_id attribute.
Create a realtor object in a test, specifying a value for duplicate_id.
Check the value of duplicate_id on the created object; it is nil instead of the expected value.

### Expected behavior
The duplicate_id should be set to the value specified manually in the test when creating a realtor object.

### Actual behavior
The duplicate_id is set to nil, ignoring both the factory-generated value and the manually specified value.

### System configuration
**factory_bot_rails (6.2.0)**:  
**rails (7.0.8)**:  
**ruby (3.2.1)**:

DoodlingDev:
a quick script we threw together to test this out...

```
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  git_source(:github) { |repo| "https://github.com/#{repo}.git" }
  gem "factory_bot", "~> 6.0"
  gem "activerecord"
  gem "sqlite3"
end

require "active_record"
require "factory_bot"
require "minitest/autorun"

ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")

ActiveRecord::Schema.define do
  create_table :realtors, force: true do |t|
    t.boolean :duplicate, default: false, null: false
    t.string :duplicate_id
  end
end

class Realtor < ActiveRecord::Base
end

FactoryBot.define do
  factory :realtor do
    duplicate { false }
    duplicate_id { "This should exist" }
  end
end

class RealtorTest < Minitest::Test
  def test_factory_assigned_value_should_be_present
    realtor = FactoryBot.create(:realtor, duplicate: true)
    assert_equal "This should exist", realtor.duplicate_id
  end

  def test_create_assigned_value_should_be_present
[...]

PR #1708, opened 2024-10-25
Add workaround for attributes with _id

FerPerales:
This has been a recurring issue that has not been documented in one single place. There are workarounds in several issues.

This PR attempts to have a single place where people can see two workarounds that have been tested.

This is not a final solution, but we want to make this less surprising for users.

PR #1709, opened 2024-10-28
BugFix: '<attribute>' and '<attribute>_id' conflict.

CodeMeister:
Fixes: #1680 
Fixes: #1142 

Addresses: #1708  

When a factory has both \<attribute\> and \<attribute\>_id, each one is considered an alias for the other
and are not assigned correctly.

This request eliminates this issue by checking if the override is an existing attribute name, before matching against an alias.

 - AttributeAssigner refactored to ignore a matching attribute alias, if it also matches the name of another attribute.
 - Renamed two methods to give more clarity to their purpose.
 - Tests added for attribute assignment.
  
It works for any recorded aliases, not just '_id'.


unused:
I had expected this problem to be much more difficult to resolve, but can't find anything wrong with your code, good job :raised_hands: When trying to understand the conflict ignored attributes - that as far as I've seen only applies to transient ones - makes the processing hard to understand. I think you managed well to make this easier to grasp.

lib/factory_bot/attribute_assigner.rb
```diff
@@ -91,22 +91,39 @@ def override_names
       @evaluator.__override_names__
     end
 
+    def attribute_names
+      @attribute_list.names
+    end
+
     def hash_instance_methods_to_respond_to
       @attribute_list.names + override_names + @build_class.instance_methods
```
DTrejo, line 99: use `attribute_names`?

CodeMeister: @DTrejo good call!   

I've updated the code to call the method.

Thanks for your comments 🫶

vburzynski (approved):

PR #1710, opened 2024-10-29
Update intro.md

AlexeyMatskevich:
Current links are not up to date


DoodlingDev:
Hi! Thanks for your PR. The links are correct when viewed from the [Factory Bot Book](https://thoughtbot.github.io/factory_bot), which is the supported way to view the documentation.

That does mean if you'd instead prefer to read the docs as markdown, you'll have to navigate to the `.md` version on your own.

Thanks again for your time.

Issue #1703, opened 2024-10-02
Traits that imply other ones can cause double running of callbacks


saty9:
### Description

When i make a trait B that implies A and build with B and A traits the after(:build) in A runs twice

### Reproduction Steps

```ruby

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  git_source(:github) { |repo| "https://github.com/#{repo}.git" }
  gem "factory_bot", "~> 6.0"
  gem "activerecord"
  gem "sqlite3"
end

require "active_record"
require "factory_bot"
require "minitest/autorun"
require "logger"

ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :posts, force: true do |t|
    t.string :body
  end
end

# TODO: Add any application specific code necessary to reproduce the bug
class Post < ActiveRecord::Base
end

FactoryBot.define do
  factory :post do
    body { "Post body" }

    trait :implies_foo do
      foo
    end

    trait :foo do
      after(:build) do |post|
        post.body += " with foo"
      end
    end
  end
end

class FactoryBotTest < Minitest::Test
  def test_factory_bot_stuff
    post = FactoryBot.build(:post, :foo, :implies_foo)

    assert_equal post.body, "Post body with foo"
  end
end

# Run the tests with `ruby <filename>`

```

### Expected behavior

I expect the after(:build) block to be run once

### Actual behavior

It runs twice instead

### System configuration
**factory_bot version**:  6.4.5
**rails version**:  7.2.1
**ruby version**: 3.3.5

colinross:
The aggregate callbacks get a `flatten.compact` but maybe the context is making them dissimilar?
We could wrap them in a registry like the decorators, but then we would need to have a name/identifier for each callback, otherwise how distinguish between one after_save and another, right?

CodeMeister:
Hi @saty9 & @colinross. I've added a pull-request that fixes this exact issue: #1712 

Have a look and let me know if it works for you.

😀

colinross:
Looks promising to me.

PR #1712, opened 2024-11-04
BugFix: Ensure callbacks only run once per instance.

CodeMeister:
Fixes: #1703 

CallbacksObserver refactored to ignore an individual callback that has already been run on the given instance.

**EXAMPLE**
```ruby
factory :user do
  name { "Jane" }

  trait :trait_a do
    trait_b
  end

  trait :trait_b do
    after(:build) { |user| user.name += " Doe" }
  end
end

user = build(:user, :trait_a, :trait_b)
```

**BEFORE:**
`user.name #=> "Jane Doe Doe"`

**AFTER:**
`user.name #=> "Jane Doe"`

See issue: #1703 for the full issue.


neilvcarvalho (approved):
Looks great - Thank you!

PR #1725, opened 2025-01-24
Update development dependencies

neilvcarvalho:
We have some outdated dependencies, and are seeing failures on the CI environment for Rails 6.1.

Most of the errors happened due to locked dependencies that are incompatible with some Ruby versions. That was happening because we added the Appraisals lockfiles to the version control, which isn't recommended. Each Ruby version (combined with each Appraisal) should resolve the dependencies themselves, downloading the latest compatible version.

I also updated all dependencies from the gemspec to their latest version. From now on, Dependabot should pick whatever dependencies are updated and we'll know if some of them broke anything. The dependencies were updated using Ruby 3.3, as Ruby 3.4 has a failed spec that I'll investigate in the near future.


neilvcarvalho:
I just noticed a _very_ important notice on the Appraisals README.md:

> Version Control
> When using Appraisal, we recommend you check in the Gemfiles that Appraisal generates within the gemfiles directory, but exclude the lockfiles there (*.gemfile.lock.) The Gemfiles are useful when running your tests against a continuous integration server.

That makes things _way_ easier.

.standard_todo.yml
```diff
@@ -0,0 +1,6 @@
+# Auto generated files with errors to ignore.
```
neilvcarvalho, line 1: An alternative to adding a TODO file is locking `standard` to a specific version. I added this because it's outside the scope of this PR.

seanpdoyle (approved):

PR #1731, opened 2025-02-01
Small update to "verbose linting" spec to allow for Ruby 3.4 backtrace changes

CodeMeister:
Update for testing in Ruby 3.4

- file:   spec/acceptance/lint_spec.rb:188

THE PROBLEM THIS FIXES

Ruby 3.4 backtrace locations now include both the object and method names in the format:
 "<object_name>#<method_name>.

So a backtrace location with "...`save!`" in ruby 3.3, now becomes "... 'InvalidThing#save!'" in Ruby 3.4.

THE SOLUTION

The solution was to update the test regex to allow for either option:
  %r{#{__FILE__}:\d*:in ('InvalidThing#save!'|`save!')}

### DIFF
--- a/spec/acceptance/lint_spec.rb
+++ b/spec/acceptance/lint_spec.rb
@@ -211,7 +211,7 @@ describe "FactoryBot.lint" do
         FactoryBot.lint(verbose: true)
       }.to raise_error(
         FactoryBot::InvalidFactoryError,
-        %r{#{__FILE__}:\d*:in `save!'}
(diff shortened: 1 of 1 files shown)


neilvcarvalho (approved):
TY! I was going to add this as an issue as I didn't have time to investigate yesterday, and forgot. This unlocks adding Ruby 3.4 to the test matrix, which I was trying to add initially.

Issue #1727, opened 2025-01-30
Show the registered traits in the "Trait not registered" error message


neilvcarvalho:
### Problem this feature will solve

Sometimes, when creating a new record with Factory Bot, we believe a trait exists, but it doesn't. That can happen because it was written in another way, because of a typo, or just because it wasn't implemented yet.

    Failure/Error: admin = create(:user, :admin)

    KeyError:
    Trait not registered: "admin"

When that happens, we have to find the factory definition to know what are the available traits. It would be less disruptive to the flow to know what traits were registered.

### Desired solution

We could see something like:

    Failure/Error: admin = create(:user, :admin)

    KeyError:
    Trait not registered: "admin". Registered traits: "administrator", "guest"

CodeMeister:
Great idea @neilvcarvalho 

Pull request  #1717  fixes an issue with inherited traits and changes how they are recorded, so the code for this feature will depend on whether #1717 is approved or rejected.

Can you take look?  🙏

PR #1732, opened 2025-02-03
 Expanded trait error message to include list of defined traits.

CodeMeister:
## Summary

Fixes: #1727

Expands trait error messages to include a list of the factory's defined traits.

## Included

The message now includes: `Registered traits: [:trait_1, :trait_2, :trait_3]`

### Internal Definition Call
factory: :user
defined traits: :accessible_trait
error: internal reference to a missing trait :inaccessible_trait

Error Message: 
  `Trait not registered: "inaccessible_trait". Registered traits: [:accessible_trait]. Referenced within "user" definition`

### User Trait Specified
factory: :user
defined traits: :accessible_trait
error: user calls an non-existent trait with `build(:user, :missing_trait)`

Error Message: 
  `Trait not registered: "missing_trait". Registered traits: [:accessible_trait]`

## Note

When the error is because a missing trait was called from within a factory's definition, the message still includes the original:
`Referenced within "user" definition`

## Changes

- lib/factory_bot/definition.rb 
    - error-rescue moved from :base_traits to :trait_by_name to catch both definition and calling errors.

- spec/acceptance/enum_traits_spec.rb 
    - updated specs for new error message layout.

- spec/acceptance/traits_spec.rb updated specs for new error message layout. added new use-case tests with:
    - no registered traits
    - single registered trait
    - multiple registered traits
    - multiple registered traits through multiple inheritance

### DIFF
--- a/lib/factory_bot/definition.rb
+++ b/lib/factory_bot/definition.rb
@@ -125,31 +125,8 @@ module FactoryBot
 
     def base_traits
       @base_traits.map { |name| trait_by_name(name) }
-    rescue KeyError => error
-      raise error_with_definition_name(error)
-    end
-
-    # detailed_message introduced in Ruby 3.2 for cleaner integration with
-    # did_you_mean. See https://bugs.ruby-lang.org/issues/18564
-    if KeyError.method_defined?(:detailed_message)
-      def error_with_definition_name(error)
-        message = error.message + " referenced within \"#{name}\" definition"
-
-        error.class.new(message, key: error.key, receiver: error.receiver)
-          .tap { |new_error| new_error.set_backtrace(error.backtrace) }
-      end
-    else
-      def error_with_definition_name(error)
-        message = error.message
-        message.insert(
-          message.index("\nDid you mean?") || message.length,
-          " referenced within \"#{name}\" definition"
-        )
-
-        error.class.new(message).tap do |new_error|
-          new_error.set_backtrace(error.backtrace)
-        end
-      end
+    rescue KeyError => key_error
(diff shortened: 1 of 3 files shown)


neilvcarvalho:
Hey @CodeMeister, I owe you a follow-up: while reviewing this PR and trying to understand how Factory Bot works with DidYouMean (and also how RSpec and Minitest work with it), and noticed that we re-raise the "Did you mean?" messages for  factory definitions, but not for traits. I wonder if ultimately showing all the traits is the best solution (as there's DidYouMean).

At the same time, I can only see the Did you mean? message when running Factory Bot in an irb session, not on RSpec or Minitest. Now I got into this rabbit hole 😄

CodeMeister:
Hi @neilvcarvalho, it's a conundrum for sure!

DidYouMean is great when it's a close match, but not if it's completely wrong.  I think it's good feedback to list all the available traits (including inherited ones), but DidYouMean would keep it consistent with factory definitions.

How about we park this for now and I'll re-do it using DidYouMean?

neilvcarvalho:
Sounds good! In the meantime, I'll check with the RSpec and MiniTest folks how to show the error's detailed message on Ruby >= 3.2. Since this version introduced the `detailed_message` method for errors (with a good reason), we're missing valuable information in the error messages. It's funny how no one seemed to notice - I even thought that it was some setting or mistake on my end.

neilvcarvalho:
For reference, I opened the following issues:
- https://github.com/rspec/rspec/issues/191
- https://github.com/minitest/minitest/issues/1026

PR #1734, opened 2025-02-06
Fix issue introduced by YARD's global #log method

thewatts:
[The YARD gem adds a global `#log` method](https://github.com/lsegal/yard/blob/efb33b5411f54ae491c10ce5cd227fa49b985fde/lib/yard/globals.rb#L20-L22), which becomes a problem if a Factory needs to declare a `log` attribute.

Ex:

```ruby
FactoryBot.define :event do
  # This would originally call out to YARD's global log method,
  # instead of defining the attribute on the factory
  log
end
```

This commit adds `FactoryBot::DefinitionProxy#log` to ensure that if defining a `log` attribute, it'll correctly get passed to the necessary methods through `method_missing`, and therefore added as an attribute to the factory.

----

I realize that this is a bit of a workaround for something that `FactoryBot` has no control over. And it feels a little off (as other gems could technically pollute the global namespace in the same way). But given the ubiquitous usage of YARD - thought this solution may be an easy fix.

Let me know if you'd prefer me to reach out to the YARD team to see if they can make an adjustment on their end.

Additionally - this was a gnarly thing to track down. We've been using the Lookbook gem, and adding that gem caused our specs to fail, because the `log` attribute in on the record was always being assigned `nil` in the Factory. It took quite a bit of digging to find that the issue was actually YARD, and that it was their log method.

We have a workaround in place - but certainly didn't want someone else to have to wrestle through figuring out what happened.

At the very least - this PR can serve as a discoverable piece of information.

---

Thanks for all your hard work on this amazing gem/resource! 👏


DoodlingDev:
Hey Hi, thanks for taking the time to write up a PR!

I get where you're coming from with the ubiquity of YARD, but it seems like adding code for a one-off fix that might not affect everyone isn't the approach we'd want to take here. And...although given the specifics I doubt it...could adversely affect a project in the other direction by introducing an unexpected behavior in the other direction.

We've got the PR as an artifact like you said to hopefully direct people who run into the same trouble you did, let's leave it at that for now. We can always revisit if it becomes a more common problem than we anticipated.

PR #1737, opened 2025-03-11
Fix definition_file_paths comment

quadule:
The comment for `definition_file_paths` is incorrect; all matching factory paths will be loaded. This is also documented correctly at https://thoughtbot.github.io/factory_bot/ref/find_definitions.html


DoodlingDev:
Looks right to me! Thank you for the PR!

Issue #1738, opened 2025-03-14
Child inheriting parent trait that has sequence will start over


ngan:
### Description

After [this change](https://github.com/thoughtbot/factory_bot/pull/1717), children factories that inherit traits from the parent that contains a sequence will be problematic for unique attributes.

### Reproduction Steps

Consider the following:
```ruby
FactoryBot.define do
  factory(:account) do
    trait(:cool_trait) do
      sequence(:unique_number)
    end

    # ...
  end

  factory(:debit_account, parent: :account) do
    # ...
  end
end
```

Now, if I call:
```
create(:account).unique_number # => 1
create(:debit_account).unique_number # => 1
``

Previous behavior before https://github.com/thoughtbot/factory_bot/pull/1717:
```
create(:account).unique_number # => 1
create(:debit_account).unique_number # => 2
```

### System configuration
**factory_bot version**:  v6.5.1
**rails version**:  
**ruby version**:

CodeMeister:
Hi @ngan, I'll look into it ....

PR #1739, opened 2025-03-23
BUGFIX: Refactored sequences to ensure cloned traits use parent sequences.

CodeMeister:
Fixes: #1738 

THE ISSUE
-----------

Child factories clone parent traits to ensure they are executes in the correct context. 
This means that they are also registering a new version of the parent's sequence.

THE SOLUTION
----------------

When registering a child sequence, check if it's already registered by the parent and, if so, use the parent's sequence.

CHANGES
----------

Update to ensure child factories do not register their own version of a paren't factory.

- added :matches?(<other_sequence>) to Sequence class
- updated DefinitionProxy to ensure registered sequences are unique

Added test  to ensure:
- parent direct sequences are used by child
- sequences within parent traits are used by child
- children can re-define a new sequence with the same name
- Global, factory & inherited factories with sequences of the same name, all maintain their own context.

### DIFF
--- a/lib/factory_bot/definition_proxy.rb
+++ b/lib/factory_bot/definition_proxy.rb
@@ -120,9 +120,16 @@ module FactoryBot
     #
     # Except that no globally available sequence will be defined.
     def sequence(name, ...)
-      sequence = Sequence.new(name, ...)
-      FactoryBot::Internal.register_inline_sequence(sequence)
-      add_attribute(name) { increment_sequence(sequence) }
+      new_sequence = Sequence.new(name, ...)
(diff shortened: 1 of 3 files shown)


neilvcarvalho (commented):
Thank you for fixing this bug! I added a few comments, and have a few suggestions fixing typos.

CodeMeister:
@neilvcarvalho, good catch on the spelling and suggestions for renaming the test strings.

Update has been submitted 🙂

neilvcarvalho (approved):
Looks great!

PR #1740, opened 2025-03-26
FEATURE: Before(:all) and after(:all) callbacks added

CodeMeister:
Fixes: #1694 
Continues: #1695 

## SUMMARY

Continuing on from #1695 it was decided that `before(:all)` and `after(:all)` callbacks would provide a solution, without exposing context internals.

## FEATURE

Both callbacks are defined in the usual way, and operate through the normal callback process.

`before(:all)` callbacks are triggered before the factory is run. Calling it with `before(:all){ |instance, context| ... }` will provide the factory context, but both the `instance` and `context.object` will be nil.

`after(:all)` callbacks are triggered once the factory has completed it's strategy. In this case, the `instance` and `context.object` are populated as normal.

## CHANGES

The only code changes are two new lines to trigger the :before and :after callbacks in the `Factory#run` method, and a little restructuring to return the instance.

## DOCUMENTATION

I've updated the documentation for the two new callbacks, and I've also added a page describing how callbacks are ordered (which initially confused me!).

## TESTING

Additional tests have been included to test multiple scenarios.

Because `before(:all)` runs BEFORE the instance is generated, there was no easy way to track the callback.  I tried using a global `$callback_output = []` but Standard.rb complained.

The best solution was to add a new module `TestLog` which acts as an array, allowing callbacks to update it in order and the test to query the results.

I hate adding new objects, but was stuck for another solution.  I'm happy to change this if you have a better idea.

### DIFF
--- a/docs/src/callbacks/summary.md
+++ b/docs/src/callbacks/summary.md
@@ -12,7 +12,7 @@ factory\_bot makes six callbacks available:
 |after(:stub)   |called after a factory is stubbed (via `FactoryBot.build_stubbed`)|
 
 
-## Examples:
+## Examples
 
 ### Calling an object's own method after building.
 
--- a/docs/src/callbacks/summary.md
+++ b/docs/src/callbacks/summary.md
@@ -1,19 +1,44 @@
 # Callbacks
 
-factory\_bot makes four callbacks available:
+factory\_bot makes six callbacks available:
 
-* after(:build)   - called after a factory is built   (via `FactoryBot.build`, `FactoryBot.create`)
-* before(:create) - called before a factory is saved  (via `FactoryBot.create`)
-* after(:create)  - called after a factory is saved   (via `FactoryBot.create`)
-* after(:stub)    - called after a factory is stubbed (via `FactoryBot.build_stubbed`)
+|Callback|Timing|
+|---|---|
+|before(:all)   |called before a factory begins generating the object.|
+|after(:all)    |called after a factory has generated the object.|
+|after(:build)  |called after a factory is built (via `FactoryBot.build`, `FactoryBot.create`)|
+|before(:create)|called before a factory is saved (via `FactoryBot.create`)|
+|after(:create) |called after a factory is saved (via `FactoryBot.create`)|
+|after(:stub)   |called after a factory is stubbed (via `FactoryBot.build_stubbed`)|
 
-Examples:
+
+## Examples:
+
+### Calling an object's own method after building.
 
 ```ruby
-# Define a factory that calls the generate_hashed_password method after the user factory is built
+## 
+# Define a factory that calls the generate_hashed_password method
+# after the user factory is built.
+#
+# Note that you'll have an instance of the object in the block
+#
 factory :user do
   after(:build) { |user, context| generate_hashed_password(user) }
 end
 ```
 
-Note that you'll have an instance of the object in the block.
+### Skipping an object's own :after_create callback
+
+```ruby
+##
+# Disable a model's own :after_create callback that sends an email 
+# on creation, then re-enable it afterwards
+#
+factory :user do
+  before(:all){ User.skip_callback(:create, :after, :send_welcome_email) }
+  after(:all){ User.set_callback(:create, :after, :send_welcome_email) }
+end
+```
+
+
--- a/docs/src/callbacks/summary.md
+++ b/docs/src/callbacks/summary.md
@@ -40,5 +40,3 @@ factory :user do
   after(:all){ User.set_callback(:create, :after, :send_welcome_email) }
 end
 ```
-
-
--- a/docs/src/callbacks/summary.md
+++ b/docs/src/callbacks/summary.md
@@ -2,14 +2,14 @@
 
 factory\_bot makes six callbacks available:
 
-|Callback|Timing|
-|---|---|
-|before(:all)   |called before a factory begins generating the object.|
-|after(:all)    |called after a factory has generated the object.|
-|after(:build)  |called after a factory is built (via `FactoryBot.build`, `FactoryBot.create`)|
-|before(:create)|called before a factory is saved (via `FactoryBot.create`)|
-|after(:create) |called after a factory is saved (via `FactoryBot.create`)|
-|after(:stub)   |called after a factory is stubbed (via `FactoryBot.build_stubbed`)|
+| Callback        | Timing                                                                                                                    |
+| --------------- | ------------------------------------------------------------------------------------------------------------------------- |
+| before(:all)    | called before a factory constructs an object (via `FactoryBot.build`, `FactoryBot.create`, or `FactoryBot.build_stubbed`) |
+| after(:build)   | called after a factory builds an object (via `FactoryBot.build` or `FactoryBot.create`)                                   |
+| before(:create) | called before a factory saves an object (via `FactoryBot.create`)                                                         |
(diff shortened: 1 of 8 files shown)


lib/factory_bot/factory.rb
```diff
@@ -43,7 +44,11 @@ def run(build_strategy, overrides, &block)
       evaluation =
         Evaluation.new(evaluator, attribute_assigner, compiled_to_create, observer)
 
-      strategy.result(evaluation).tap(&block)
+      evaluation.notify(:before_all, nil)
+      instance = strategy.result(evaluation).tap(&block)
+      evaluation.notify(:after_all, instance)
```
smudge, line 49: Suggestion: if one goal of these callbacks is to disable and then re-enable certain global side effects (like sending emails or other model callbacks), then it might make sense to put this in an `ensure` block, so that it will trigger even if `strategy.result(...)` raises an exception. Without an `ensure`, then a factory creation failure would make it difficult to trigger a reset of global state.

(I would liken this to how `rspec` guarantees the execution of `after` callbacks even when tests fail.)

WDYT?

docs/src/callbacks/summary.md
```diff
@@ -1,19 +1,44 @@
...
+|before(:all)   |called before a factory begins generating the object.|
+|after(:all)    |called after a factory has generated the object.|
+|after(:build)  |called after a factory is built (via `FactoryBot.build`, `FactoryBot.create`)|
+|before(:create)|called before a factory is saved (via `FactoryBot.create`)|
+|after(:create) |called after a factory is saved (via `FactoryBot.create`)|
+|after(:stub)   |called after a factory is stubbed (via `FactoryBot.build_stubbed`)|
 
-Examples:
+
+## Examples:
```
vburzynski, line 15: **nit:** Remove trailing punctuation

```suggestion
## Examples
```

docs/src/callbacks/summary.md
```diff
@@ -1,19 +1,44 @@
...
+
+### Calling an object's own method after building.
```
vburzynski, line 17: **nit:** Remove trailing punctuation

```suggestion
### Calling an object's own method after building
```

spec/support/containers/test_log.rb
```diff
@@ -0,0 +1,26 @@
+##
+# Designed for tests to log output for later evaluation
+#
+module TestLog
+  class << self
+    require "forwardable"
```
vburzynski, line 6: **suggestion:** Move the `require` to the top of the file

spec/support/containers/test_log.rb
vburzynski, line 1: **praise:** 😎 simple yet effective

docs/src/callbacks/summary.md
```diff
@@ -1,19 +1,44 @@
...
-* after(:create)  - called after a factory is saved   (via `FactoryBot.create`)
-* after(:stub)    - called after a factory is stubbed (via `FactoryBot.build_stubbed`)
+|Callback|Timing|
+|---|---|
+|before(:all)   |called before a factory begins generating the object.|
+|after(:all)    |called after a factory has generated the object.|
+|after(:build)  |called after a factory is built (via `FactoryBot.build`, `FactoryBot.create`)|
+|before(:create)|called before a factory is saved (via `FactoryBot.create`)|
+|after(:create) |called after a factory is saved (via `FactoryBot.create`)|
+|after(:stub)   |called after a factory is stubbed (via `FactoryBot.build_stubbed`)|
```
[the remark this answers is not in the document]
CodeMeister: Want me to have a go at the other documentation, to make it all similar phrasing?

smudge:
@CodeMeister Heh, naming things is hard!

I don't want to over-complicate this, and I don't think it merits adding an additional callback type just to have one use an `ensure`.

I was toying around with wrapping FactoryBot calls and realized that the ideal `ensure` guarantees might come in the form of an "around" hook, so maybe it would make sense for FactoryBot to support `around` in addition to `before` and `after`.

To prove this out, I ended up defining a custom extension [here](https://github.com/Betterment/demo_mode/blob/61a0a3b2d94eac18c2b60cf48b9c826c32e43af5/lib/demo_mode/factory_bot_ext.rb#L16). The usage looks like this:

```ruby
FactoryBot.around_each do |&blk|
  was_enabled = MyLogger.enabled?
  MyLogger.disable!
  blk.call
ensure
  MyLogger.enable! if was_enabled
end
```

Note that I favored the usage of "each" instead of "all" or "run" -- maybe this was a little rspec-inspired, but "all" communicates to me that it will only run once (e.g. similar to `around(:suite)`) whereas `around(:each)` (or `around_each` in the above example) communicates that it wraps each factory instance's action.

Those are my thoughts, take them or leave them! :heart:

vburzynski:
> @vburzynski , do you have thoughts on the above `before(:run)` & `after(:run)` idea?

🤔 I think the renaming could add some clarity. Though, while the `run` keyword could work, it's making reference to the `FactoryBot::Factory#run` method which isn't generally encountered when using `factory_bot`. Though it does make sense in that the hook is invoked before any build/construction strategy. So I don't think `all` is the best fit: `"all"` indicates every item in a collection while `"any"` references some of the collection or a subset. 

I would recommend using `before(:any)` & `after(:any)`

For the other element — adding another final callback hook invoked in an `ensure` — I'm not seeing the case for that. That feels like it's outside of the concerns and "language" for `factory_bot`. It feels a bit more like a concern that belongs in a level of abstraction above where factory_bot operates.

vburzynski:
TL;DR — After re-evaluation `before(:all)` and `after(:all)` feels more natural

I think I need to backtrack on my previous comments about suggesting `before(:any)` and `after(:any)`. I started experimenting with the change; it felt odd; and then I made a realization...

I think there are two ways of looking at the callback events:

1. They represent lifecycle events surrounding the execution of _specific strategies_
2. They represent lifecycle events surrounding operations/steps/phases that occur while a factory is constructing the object instance (or product) that is returned.

So looking at the three main default strategies and the events that they trigger...

- invoking `factory.build`:
  - before_all
  - after_build
  - after_all
- invoking `factory.create`:
  - before_all
  - after_build
  - before_create
  - after_create
  - after_all
- invoking `factory.build_stubbed`
  - before_all
  - _FactoryBot "builds" the object but `after_build` is NOT invoked_
  - after_stub
  - after_all

So the Create strategy is unique in that it identifies two stages (build and create). It also uses "create" not only to indicate the strategy that ran, but also to indicate the object was saved ("created").

The Build strategy fires off `after_build` to indicate that the build strategy completed and also indicates that all the factory attributes were evaluated (the object was "built").

And then there's the Build Stubbed strategy. It still evaluates all the factory attributes, but does NOT fire an `after_build` event. It only fires an `after_stub` event to indicate the strategy completed.

Which I say all of that only because it's given me some ideas to think about beyond the scope of this PR... 🤔 🧠

So.... `before(:all)` and `after(:all)` feels like the right choice for now.

vburzynski:
Oh, one other possible confusion with my prior idea of using `:any`: 
It only makes sense when you frame the callback events in terms of the strategy used. So using `after(:any)` reads as "run this callback after _any_ strategy is used".  However, it causes confusion if you look at the events in terms of operations or steps. In that mindset, `after(:any)` is confusing, because it could mean "run this after any step/phase/operation that occurs while a factory is constructing an object". In such a scenario, it's ambiguous of whether it should run after the object is "built", or after the object is "created", or even after both. Yet it would run after everything is done.

So again `before(:all)` and `after(:all)` feels like the right choice

Issue #1751, opened 2025-06-02
Undocumented breaking change in 6.5.3 for attributes named `definition`


vlad-pisanov:
as of 6.5.3, model attributes named `definition` are no longer populated

```ruby
definition do
  'foo'
end
```

changing this to (or downgrading to 6.5.2) fixes it:
```ruby
add_attribute(:definition) do
  'foo'
end
```

arcreative:
Had this same issue, was a bit miffed how something so stable suddenly broke--fixed it the same way.

arcreative:
While it's not a _huge_ issue, I'm sure most people would prefer to not have their CI fail suddenly on a patch version upgrade!

I would say if it's simple to update to root_definition or something more obscure, that would be great!

CodeMeister:
I've updated the PR so that no changes to the API will be necessary.

neilvcarvalho:
Released on [v6.5.4](https://github.com/thoughtbot/factory_bot/releases/tag/v6.5.4)

Thank you all for reporting (and for the patience)

PR #1752, opened 2025-06-02
FactoryBot::DefinitionProxy: remove `definition` `attr_reader`

CodeMeister:
Fixes: #1751 

A new `attr_reader :definition` was added to FactoryBot::DefinitionProxy.

When trying to set an instance attribute named `:definition` using `:method_missing`, it's targeting the proxy `:definition` rather than creating an attribute, leading to an exception

I've removed the `:instance` method and changed all  references to  `@instance` 

🤔

### DIFF
--- a/lib/factory_bot/definition_proxy.rb
+++ b/lib/factory_bot/definition_proxy.rb
@@ -19,12 +19,12 @@ module FactoryBot
       undef_method(method) unless UNPROXIED_METHODS.include?(method.to_s)
     end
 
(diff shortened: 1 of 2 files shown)


CodeMeister:
Hi @neilvcarvalho,

No matter what we re-name the method to, there will always be the potential for a conflict.  I suggest leaving things as they are and updating the documentation to explain the possible 'gotcha'.

Strangely, there has always been the potential for this clash with the `:child_factories` method too 🤷

Issue #1754, opened 2025-07-10
`#generate` method is throwing an error when used inside of a factory definition when `FactoryBot::Syntax::Methods` is included


alanoliveira:
### Description

When `include FactoryBot::Syntax::Methods` is used, the `#generate` method throws an error when its called inside the definition of a factory.
I think this was introduced on this commit https://github.com/thoughtbot/factory_bot/commit/0abe26c81a8c78c0214494e5d399cc856b95d8bb

### Reproduction Steps

```ruby
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'factory_bot'
end

require 'factory_bot'
require 'minitest/autorun'

class Person
  attr_accessor :n1, :n2
end

FactoryBot.define do
  sequence :num, 1
end

FactoryBot.define do
  factory :person do
    n1 { FactoryBot.generate(:num) } # ok
    n2 { generate(:num) } # ArgumentError: wrong number of arguments (given 3, expected 1)
  end
end

include FactoryBot::Syntax::Methods

class BugTest < ActiveSupport::TestCase
  def test_sequence_generation
    assert_equal 1, generate(:num) # ok

    p = build(:person) # error
    assert_equal 2, p.n1
    assert_equal 3, p.n2
  end
end

```

### Expected behavior

The test pass

### Actual behavior

The following error is thrown
```
# Running:

E

Finished in 0.001883s, 530.9938 runs/s, 530.9938 assertions/s.

  1) Error:
BugTest#test_sequence_generation:
ArgumentError: wrong number of arguments (given 3, expected 1)
    .../factory_bot-6.5.4/lib/factory_bot/evaluator.rb:54:in 'increment_sequence'
    .../factory_bot-6.5.4/lib/factory_bot/syntax/methods.rb:126:in 'FactoryBot::Syntax::Methods#generate'
    main.rb:23:in 'block (3 levels) in <main>'
    .../factory_bot-6.5.4/lib/factory_bot/attribute/dynamic.rb:16:in 'BasicObject#instance_exec'
    .../factory_bot-6.5.4/lib/factory_bot/attribute/dynamic.rb:16:in 'block in FactoryBot::Attribute::Dynamic#to_proc'
    .../factory_bot-6.5.4/lib/factory_bot/evaluator.rb:75:in 'BasicObject#instance_exec'
    .../factory_bot-6.5.4/lib/factory_bot/evaluator.rb:75:in 'block in define_attribute'
[...]

PR #1756, opened 2025-07-12
BugFix: standardise parity on :increment_sequence

CodeMeister:
Fixes: #1754

The issue was `.increment_sequence` within `FactoryBot::Evaluator` was un-defining, then replacing,  the `.increment_sequence` method included from `FactoryBot::Syntax::Methods`, but there was a parity miss-match.

- Refactored :increment_sequence in both modules to remove the URI requirement.
- Updated sequence specs for both direct and implicit strategy use.
  - FactoryBot.build(:user)
  - build(:user)

No change to the user API.

### DIFF
--- a/lib/factory_bot/evaluator.rb
+++ b/lib/factory_bot/evaluator.rb
@@ -51,8 +51,15 @@ module FactoryBot
       @overrides.keys
     end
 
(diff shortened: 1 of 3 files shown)


neilvcarvalho (approved):
Thank you!

PR #1757, opened 2025-07-18
Cleanup Active Support core extensions

neilvcarvalho:
Reviewing the Active Support usage, I noticed two `require` calls that are unnecessary:

We don't use the `Hash#except!` method defined on `active_support/core_ext/hash/except.rb`, nor the `Hash#with_indifferent_access` defined on `active_support/hash_with_indifferent_access.rb`.

We _do_ use the `HashWithIndifferentAccess` class directly, so we're now requiring `"active_support/hash_with_indifferent_access"` instead.

### DIFF
--- a/lib/factory_bot/evaluator.rb
+++ b/lib/factory_bot/evaluator.rb
@@ -1,4 +1,3 @@
-require "active_support/core_ext/hash/except"
 require "active_support/core_ext/class/attribute"
 
 module FactoryBot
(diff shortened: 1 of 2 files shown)


neilvcarvalho:
`HashWithIndifferentAccess` loads `active_support/core_ext/hash/except`, so this PR is mostly a no-op. Since we don't use it directly, it's still better not to require it.

I'm merging right away since this is a safe change.

PR #1760, opened 2025-08-01
Add support for before build callback

### LINKED ISSUE #1633: Support for callback before(:build)

I want to disable our `Bullet` gem when FactoryBot is building instances and creating records, because I don't care about n+1 problems inside my test setup.

The problem is, that there seems to be no callback called `before(:build)`. I had a look into the source code but didn't find the right place to patch it.

Any help is appreciated where to start, so I maybe I can try for a PR.

### Desired solution

I want to be able to use something like this

```
FactoryBot.define do
  before(:build) do |record|
    Bullet.enable = false
  end

  after(:create) do |record|
    Bullet.enable = true
  end
end
```

neilvcarvalho:
(originally from @mohammednasser-32 - I couldn't push modifications to your branch, so I'm creating a new one with your original commits and my new ones)

Fixes #1633
Supersedes #1639

add support for `before(:build)` callback as requested in the issue

Tested the changes by adding the local gem to a project
image
image

### DIFF
--- a/lib/factory_bot/strategy/build.rb
+++ b/lib/factory_bot/strategy/build.rb
@@ -6,6 +6,8 @@ module FactoryBot
       end
 
       def result(evaluation)
+        evaluation.notify(:before_build, nil)
+
         evaluation.object.tap do |instance|
           evaluation.notify(:after_build, instance)
         end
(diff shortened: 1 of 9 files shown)


lib/factory_bot/strategy/build.rb
```diff
@@ -6,6 +6,8 @@ def association(runner)
       end
 
       def result(evaluation)
+        evaluation.notify(:before_build, nil)
```
neilvcarvalho, line 9: Added only to the strategies that also trigger `after_build`, skipping `build_stubbed` (debatable) and `attributes_for`

PR #1761, opened 2025-08-07
Allow the initial value of a sequence to be lazy loaded

ajesler:
## Allow the initial value of a sequence to be evaluated on first use

Note: This is a re-implementation of the changes in https://github.com/thoughtbot/factory_bot/pull/1674 to remove the need to use a `lazy` kwarg.

In order to be able to use factory_bot to generate seeds for a database where the database might not have the schema set up yet or might have data already populated, it is useful to be able to defer assigning the initial value to a sequence until it is needed. 
Currently, factory_bot needs the initial value of the sequence to be defined when the factory file is loaded.

This lazy loading behaviour has been very handy for use when we need to create objects that have unique sequential id in formats specific to third party systems. We can use these factories in our development and CI environments without having to think about which values have already been used.

I found https://github.com/thoughtbot/factory_bot/pull/1434 and thought it was worth submitting this PR.

This PR adds the ability to pass a proc as the initial value of a sequence method. The supplied proc will be called the first time as sequence is used via `next`, to set the initial value.
After the initial `next` call, these sequences behave exactly the same as any other factory_bot sequence.

Rewinding a lazy sequence uses the originally calculated value, it does not re-call the lazy Proc object. I'm not sure if it would make more sense to re-evaluate the proc on a rewind call?

Example usage:

```ruby
FactoryBot.define do
  sequence(:email, proc {  User.count + 1 }) { "user#{_1}@example.com" }
end

generate(:email) # user42@example.com if you already have 41 users
```

### DIFF
--- a/docs/src/sequences/initial-value.md
+++ b/docs/src/sequences/initial-value.md
@@ -8,3 +8,11 @@ factory :user do
   sequence(:email, 1000) { |n| "person#{n}@example.com" }
 end
 ```
+
+The initial value can also be lazily set by passing a Proc as the value. This Proc will be called the first time the `sequence.next` is called.
+
+```ruby
+factory :user do
+  sequence(:email, proc { Person.count + 1 }) { |n| "person#{n}@example.com" }
+end
+```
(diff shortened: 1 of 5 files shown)


ajesler:
That much more elegant `EnumeratorAdapter` @vburzynski , let's go with that.

I've pushed a change to implement your suggested refactoring, and to add a spec that lazy sequences work with any object that responds to `call`.

PR #1768, opened 2025-09-12
Fix association override precedence over trait foreign keys

### LINKED ISSUE #1767: Associations cannot override trait-defined ID attributes since v6.5.5
### Description

https://github.com/thoughtbot/factory_bot/pull/1709 introduces surprising behavior, particularly for Rails apps, where associations can no longer override traits that set those associations by ID. This is is particularly surprising in the Rails case, where developers can generally expect setting association IDs and associations themselves to behave similarly.

There may be a best-practices argument for encouraging developers to not set both the ID and association attributes, but without a way of enforcing that, it's not a particularly scalable solution, and unfortunately the failure mode for this new behavior isn't terribly obvious.

### Reproduction Steps

Here's the modified reproduction script. This test passes in 6.5.4 but fails in 6.5.5.
```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"
  gem "factory_bot", "~> 6.0"
  gem "activerecord"
  gem "sqlite3"
end

require "active_record"
require "factory_bot"
require "minitest/autorun"
require "logger"

ActiveRecord::Base.establish_connection(adapter: "sqlite3", database: ":memory:")

  create_table :users, force: true do |t|
    t.string :name
  end
  
  create_table :posts, force: true do |t|
    t.string :body
    t.integer :user_id
  end
end

class User < ActiveRecord::Base
end

class Post < ActiveRecord::Base
  belongs_to :user
end

FactoryBot.define do
  factory :user do
    name { "Test User" }
  end
  
  factory :post do
    body { "Post body" }
    association :user
    
    trait :with_system_user_id do
      user_id { 999 }
    end
  end
end

  def test_association_overrides_trait_user_id
    # This test passes in 6.5.4 but fails in 6.5.5
    user = FactoryBot.create(:user)
    post = FactoryBot.build(:post, :with_system_user_id, user: user)
    
    assert_equal user.id, post.user_id, 
      "Association should override trait's user_id. Got #{post.user_id} instead of #{user.id}"
  end
end
```

### Expected behavior

We expect `post.user_id` to be `user.id` and not `999`

### Actual behavior

`post.user_id` is still the ID set in the trait (`999`)

### System configuration
**factory_bot version**:  6.5.5
**rails version**:  7.1
**ruby version**: 3.4.5

JinOketani:
Hello, this is my first contribution to factory_bot. While using it, I came across what seems to be an unexpected behavior, so I’d like to propose a fix. This PR fixes #1767 — since v6.5.5, associations cannot override trait-defined ID attributes.

## Background / Problem
In 6.5.4, passing an association explicitly would override any *_id values provided by traits or factory defaults. After 6.5.5 (related to PR #1709), the *_id value can take precedence, causing the association and foreign key to diverge.

```ruby
FactoryBot.define do
  factory :user do
    name { "Test User" }
  end

  factory :post do
    association :user
    title { "Test Post" }

    trait :with_user_id_999 do
      user_id { 999 }
    end
  end
end

user = FactoryBot.create(:user)
post = FactoryBot.build(:post, :with_user_id_999, user: user)

# 6.5.5 (broken):
post.user_id # => 999 (trait wins; inconsistent with post.user)
# 6.5.4 (expected) and after this fix:
post.user_id # => user.id (association wins; consistent)
```

## Root Cause
The issue stems from changes made in PR #1709 to resolve attribute/attribute_id conflicts. The modified `AttributeAssigner#aliased_attribute?` method now returns `false` when the override is an actual attribute name, but this inadvertently affects association overrides.

## Solution / Approach
Refine alias/override precedence so that:
1. Explicit association overrides win over trait/factory defaults for the corresponding *_id.
2. The attribute vs. attribute_id conflict fix from PR #1709 remains intact.
3.  Behavior is unchanged when no association override is provided (i.e., trait-defined *_id still applies).

This change enhances the alias check in AttributeAssigner to ignore the *_id value when the corresponding association is explicitly overridden. While this serves as a temporary fix for the bug, a more fundamental solution may ultimately be needed.

## Changes
- Modify lib/factory_bot/attribute_assigner.rb to prioritize association overrides while preserving PR #1709’s fix.
- Add spec/acceptance/association_override_regression_spec.rb covering:
- Association override > trait-defined foreign key
- Trait-defined foreign key when no association override is present
- Multiple traits defining foreign keys

## Tests
- New regression specs added (see above).
- Existing test suite passes locally and on CI.
- Alias behavior introduced by PR #1709 remains verified.

## Backward Compatibility / Risk
- Restores the intuitive, pre-6.5.5 precedence without altering behavior where no override is supplied.
- Touches alias resolution only; no performance impact expected.
- No changes to the public API.

## Before / After

```ruby
# Before (6.5.5 - broken)
user = FactoryBot.create(:user)
post = FactoryBot.build(:post, :with_user_id_999, user: user)
post.user_id # => 999 (WRONG)
post.user    # => user (inconsistent)

# After (this PR - correct)
user = FactoryBot.create(:user)
post = FactoryBot.build(:post, :with_user_id_999, user: user)
post.user_id # => user.id (CORRECT)
post.user    # => user (consistent)
```

### DIFF
--- a/spec/acceptance/attribute_aliases_spec.rb
+++ b/spec/acceptance/attribute_aliases_spec.rb
@@ -389,7 +389,7 @@ describe "attribute aliases" do
 
         # A regression in v6.5.5 created an erroneous second user and assigned
         # that to post.user and post.user_id.
-        expect(post.user).to eq user
+        expect(post.user.id).to be user.id
         expect(post.user_id).to eq user.id
         expect(User.count).to eq 1
       end
@@ -434,7 +434,10 @@ describe "attribute aliases" do
 
       it "handles an override of the foreign key when :user_id is declared last" do
         FactoryBot.define do
-          factory :user
+          factory :user do
+            name { "tester" }
+            age { 99 }
+          end
           factory :post do
             user
             # this :user_id attribute is purposely declared after :user
@@ -448,7 +451,7 @@ describe "attribute aliases" do
         # A regression from v6.5.5 assigns the expected values to post.user and post.user_id
         # An erroneous second user instance, however, is created in the background
         expect(post.user_id).to eq user.id
-        expect(post.user).to eq user
+        expect(post.user.id).to be user.id
         expect(User.count).to eq 1
       end
     end


neilvcarvalho:
Thanks, @JinOketani! I'm playing a bit with the code and trying to find ways to break it. I'm calling it a day today, but I'll be back to continue reviewing this PR next week.

JinOketani:
Thanks @neilvcarvalho! Hope this helps. Here are the reproduction steps.

### Environment
- **Ruby**: 3.3.6
- **Rails**: 8.0.2
- **ActiveRecord**: 8.0.2 (SQLite3)
- **FactoryBot**: 6.5.5
- **FactoryBot Rails**: 6.5.1
- **RSpec Rails**: 8.0.2

### Expected Behavior (v6.5.4 and earlier)
When passing both a trait with a foreign key and an explicit association, the explicit association should take precedence and the foreign key should be consistent.

```ruby
book = create(:book)
book_author = build(:book_author, :with_book_id_999, book: book)

# Expected:
book_author.book_id   # => book.id (1)
book_author.book      # => book (consistent)
```

### Actual Behavior (v6.5.5)
The trait-defined foreign key overrides the explicit association, creating inconsistent state.

```ruby
book = create(:book)
book_author = build(:book_author, :with_book_id_999, book: book)

# Actual:
book_author.book_id   # => 999 (from trait)
book_author.book      # => nil (association ignored)
```

## Reproduction Steps
### 1. Set up models and factories

**Models:**
```ruby
class Author < ApplicationRecord
  has_many :book_authors, dependent: :destroy
  has_many :books, through: :book_authors

  validates :name, presence: true
  validates :email, presence: true, format: { with: URI::MailTo::EMAIL_REGEXP }
end

class Book < ApplicationRecord
  has_many :book_authors, dependent: :destroy
  has_many :authors, through: :book_authors

  validates :title, presence: true
  validates :isbn, presence: true, uniqueness: true
  validates :published_at, presence: true
end

class BookAuthor < ApplicationRecord
  belongs_to :book
  belongs_to :author

  validates :book, presence: true
  validates :author, presence: true
  validates :role, inclusion: { in: ['author', 'co-author', 'editor', 'translator'] }
end
```

**Factories:**
```ruby
FactoryBot.define do
  factory :book_author do
    role { 'author' }
    contribution_percentage { 100 }
    notes { 'Primary author of this work' }
    active { true }

    association :book
    association :author

    # This trait reproduces the PR #1768 regression
    trait :with_book_id_999 do
      book_id { 999 }
    end
  end

  factory :author do
    sequence(:name) { |n| "Author #{n}" }
    sequence(:email) { |n| "author#{n}@example.com" }
    bio { "A talented writer" }
    active { true }
  end

  factory :book do
    sequence(:title) { |n| "Book Title #{n}" }
    sequence(:isbn) { |n| "978-#{sprintf('%010d', n)}" }
    published_at { 1.year.ago }
    genre { "Fiction" }
    active { true }
  end
end
```

### 2. Create test case that demonstrates the issue

```ruby
RSpec.describe 'FactoryBot: Trait Foreign Keys Override Explicit Associations' do
  it 'demonstrates trait foreign key overriding explicit association' do
    # Create a book we want to associate
    book = create(:book, title: 'Expected Book')

    # This should use the explicit book association, ignoring trait's book_id
    book_author = build(:book_author, :with_book_id_999, book: book)

    puts "Expected book_id: #{book.id}"
    puts "Actual book_id: #{book_author.book_id}"
    puts "Association: #{book_author.book.inspect}"

    # This fails - trait foreign key wins over explicit association
    expect(book_author.book_id).to eq(book.id), 
      "Expected explicit association to override trait foreign key"
    expect(book_author.book).to eq(book), 
      "Expected association to be set correctly"
  end
end
```

### 3. Run the test to see the failure

```bash
bundle exec rspec spec/models/book_author_factory_regression_spec.rb
```

**Expected Output:**
```
Expected book_id: 1
Actual book_id: 999
Association: nil

Failures:
1) Expected explicit association to override trait foreign key
```

PR #1770, opened 2025-10-03
Refactor factory calculator

vburzynski:
This is a refactoring to remove the need to instantiate an instance of `StrategyCalculator` and use a more functional approach.


neilvcarvalho (approved):
Awesome - thanks!

PR #1771, opened 2025-10-03
Refactor FactoryBot module

vburzynski:
This is an experiment with refactoring some of the sprawl of the root/core FactoryBot module. 

- previously it was necessary for developers to discover that the files for aliases, find definitions, reload, and the default syntax extended the `FactoryBot` module
- with the refactoring the code in `lib/factory_bot.rb` explicitly reveals which modules extend the `FactoryBot` module


neilvcarvalho (approved):
I like it. It should break a few monkey patches given how things are moving around, but that's a risk of monkey patching. This should be included in at least a minor version release, or perhaps a major version bump depending on how far this refactor goes.

PR #1774, opened 2025-10-03
refactor: StrategySyntaxMethodRegistrar doesn't need to be a class

vburzynski:
**NOTE** — this is dependent on the `refactor-internal-module` branch #1772

The `StrategySyntaxMethodRegistrar` doesn't need to  be instantiated (and doesn't need to be a class) to perform the behavior it implements. This refactors it into a more functional module.

### DIFF
--- a/lib/factory_bot/strategy_syntax_method_registrar.rb
+++ b/lib/factory_bot/strategy_syntax_method_registrar.rb
@@ -1,37 +1,20 @@
 module FactoryBot
   # @api private
-  class StrategySyntaxMethodRegistrar
-    def initialize(strategy_name)
-      @strategy_name = strategy_name
+  module StrategySyntaxMethodRegistrar
+    def self.define_strategy_methods(strategy_name)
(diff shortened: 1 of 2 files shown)


lib/factory_bot/strategy_syntax_method_registrar.rb
```diff
@@ -1,37 +1,20 @@
...
   # @api private
-  class StrategySyntaxMethodRegistrar
-    def initialize(strategy_name)
-      @strategy_name = strategy_name
+  module StrategySyntaxMethodRegistrar
+    def self.define_strategy_methods(strategy_name)
+      define_singular_strategy_method(strategy_name)
+      define_list_strategy_method(strategy_name)
+      define_pair_strategy_method(strategy_name)
     end
```
neilvcarvalho, line 8: Hum, I think I prefer the object oriented approach given how how `strategy_name` is passed around. I'd consider defining the `define_strategy_methods` class method as `new(strategy_name).define_strategy_methods`

PR #1790, opened 2025-10-30
Ensure linting transactions don't interfere with ActiveRecord lifecycle

NikolasPilavakis:
Our FactoryBot linting started failing for a factory despite evidence that the trait combination was producing a valid model in test. We eventually traced that down to the change made in #1726. Introducing a `ActiveRecord::Base.transaction` meant that the `after_commit` hooks of our model weren't running during linting.

This is because of the way `ActiveRecord::Base` handles transactions. When changes are inserted for an ActiveRecord within an `ActiveRecord::Base.transaction`, that transaction keeps a list of all the records it contains changes for. as part of the transaction committing, `*_commit` record callbacks are fired

`ActiveRecord::Base.transaction`s are `joinable` by default, which means that `ActiveRecord` can choose to join an existing transaction instead of inserting a new one. This means that `*_commit`s that would previously result from the object's lifecycle are not sent until the end of the linting transaction.

To reinstate the previous behaviour, this PR ensures that the linting transaction is not joinable and any `*_commit` callbacks are respected.

### DIFF
--- a/lib/factory_bot/linter.rb
+++ b/lib/factory_bot/linter.rb
@@ -109,7 +109,7 @@ module FactoryBot
 
     def in_transaction
       if defined?(ActiveRecord::Base)
-        ActiveRecord::Base.transaction do
+        ActiveRecord::Base.transaction(joinable: false) do
           yield
           raise ActiveRecord::Rollback
         end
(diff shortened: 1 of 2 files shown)


NikolasPilavakis:
Hi @neilvcarvalho and @vburzynski, I'd appreciate a review of this, if you have some time to spare, please.

neilvcarvalho (approved):
Thank you!

PR #1793, opened 2026-02-07
Update CI matrix

berkos:
Adds Rails 8.0x and Ruby 4.

Removes some EOL Ruby and Rails 3.0 and Rails 6.1

https://endoflife.date/rails
https://endoflife.date/ruby


neilvcarvalho (approved):
Thank you! I added a couple of commits removing the old 6.1 gemfile and changing a bit of the build.yml based to remove a duplicated entry and unnecessary comment.

PR #1824, opened 2026-07-15
Expanded trait error message to include list of defined traits.

### LINKED ISSUE #1727: Show the registered traits in the "Trait not registered" error message

iannorriswork:
Fixes: https://github.com/thoughtbot/factory_bot/issues/1727, building off of https://github.com/thoughtbot/factory_bot/pull/1732, but preserving DidYouMean behavior in traits as well.

Expands trait error messages to include a list of the factory's defined traits.

Included
The message now includes: Registered traits: [:trait_1, :trait_2, :trait_3]

Example
Internal Definition Call
factory: :user
defined traits: :accessible_trait
error: internal reference to a missing trait :inaccessible_trait

Error Message:
Trait not registered: "inaccessible_trait". Registered traits: [:accessible_trait]. Referenced within "user" definition

User Trait Specified
factory: :user
defined traits: :accessible_trait
error: user calls an non-existent trait with build(:user, :missing_trait)

Error Message:
Trait not registered: "missing_trait". Registered traits: [:accessible_trait]

Note
When the error is because a missing trait was called from within a factory's definition, the message still includes the original:
Referenced within "user" definition

Changes
lib/factory_bot/definition.rb

error-rescue moved from :base_traits to :trait_by_name to catch both definition and calling errors.
spec/acceptance/enum_traits_spec.rb

updated specs for new error message layout.
spec/acceptance/traits_spec.rb updated specs for new error message layout. added new use-case tests with:

no registered traits
single registered trait
multiple registered traits
multiple registered traits through multiple inheritance
trait DidYouMean tests

### DIFF
--- a/lib/factory_bot/definition.rb
+++ b/lib/factory_bot/definition.rb
@@ -136,13 +136,6 @@ module FactoryBot
       raise error_with_definition_name(error)
     end
 
-    def registered_trait_message(all_registered_traits)
-      if all_registered_traits.empty?
-        "No registered traits"
-      else
-        "Registered traits: #{all_registered_traits.map(&:to_sym).sort.inspect}"
-      end
-    end
 
     def all_registered_trait_names
       (defined_traits_names + Internal.traits.map(&:name)).uniq
@@ -163,6 +156,13 @@ module FactoryBot
       end
     end
 
+    def registered_trait_message(all_registered_traits)
+      if all_registered_traits.empty?
+        "No registered traits"
+      else
+        "Registered traits: #{all_registered_traits.map(&:to_sym).sort.inspect}"
+      end
+    end
     # detailed_message introduced in Ruby 3.2 for cleaner integration with
     # did_you_mean. See https://bugs.ruby-lang.org/issues/18564
     if KeyError.method_defined?(:detailed_message)
--- a/lib/factory_bot/definition.rb
+++ b/lib/factory_bot/definition.rb
@@ -127,20 +127,60 @@ module FactoryBot
     def base_traits
       @base_traits.map { |name| trait_by_name(name) }
     rescue KeyError => error
-      raise error_with_definition_name(error)
+      raise error_with_definition_name(error_with_registered_traits(error))
+    end
+
+    def additional_traits
+      @additional_traits.map { |name| trait_by_name(name) }
+    rescue KeyError => error
+      raise error_with_registered_traits(error)
+    end
+
+    def error_with_registered_traits(error)
+      message = error.message.rstrip
+      message += "." unless message.end_with?(".")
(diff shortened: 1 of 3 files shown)


lib/factory_bot/definition.rb
```diff
@@ -127,20 +127,68 @@ def callback(*names, &block)
     def base_traits
       @base_traits.map { |name| trait_by_name(name) }
     rescue KeyError => error
-      raise error_with_definition_name(error)
+      raise error_with_definition_name(error_with_registered_traits(error))
+    end
+
+    def additional_traits
```
vburzynski: **suggestion:** To match some of the other method names, this could append `_names` to the end.

lib/factory_bot/definition.rb
```diff
@@ -127,20 +127,68 @@ def callback(*names, &block)
...
+      @additional_traits.map { |name| trait_by_name(name) }
+    rescue KeyError => error
+      raise error_with_registered_traits(error)
+    end
+
+    def error_with_registered_traits(error)
+      message = error.message.rstrip
+      message += "." unless message.end_with?(".")
+      message += " #{registered_trait_message(all_registered_trait_names)}."
+      new_error = error.class.new(message, **error_options(error))
```
vburzynski: **thought:** One thing crossing my mind is that with `error_with_registered_traits` invoked and then passed into `error_with_registered_traits`, the code is creating two error objects. It could be nice to condense that down to creating a single error object.

vburzynski (commented):
Added some initial thoughts 😎. So far this looks great overall.

lib/factory_bot/definition.rb
vburzynski: 😎

vburzynski (commented):
Looks good on a scan through. 😎 

I think I'd like to allow @neilvcarvalho a chance to review (he opened the related issues) and review the specs with some more focus review the specs closer after the Hack Space.

vburzynski:
Just wanted to pop in and say I still have this on my radar to review. Things have been a bit busy after getting back from rubyconf 😆

vburzynski (approved):
Thanks for working on this 😎

PR #1830, opened 2026-07-22
Avoid `ruby2_keywords` usage

Earlopain:
`factory_bot` needs ruby >= 3.0, so there's no need for this anymore.
Ruby might deprecate this finally: https://bugs.ruby-lang.org/issues/22205

### DIFF
--- a/lib/factory_bot/decorator/invocation_tracker.rb
+++ b/lib/factory_bot/decorator/invocation_tracker.rb
@@ -6,11 +6,10 @@ module FactoryBot
         @invoked_methods = []
       end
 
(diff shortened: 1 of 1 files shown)


CodeMeister:
Hi @Earlopain.  I appreciate you want to update the method definitions, but I don't think this is wise. 

There are still many applications running on ruby 2.7 (yes I know they should upgrade, but that's another issue).  

Dropping 2.7 support just for the sake of better method definition is overkill.

Earlopain:
factory_bot already requires ruby 3.0 and will not install on earlier versions

neilvcarvalho (approved):
Thank you!

